PR #1, opened 2019-05-17
Draft entitlements file support

srfraser:
Supports both a local file (perhaps in an expanded app) and a download url.

I'm unsure whether we should explicitly say where in the `work_dir` to find the file, or give it the path inside the app and use `Path(config["work_dir"]).glob(f"**/path")` or similar to find it.


srfraser:
This seems to have been re-implemented, so I'm assuming this isn't needed any more.

PR #9, opened 2019-07-20
Treescript 1.2.1 port to the monorepo

escapewindow:
This won't be the repository of record until we get taskcluster automation in the monorepo and can ship docker images to docker hub securely. However, this is a working treescript with retries, strip/purge, and l10n bumping, and includes all of https://github.com/mozilla-releng/treescript history.


treescript/tests/test_mercurial.py
```diff
@@ -263,6 +322,55 @@ def info(msg, *args):
...
+    mocked_source_repo.return_value = "https://hg.mozilla.org/treescript-test"
+    await mercurial.log_outgoing(config, task, config["work_dir"])
+
+    assert len(called_args) == 1
+    assert "repo_path" in called_args[0][1]
+    assert is_slice_in_list(("out", "-vp"), called_args[0][0])
+    assert is_slice_in_list(("-r", "."), called_args[0][0])
+    assert is_slice_in_list(
+        ("https://hg.mozilla.org/treescript-test",), called_args[0][0]
+    )
```
Callek: why not `assert "https://hg.mozilla.org/treescript-test" in called_args[0][0]` here since we don't need the is_slice_in_list if only one arg being tested...

treescript/tests/test_mercurial.py
```diff
@@ -315,23 +423,17 @@ def info(msg, *args):
...
-    assert "repo_path" in called_args[0][1]
-    assert is_slice_in_list(("out", "-vp"), called_args[0][0])
-    assert is_slice_in_list(("-r", "."), called_args[0][0])
-    assert is_slice_in_list(
-        ("https://hg.mozilla.org/treescript-test",), called_args[0][0]
-    )
+    mocker.patch.object(mercurial, "run_hg_command", new=blow_up)
+    mocker.patch.object(mercurial, "strip_outgoing", new=clean_up)
+    with pytest.raises(PushError):
+        await mercurial.push(config, task, tmpdir)
```
Callek, line 441: note, this looks like if we fail to actually run clean_up, we're not *actually* going to fail here...

we might need something like `mocked_strip.assert_called_once` after the `with pytest.raises()` block.

treescript/tests/test_l10n.py
```diff
@@ -5,6 +5,10 @@
 import treescript.l10n as l10n
 
 
+async def noop_async(*args, **kwargs):
```
Callek, line 8: good fodder for a shared test module :-)

Callek (approved):
This looks good, I left a few inline comments on the csets that changed, none of them are blocking the merge though.

treescript/tests/test_l10n.py
escapewindow: +1. The shared test module broke when i moved `test/` out from under `treescript/`. I'm now debating whether we should move back.

(I suppose we could have a shared test module as a sibling to `treescript` that all \*scripts tests can import..... hm... )

treescript/tests/test_mercurial.py
escapewindow: Done.

treescript/tests/test_mercurial.py
escapewindow: Done.

escapewindow:
Thanks!

PR #10, opened 2019-07-26
Fix tests

escapewindow:
This PR:

- greens up the tests
- adds pinned shas for upstream deps. This doesn't include the modules in the monorepo because there are no published shas to use yet. We may be able to pin these if we either push them to pypi or allow for using taskcluster artifacts from pip.
- addresses some warnings in mypy and flake8. This includes changes in the vendored/ directory tree. Changing vendored files isn't great, but afaict the worst case will be that when we re-vendor next, we'll have to re-fix the deprecation warnings, and meanwhile we'll have cleaner test output.


Callek (approved):

escapewindow:
Phew! Green tests both locally and in travis, and I've also gotten rid of the ugly warnings and noise. I've changed this quite a bit since you approved, but a) I think we're better off with these changes than without, and b) I'm happy to follow up if there's anything in here you want changed.

PR #11, opened 2019-08-03
Langpack

Callek:
I'm opening this PR now, incase Aki is looking to work over the weekend - I don't yet have 100% coverage but I have the bulk of this done already, and it can benefit by a set of eyes, before I push coverage updates, incase the eyes beat me.

The main intent here is to add support for signing any language pack via autograph, but in-tree code will only use these servers for ja-JP-mac since it is only ever built on mac repacks, and we can more-easily do this alongside the notarization process in the same task.


iscript/src/iscript/autograph.py
```diff
@@ -472,3 +488,59 @@ def _get_omnija_signing_files(file_list):
         sig = widevine.generate_widevine_signature(signature, certificate, flags)
         fout.write(sig)
     return to
+
+
+def langpack_id(app):
+    """Return a list of id's for the langpacks.
+
+    Side Affect of checking if filenames are actually langpacks.
```
escapewindow: nit: sp: side effect

iscript/src/iscript/mac.py
```diff
@@ -494,6 +499,20 @@ def get_app_paths(config, task):
     return all_paths
 
 
+def filter_apps(all_paths, fmt, inverted=False):
+    """Filtres through a list of apps and returns a list of apps back based on filter.
```
escapewindow: nit: sp: filters

iscript/src/iscript/mac.py
```diff
@@ -494,6 +499,20 @@ def get_app_paths(config, task):
...
+
+    Args:
+        all_paths: list of App objects
+        fmt: format name to filter
+        inverted (default: False): whether or not to invert the list.
+    """
```
escapewindow, line 510: ./src/iscript/mac.py:503:1: D413 Missing blank line after last section

flake8_docstrings just wants an empty line before the final `"""`.

iscript/src/iscript/autograph.py
```diff
@@ -472,3 +488,59 @@ def _get_omnija_signing_files(file_list):
...
+    id = None
+    with langpack.open("manifest.json", "r") as f:
+        manifest = json.load(f)
+        if not (
+            "languages" in manifest
+            and "langpack_id" in manifest
+            and "applications" in manifest
+            and "gecko" in manifest["applications"]
+            and "id" in manifest["applications"]["gecko"]
+            and LANGPACK_RE.match(manifest["applications"]["gecko"]["id"])
```
escapewindow, line 512: We may want to either use `.get()` or catch `KeyError` in these checks, in case the schema changes. Not the worst thing though, as we'll still die, just without a nice error message.

iscript/src/iscript/autograph.py
```diff
@@ -472,3 +488,59 @@ def _get_omnija_signing_files(file_list):
...
+        log.info("Identified {} as extension id: {}".format(app.orig_path, id))
+        makedirs(os.path.dirname(app.target_tar_path))
+        await sign_file_with_autograph(
+            key_config,
+            app.orig_path,
+            "autograph_langpack",
+            to=app.target_tar_path,
+            extension_id=id,
+        )
+    return
```
escapewindow: This is a noop return which we can drop.

iscript/src/iscript/mac.py
```diff
@@ -1149,6 +1168,10 @@ def get_notarization_status_from_log(log_path):
     entitlements_path = await download_entitlements_file(config, key_config, task)
 
     all_paths = get_app_paths(config, task)
+    langpack_apps = filter_apps(all_paths, fmt="autograph_langpack")
+    if langpack_apps:
```
escapewindow, line 1173: If `sign_langpacks` handles an empty list gracefully, then you don't need this `if` block (we can execute the next 2 lines of code unconditionally). This makes testing coverage easier.

escapewindow (approved):
Thanks! You're close to full coverage. I think with my suggested change, you'll get full coverage. It would be useful (and hopefully straightforward) to also add a test to make sure `filter_apps` does the right thing for both non-inverted and inverted.

iscript/tests/test_autograph.py
```diff
@@ -9,19 +9,28 @@
...
 @pytest.fixture(scope="function")
 def key_config():
     return {
         "widevine_url": "https://autograph-hsm.dev.mozaws.net",
         "widevine_user": "widevine_user",
         "widevine_pass": "widevine_pass",
         "widevine_cert": "widevine_cert",
+        "langpack_url": "https://autograph-hsm.dev.mozaws.net/langpack",
+        "langpack_user": "langpack_user",
+        "langpack_pass": "langpack_pass",
```
escapewindow, line 33: let's update `config_example.yaml` and https://github.com/mozilla-releng/scriptworker-scripts/wiki/Testing-iscript#script_configyaml

iscript/src/iscript/autograph.py
Callek: This could be worth a followup, but I'm leaving it for now (at least because this matches the `signingscript` code)

Callek: Filed https://github.com/mozilla-releng/scriptworker-scripts/issues/12

iscript/src/iscript/mac.py
Callek: I updated testing coverage, and was of the mindset that if we call the "sign langpack" code we should expect to have (at least one) langpack to sign with.

JohanLorenzo:
Linking back this PR to [bug 1939751](https://bugzilla.mozilla.org/show_bug.cgi?id=1939751) which we discovered yesterday.

PR #19, opened 2019-08-21
Lockfile asynccontextmanager, LockfileFuture

escapewindow:
The `lockfile` asynccontextmanager allows us to try to find a lockfile that doesn't exist, and lock it via fcntl. (We make sure it doesn't exist to get around [intra-process locking](http://0pointer.de/blog/projects/locking.html).) If all of the lockfiles exist, we make a configurable number of attempts with a configurable sleep time in between.

The `LockfileFuture` object allows us to wrap a coroutine with this lockfile path, and change its arguments and keyword arguments with a replacement dictionary from the `lockfile_map`. For example, if we get the lockfile path `/Users/notary1/.lockfile`, perhaps we replace `%(user)s` in the args or kwarg values with `notary1`. This object also allows for wrapping with `retry_async`, should the coroutine require multiple attempts on failure.

@mozbhearsum do you think you're comfortable reviewing this? It's a sizeable patch but the bulk of it is requirements hashes.


bhearsum:
I'll have a look later today.

escapewindow:
I added https://github.com/escapewindow/scriptworker-scripts/commit/e89bea00eea9e22f6211cd2a32adfc1eba38f88b in my local repo. I can add that to this PR, or submit it as a new PR, depending where you are in this review.

bhearsum:
> I added escapewindow@e89bea0 in my local repo. I can add that to this PR, or submit it as a new PR, depending where you are in this review.

Go ahead and push it, I was about to start looking at this now.

scriptworker_client/src/scriptworker_client/aio.py
```diff
@@ -81,6 +89,163 @@
...
+                # our own lockfiles.
+                with open(path, "x") as fh:
+                    # Acquire an fcntl lock, in case other processes
+                    # use something other than ``lockfile`` to acquire
+                    # locks
+                    try:
+                        fcntl.lockf(fh, fcntl.LOCK_EX | fcntl.LOCK_NB)
+                        log.debug(acquired_msg, path)
+                        yield path
+                        return
```
bhearsum, line 149: yield followed by return is a bit of a code smell for me - it effectively turns this into a regular function. The alternative would be dropping the context manager, and just straight assigning this to a value, I think? It's possible I'm missing one of the benefits of using a context manager for this.

As I wrote the above I think I realized this _does_ actually do something on __aexit__, namely: remove the lockfile. That appears to be in `finally` to clean it up for both the success and failure case. If all of that's correct, I'm fine with this as is, but could you add a brief explanatory comment for posterity, as it's not obvious?

scriptworker_client/src/scriptworker_client/aio.py
```diff
@@ -81,6 +89,163 @@
...
+        Get a lockfile, update ``self.args`` and ``self.kwargs`` with the
+        corresponding ``self.lockfile_map`` value, then await ``self.coro``
+        (wrapped with ``retry_async`` if ``self.use_retry_async``).
+
+        """
+        async with lockfile(
+            self.lockfile_map.keys(), **self.lockfile_kwargs
+        ) as lockfile_path:
+            args = self.replace_args(self.args, self.lockfile_map[lockfile_path])
+            kwargs = self.replace_args(self.kwargs, self.lockfile_map[lockfile_path])
```
bhearsum, line 241: Doing replacement in all of the args/kwargs seems a bit excessive considering the only purpose in real usage (as I understand it) is to pass the username to the coro. Have you considered passing that as a separate arg or kwarg instead? It's hard for me to envision whether or not that's viable/better, but I wanted to raise it at least.

bhearsum (approved):
I'm approving this, but there's a couple of sanity checks below.

scriptworker_client/src/scriptworker_client/aio.py
escapewindow: Yes, this is to avoid the `LockfileError` below and also to hit the `finally` block. Sure, I can comment.

scriptworker_client/src/scriptworker_client/aio.py
escapewindow: If the coro is, for example, `run_command`, then my args might include something like `["sudo", "su", "%(user)s", "-c", "bash blah blah"]`. By allowing the recursive replacement, I can do this directly. Without it, I have to write a wrapper coro that takes the user arg and other info, to call `run_command`.

PR #21, opened 2019-09-30
Notarize packages too

nthomas-mozilla:
I've tested this with several staging releases see details on [bug 1578497](https://bugzilla.mozilla.org/show_bug.cgi?id=1578497), and in the commit messages of each change.

### DIFF
--- a/iscript/src/iscript/mac.py
+++ b/iscript/src/iscript/mac.py
@@ -282,6 +282,7 @@ async def sign_app(key_config, app_path, entitlements_path):
             # Deal with inner .app's above, not here.
             if top_dir[app_path_len:].count(".app") > 0:
                 log.debug("Skipping %s because it's part of an inner app.", abs_file)
+                continue
             # app_executable gets signed with the outer package.
             if file_ == app_executable:
                 log.debug("Skipping %s because it's the main executable.", abs_file)
--- a/iscript/src/iscript/mac.py
+++ b/iscript/src/iscript/mac.py
@@ -992,7 +992,7 @@ async def staple_notarization(all_paths, path_attr="app_path"):
             asyncio.ensure_future(
                 retry_async(
                     run_command,
-                    args=[["xcrun", "stapler", "staple", "-v", path]],
+                    args=[["xcrun", "stapler", "staple", path]],
                     kwargs={
                         "cwd": cwd,
                         "exception": IScriptError,
--- a/iscript/src/iscript/mac.py
+++ b/iscript/src/iscript/mac.py
@@ -1089,7 +1089,6 @@ async def create_pkg_files(config, key_config, all_paths):
                     semaphore,
                     run_command(
                         [
-                            "sudo",
                             "pkgbuild",
                             "--keychain",
                             key_config["signing_keychain"],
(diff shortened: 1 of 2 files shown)


iscript/src/iscript/mac.py
```diff
@@ -282,6 +282,7 @@ def _get_sign_command(identity, keychain):
             # Deal with inner .app's above, not here.
             if top_dir[app_path_len:].count(".app") > 0:
                 log.debug("Skipping %s because it's part of an inner app.", abs_file)
+                continue
```
escapewindow, line 285: oops :)

iscript/src/iscript/mac.py
```diff
@@ -569,30 +570,34 @@ def filter_apps(all_paths, fmt, inverted=False):
 
 
 # create_all_notarization_zipfiles {{{1
-async def create_all_notarization_zipfiles(all_paths, path_attr="app_path"):
+async def create_all_notarization_zipfiles(all_paths, path_attr):
```
escapewindow: Maybe name this `path_attrs` if it's a list?

iscript/src/iscript/mac.py
```diff
@@ -1081,7 +1090,6 @@ def get_notarization_status_from_log(log_path):
                     semaphore,
                     run_command(
                         [
-                            "sudo",
```
escapewindow, line 1084: This might mean we no longer need sudo on dep boxes. (That may change again if/when we revisit having scriptworker claim multiple concurrent tasks.)

iscript/src/iscript/mac.py
```diff
@@ -1173,16 +1181,23 @@ def get_notarization_status_from_log(log_path):
...
+
+    # app
     await extract_all_apps(config, all_paths)
     await unlock_keychain(
         key_config["signing_keychain"], key_config["keychain_password"]
     )
     await update_keychain_search_path(config, key_config["signing_keychain"])
     await sign_all_apps(config, key_config, entitlements_path, all_paths)
 
+    # pkg
```
escapewindow, line 1193: Depending on how long things take, we may want to re-`unlock_keychain`, especially in the l10n case where we might be signing 5 different apps and 5 different pkgs? We may hit the 10m timeout.

escapewindow (approved):
Thanks! `sign_and_pkg` should be unaffected, right?

iscript/src/iscript/mac.py
nthomas-mozilla: Yep, will do.

iscript/src/iscript/mac.py
nthomas-mozilla: Looks like about 6 minutes between unlock for signing and end of pkg signing on try, so that's a good safeguard. I'll add it.

nthomas-mozilla:
> Thanks! `sign_and_pkg` should be unaffected, right?

Yes, separate code path. Ran a try push to verify: [Bs log](https://queue.taskcluster.net/v1/task/UJoH9fh_SZmNgh1PQ3E99w/runs/0/artifacts/public/logs/live_backing.log).

iscript/src/iscript/mac.py
nthomas-mozilla: Yes, that sounds correct to me.

PR #23, opened 2019-10-04
Taskclusterify

garbas:
This PR runs tox for all current/existing project in this repo on taskcluster and removes travis integration.


docker.d/init.sh
```diff
@@ -0,0 +1,112 @@
...
+# Check for certain variables which should be set
+#
+test $PROJECT_NAME
+test $ENV
+test $COT_PRODUCT
+test $TASKCLUSTER_CLIENT_ID
+test $TASKCLUSTER_ACCESS_TOKEN
+if [ "$ENV" == "prod" ]; then
+  test $ED25519_PRIVKEY
+fi
```
escapewindow, line 23: In signingscript oremj+catlee landed a fix to [echo missing variables](https://github.com/mozilla-releng/signingscript/commit/5b5691905e9445b4cfa0e033dc3955018f9a6d04#diff-b8e93dc171018744d01390a4b1191c85) to make it easier to debug. Not sure if we want to carry that over here.

docker.d/init.sh
```diff
@@ -0,0 +1,112 @@
...
+export WORK_DIR=/app/workdir
+export WORKER_TYPE="${TRUST_DOMAIN}-${TRUST_LEVEL}-${PROJECT_NAME}${WORKER_SUFFIX}"
+export WORKER_GROUP=${WORKER_TYPE}
+export WORKER_ID_PREFIX="${WORKER_TYPE}-"
+
+#
+# ensure configuration folder exists we can write to it
+#
+mkdir -p -m 700 $CONFIG_DIR
+echo $ED25519_PRIVKEY | base64 -d > $ED25519_PRIVKEY_PATH
```
escapewindow: Also in signingscript, we [stopped double-base64-encoding the ed25519 key](https://github.com/mozilla-releng/signingscript/commit/b8c26564f137e5aa77c4d371f2c1e492cec78712#diff-b8e93dc171018744d01390a4b1191c85). We may want to standardize.

treescript/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM python:3.7
+
+RUN groupadd --gid 10001 app && \
+    useradd -g app --uid 10001 --shell /usr/sbin/nologin --create-home --home-dir /app app
+
+COPY . /app
```
escapewindow, line 6: Hm, does this work? I'm guessing we're going to copy, e.g. `treescript/.` in as `/app/.` However, treescript requires `../scriptworker_client` (iscript requires that, plus `../vendored/mozbuild`) to install. We either need to also copy those directories in, or we need to populate the wheels somewhere.

escapewindow (approved):
Thank you! Looks like we need to standardize init.sh and fix the dockerfile.
At some point it would be best if we only trigger tasks if a given project has changed (e.g., if only treescript changes, don't trigger iscript tasks), but that may require some ci-taskgraph changes first.

docker.d/init.sh
garbas: this is exactly the reason why i would like to move all of the signing script as soon as possible. I'll update the init.sh with that version.

docker.d/init.sh
garbas: i'll chat with catlee and mihai to see what is the latest state with this and implement it here as well.

treescript/Dockerfile
garbas: it doesn't work. for now (in this PR) i'm ignoring it since i'm only replacing that tox is running on taskcluster and not on travis.

PR #25, opened 2019-10-07
Migrate beetmoverscript

garbas:


MihaiTabara:
@garbas we need to green up the build-docker-image before merging this https://tools.taskcluster.net/groups/OUjqd8KYTC2GDP1UeANdkg/tasks/CYsXpMOgSNafYmn8ydyvSg/details

.taskcluster.yml
```diff
@@ -10,13 +10,15 @@ tasks:
...
-      - ['treescript',    '37',             ''] # mozilla/releng-treescript'] # until we figure out how to install it
+      # [ <PROJECT NAME>,   <PYTHON VERSION>, <DOCKERHUB REPO>]
+      - ['client',          '36',             '']
+      - ['client',          '37',             '']
+      - ['iscript',         '36',             '']
+      - ['iscript',         '37',             '']
+      - ['treescript',      '36',             '']
+      - ['treescript',      '37',             ''] # mozilla/releng-treescript'] # until we figure out how to install it
+      - ['beetmoverscript', '36',             '']
+      - ['beetmoverscript', '37',             mozilla/releng-beetmoverscript']
```
escapewindow: It looks like we need to rename this. `mozilla/releng_beetmoverscript`?

`Invalid repository name (releng-beetmoverscript'), only [a-z0-9-_.] are allowed`

escapewindow (approved):
Thanks! Same comment as balrogscript - we should probably file issues about followups needed.

.taskcluster.yml
escapewindow: Hm, maybe it's complaining about the unpaired `'`. We either need to add the missing `'` before `mozilla`, or we need to remove the trailing `'`.

garbas: ah, that was a typo. i missed one initial `'`

PR #28, opened 2019-10-08
Migrate remaining scriptworkers

garbas:
- addonscript
- bouncerscript
- pushsnapscript
- pushapkscript
- shipitscript
- signingscript
- treescript


shipitscript/docker.d/init_worker.sh
```diff
@@ -1,5 +1,14 @@
 #!/bin/bash
-set -e
+set -o errexit -o pipefail
+
+test_var_set() {
```
escapewindow, line 4: Hm, should we define this once in `init.sh` and use that? I'm under the impression init.sh calls init_worker.sh, so we have access to the functions in both files.

signingscript/Dockerfile
```diff
@@ -15,6 +15,7 @@ USER app
 WORKDIR /app
 
 RUN python -m venv /app \
+ && sed -i -e "s@ | base64 -d@@g" /app/docker.d/init.sh \
```
escapewindow, line 18: clever :)

.taskcluster.yml
```diff
@@ -144,7 +164,7 @@ tasks:
                     cp ../docker.d/* docker.d/ &&
                     ./docker.d/generate_version_json.sh &&
                     ./docker.d/build_image.sh /tmp/image.tar &&
-                    if [ "$push_docker_image" == "1" ]; then
+                    if [ "$PUSH_DOCKER_IMAGE" == "1" ]; then
```
MihaiTabara, line 167: Thanks for fixing this!

escapewindow (approved):
Thanks!

signingscript/Dockerfile
garbas: its just a temporary hack. you learn this with Nix :)

shipitscript/docker.d/init_worker.sh
garbas: i'm not sure, but i will check

MihaiTabara (approved):
We still need to address the `init.sh` special case for signingscript while my temporary hack is still in there.

signingscript/Dockerfile
MihaiTabara: Ah, never mind my comment below, I didn't refresh before commenting :)

PR #32, opened 2019-10-10
Bug 1587068 - remove ED25519 double-encoding.

MihaiTabara:
Prepping this patch up to merge at the same time with @oremj landing it on CloudOps side.

### DIFF
--- a/docker.d/init.sh
+++ b/docker.d/init.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/bin/bas
 set -o errexit -o pipefail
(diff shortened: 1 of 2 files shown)


catlee (approved):

garbas:
Also the sed command in signingscript/Dockerfile should be removed.

MihaiTabara:
> Also the sed command in signingscript/Dockerfile should be removed.

Good catch, I added that too.

garbas (approved):

PR #33, opened 2019-10-11
Use aiohttp for calling autograph

catlee:
Start using file objects to represent the request instead of storing it
in memory.

This is a big change. I'm happy to explain things that aren't clear and/or add comments as necessary.


signingscript/src/signingscript/script.py
```diff
@@ -66,6 +66,8 @@
                 all_signing_formats,
             )
 
+    async with aiohttp.ClientSession() as session:
+        context.session = session
```
bhearsum, line 73: It looks like this overwrites as existing `context.session` (set by https://github.com/mozilla-releng/scriptworker/blob/master/scriptworker/client.py#L203, which is already a `ClientSession`). It probably deserves a comment.

signingscript/src/signingscript/sign.py
```diff
@@ -1002,24 +1001,101 @@ def _owner_filter(tarinfo_obj):
...
+
+    resp = await session.post(
+        url,
+        data=request_body,
+        headers={"Authorization": auth_header, "Content-Type": content_type},
+    )
+    log.debug("Autograph response: %s", resp.status)
+    resp.raise_for_status()
+    # TODO: Write this out to temporary file?
+    return await resp.json()
```
bhearsum: Because the response may potentially be large enough to have an impact on memory usage?

signingscript/tests/integration/test_autograph.py
```diff
@@ -305,20 +307,23 @@ def _extract_apk_signature_algorithm(apk_path):
...
-                    "autograph",
-                ]
-            )
-        ]
-    }
-    context.task = _craft_task([file_name], signing_format=format)
-
-    await sign_file_with_autograph(context, apk_path, format)
-    assert _extract_apk_signature_algorithm(apk_path) == expected_algorithm
+    async with aiohttp.ClientSession()as session:
```
bhearsum: Looks like you're missing a space here. TIL this is not a syntax error :)

bhearsum (approved):

signingscript/src/signingscript/sign.py
catlee: Yes. I've added a comment explaining this more.

bhearsum:
Much cleaner!

MihaiTabara (approved):

PR #34, opened 2019-10-11
bug 1588081 - wip - notarize geckodriver

escapewindow:


iscript/src/iscript/mac.py
```diff
@@ -214,20 +214,6 @@ def _get_sign_command(identity, keychain):
             },
             retry_exceptions=(IScriptError,),
         )
-        env = deepcopy(os.environ)
```
escapewindow, line 217: we need to staple before we tar, so let's move tarring out of sign_geckodriver

iscript/src/iscript/mac.py
```diff
@@ -1308,17 +1294,47 @@ def get_notarization_status_from_log(log_path):
 
     """
     key_config = get_key_config(config, task, base_key="mac_config")
+    work_dir = config["work_dir"]
 
     all_paths = get_app_paths(config, task)
-    langpack_apps = filter_apps(all_paths, fmt="autograph_langpack")
```
escapewindow, line 1313: no langpacks in geckodriver. this was added afterwards, and afaict is extraneous.

iscript/tests/test_mac.py
```diff
@@ -1233,19 +1232,9 @@ def fake_get_notarization_status_from_log(path):
...
     }
-    if use_langpack:
-        mocker.patch.object(mac, "sign_langpacks", new=noop_async)
-        task["payload"]["upstreamArtifacts"].append(
-            {
-                "taskId": "task3",
-                "formats": ["autograph_langpack"],
-                "paths": ["public/build3/target.langpack.xpi"],
-            }
-        )
```
escapewindow, line 1244: we probably have to mock the notarization calls to get these tests to pass. we also need to make sure this works irl.

PR #36, opened 2019-10-18
Use `openssl` to generate random password

rail:
For some reason Kubernetes fails generating a random password using the current approach. Could be related to the fact that we try to use `/dev/urandom`. Let's use `openssl` instead.


MihaiTabara (approved):
Works like a charm in dev and it's also the only place where we use this. Let's 🚢 it! 
Thanks for fixing it

PR #37, opened 2019-10-22
Use async requests and temporary files for signing

catlee:
Also: drop py3.6, add 3.8 for signingscript


signingscript/requirements/base.py36.txt
```diff
@@ -1,272 +0,0 @@
-# SHA1:e2e6bd7fcef883d3413ffa98e29a0ac85688d400
```
escapewindow, line 1: Should we create `{base,test}.py38.txt` files? They may diverge from py37 at some point. I suppose it depends on what's using them.

signingscript/setup.py
```diff
@@ -4,10 +4,16 @@
 with open(os.path.join(os.path.abspath(os.path.dirname(__file__)), "version.txt")) as f:
     version = f.read().rstrip()
 
-with open(
-    os.path.join(os.path.abspath(os.path.dirname(__file__)), "requirements/base.in")
-) as f:
-    install_requires = f.readlines()
+install_requires = [
```
escapewindow, line 7: Hm, did I see something's broken if we don't do this?
If so, what's broken? Do we need to package `requirements/base.in` ?

signingscript/src/signingscript/sign.py
```diff
@@ -11,19 +11,19 @@
...
-# TODO: Use aiohttp for this.
-import requests
 import shutil
 import subprocess
 import sys
 import tarfile
 import tempfile
 import zipfile
 
+from io import BytesIO
```
escapewindow, line 23: Did you figure out what was hanging? New tempfile per retry?

escapewindow (approved):

signingscript/requirements/base.py36.txt
catlee: not until we need it

signingscript/setup.py
catlee: This breaks if you want to add URLs into `requirements/base.in`, e.g. to a URL on github for testing a prerelease version of one of the dependencies.

signingscript/src/signingscript/sign.py
catlee: threads deadlocking. I was using a thread to call winsign.sign.sign_file, and aiohttp also uses threads for reading from the temporary files. With so many concurrent requests, the threadpool wasn't large enough, and so deadlocked.

The solution was to make sign_file async so I don't need the first level of threads.

PR #38, opened 2019-10-23
bug 1573529 - send a generic keyid for autograph widevine signing

escapewindow:
Looks like using `widevine_dep1_genericrsa` worked.


escapewindow:
Looks like we want to move the keyids into either passwords.json or into the format.

escapewindow:
Per @g-k , we no longer need this (the configs will roll out behind the scenes).

PR #43, opened 2019-10-28
Support authenticode_comment

Callek:


signingscript/src/signingscript/script.py
```diff
@@ -76,7 +76,7 @@
             copy_to_dir(path_dict["full_path"], context.config["work_dir"], target=path)
             log.info("signing %s", path)
             output_files = await sign(
-                context, os.path.join(work_dir, path), path_dict["formats"]
+                context, os.path.join(work_dir, path), path_dict["formats"], comment=path_dict.get('comment')
```
escapewindow: Hm. I don't love how we're surfacing authenticode-specific args in generic functions.
Would something like this work?

```
    full_path = path_dict.pop("full_path")
    signing_formats = path_dict.pop("formats")
    ...
    output_files = await sign(context, os.path.join(work_dir, path), signing_formats, **path_dict)

async def sign(context, path, signing_formats, **kwargs):
    ...
    for fmt in signing_formats:
        ...
        output = await signing_func(context, output, fmt, **kwargs)
```

This basically passes on anything that's in `path_dict` to the `signing_func` without having to know any details. We're already checking to see if `authenticode` is in the format before adding `comment` to `path_dict` in `build_filelist_dict`, so we don't necessarily need to check for it again in `sign`. Also, this allows for adding additional kwargs to other formats later on without extending this `signing_func` line any further.

We could keep the `comment=None` in `sign_authenticode`, which would cause us to error out if we send an unexpected kwarg that way.

signingscript/src/signingscript/sign.py
```diff
@@ -1551,13 +1554,20 @@ def verify_mar_signature(cert_type, fmt, mar, keyid=None):
     else:
         crosscert = None
 
+    if comment and orig_path.endswith('.msi'):
+        log.info(f"Using comment '{comment}' to sign {orig_path}")
+    elif comment:
+        log.info("Not using specified comment to sign {orig_path}, not yet implemented.")
```
escapewindow: Guessing you were intending to use an f-string here?

signingscript/src/signingscript/task.py
```diff
@@ -195,7 +195,7 @@ def _check_scopes_exist_and_all_have_the_same_prefix(scopes, prefixes):
 
 
 # sign {{{1
-async def sign(context, path, signing_formats):
+async def sign(context, path, signing_formats, comment=None):
```
escapewindow: If we go with the above suggestion, this would be `async def sign(context, path, signing_formats, **kwargs):`

signingscript/src/signingscript/task.py
```diff
@@ -217,7 +217,10 @@ def _check_scopes_exist_and_all_have_the_same_prefix(scopes, prefixes):
         except OSError:
             size = "??"
         log.info("sign(): Signing %s bytes in %s with %s...", size, output, fmt)
-        output = await signing_func(context, output, fmt)
+        kwargs = {}
+        if comment and "authenticode" in fmt:
+            kwargs['comment'] = comment
```
escapewindow: And we could get rid of these three lines.

signingscript/src/signingscript/task.py
```diff
@@ -301,6 +316,13 @@ def build_filelist_dict(context):
                 "full_path": full_path,
                 "formats": _sort_formats(artifact_dict["formats"]),
             }
+            if authenticode_comment and 'msi' in path:
+                if isinstance(authenticode_comment, type({})):
+                    if path in authenticode_comment:
+                        filelist_dict[path]["comment"] = authenticode_comment[path]
+                else:
+                        filelist_dict[path]["comment"] = authenticode_comment
+
```
escapewindow, line 227: We're getting to a level of complexity where we may want to call a helper function to build `filelist_dict[path]`.

escapewindow (approved):
@catlee Do you have any concerns here? Looks like the only conflicts are in `test_script.py`.

signingscript/src/signingscript/script.py
Callek: I think this could cause problems in the cases where we have one file with multiple signing formats...
```
      "formats": [
        "autograph_authenticode",
        "autograph_widevine",
        "autograph_omnija"
      ],
      "paths": [
        "public/build/target.zip"
      ],
      "taskId": "cJEuj31sQk29jEASng_e5w",
      "taskType": "build"
```
(from https://tools.taskcluster.net/groups/C_rQ8yseSTun4r6QRfGDsQ/tasks/HuHr-aB9QMCR-qVE-9LGUQ/details)

@escapewindow thoughts?

signingscript/src/signingscript/sign.py
Callek: Good catch, yes.

signingscript/src/signingscript/script.py
escapewindow: True. We could make it `authenticode_comment` and take `**kwargs` in all signing functions, only paying attention to the kwargs that each cares about. I'm not sure if that's a win or not.

signingscript/src/signingscript/sign.py
catlee: In general we should use %s for logging. e.g.
```log.info("Using comment %s to sign %s", comment, orig_path)```

This avoids problems where the f-string evalutation fails and your logging call ends up raising an exception and causing your program to crash.

File: signingscript/tox.ini [hunk 6]
```diff
@@ -21,7 +21,7 @@ deps=
     py37: -rrequirements/test.txt
 
 commands =
-    py.test --cov-config .coveragerc --cov=signingscript --cov-append --cov-report term-missing {posargs}
+    py.test --cov-config .coveragerc --cov=signingscript --cov-append --cov-report term-missing {posargs} --pdb
```
catlee: is this change intentional?

catlee: also, for your local use, you can do the equivalent by running `tox -e py37 -- --pdb`

Callek: this no longer seems to work locally, fwiw.  (I tried `tox -e py38 -- --pdb` and it tried to run a pdb command and nothing more)

escapewindow: IIRC, tox.ini changed when Ben landed the standardization patches?

Callek: No worry in the end, here, was just a passing comment

signingscript/src/signingscript/task.py
```diff
@@ -209,11 +210,25 @@ def build_filelist_dict(context):
...
+        if authenticode_comment and not any(path.endswith(".msi") for path in artifact_dict["paths"]):
+            # Don't have to think about .zip and such unpacking for the comment
+            raise TaskVerificationError("There is no support for authenticode_comment outside of msi's at this time")
         for path in artifact_dict["paths"]:
             full_path = os.path.join(context.config["work_dir"], "cot", artifact_dict["taskId"], path)
             if not os.path.exists(full_path):
                 messages.append("{} doesn't exist!".format(full_path))
             filelist_dict[path] = {"full_path": full_path, "formats": _sort_formats(artifact_dict["formats"])}
+            if authenticode_comment and "msi" in path:
```
escapewindow: I'm wondering if we can simply accept `authenticode_comment`, and pass it down, since `sign_authenticode_file` already checks to make sure the file is an msi.

Could we simply the changes to this function to just pass down the `authenticode_comment`, and not have to worry about `msi` or `path` ? Would we ever need to support more than 1 msi file in an `upstreamArtifact`, where we want to supply different `authenticode_comment`s per msi? If so, would it be bad to require another `upstreamArtifact` dict? In that case, `authenticode_comment` could always be a string.

escapewindow (approved):
I think this is good. I'm wondering if the requirements are complicating things, or if we're complicating things without needing to. (Re: all the conditional `authenticode_comment` checks, plus allowing for a dictionary or string.) I'm not familiar with the requirements here, but unless the requirements are complex, let's not add an overly complex solution.

signingscript/src/signingscript/sign.py
```diff
@@ -1306,7 +1309,15 @@ def verify_mar_signature(cert_type, fmt, mar, keyid=None):
     else:
         crosscert = None
 
-    if not await winsign.sign.sign_file(infile, outfile, digest_algo, certs, signer, url=url, crosscert=crosscert, timestamp_style=timestamp_style):
+    if authenticode_comment and orig_path.endswith(".msi"):
+        log.info("Using comment '%s' to sign %s", authenticode_comment, orig_path)
+    elif authenticode_comment:
+        log.info("Not using specified comment to sign %s, not yet implemented.", orig_path)
```
catlee: can you change this to say "not yet implemented for non-msi files" or something like that?

signingscript/src/signingscript/task.py
Callek: Yea, this was a premature complexity, I think.  Nothing needs it.

PR #47, opened 2019-11-01
Bump scriptworker to 27.2.0.

tomprince:


MihaiTabara (approved):
Thanks for updating these!

Do we want to update https://github.com/mozilla-releng/scriptworker-scripts/blob/master/maintenance/pin-helper.sh#L8 to encompass all the scripts?

PR #48, opened 2019-11-01
Pin all scripts.

tomprince:


maintenance/pin-helper.sh
```diff
@@ -5,7 +5,22 @@ SUFFIX=${SUFFIX:-txt}
...
+
+for dir in \
+	addonscript \
+	balrogscript \
+	beetmoverscript \
+	bouncerscript \
+	iscript \
+	pushapkscript \
+	pushsnapscript \
+	scriptworker_client \
```
escapewindow, line 19: At some point, we may want this pinned before the others, but I don't think that's necessary yet.

escapewindow (approved):

MihaiTabara:
Thanks for taking care of this @escapewindow to make our lives easier in the future when we want to update deps at bulk 👍

escapewindow:
Thanks Tom for the patch!

PR #51, opened 2019-11-04
Bug 1588590 - fix GCP mobile worker configs for prod/dev

MihaiTabara:
This is to match the `script_config.json` we generate on puppet [here](https://github.com/mozilla-releng/build-puppet/blob/master/modules/pushapk_scriptworker/manifests/settings.pp)

### DIFF
--- a/pushapkscript/docker.d/worker.yml
+++ b/pushapkscript/docker.d/worker.yml
@@ -8,6 +8,7 @@ taskcluster_scope_prefixes:
     - "project:mobile:focus:releng:googleplay:product:"
     - "project:mobile:reference-browser:releng:googleplay:product:"
     - "project:mobile:fenix:releng:googleplay:product:"
+    - "project:mobile:firefox-tv:releng:googleplay:product:"
   else:
     - "project:releng:googleplay:"
 products:
@@ -47,6 +48,7 @@ products:
           apps:
             dep:
               service_account: dummy
+              default_track: beta
               package_names:
                 - "org.mozilla.fennec_aurora"
               credentials_file: { "$eval": "GOOGLE_CREDENTIALS_FIREFOX_DEP_PATH" }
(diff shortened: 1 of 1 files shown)


pushapkscript/docker.d/worker.yml
```diff
@@ -97,21 +102,34 @@ products:
...
+          skip_checks_fennec: true
+          skip_check_signature: true
+          skip_check_ordered_version_codes: true
+          override_channel_model: "single_google_app"
+          apps:
+            production:
+              package_names:
+                - "org.mozilla.tv.firefox"
+              amazon:
+                client_id: 'dummy'
```
MihaiTabara, line 122: I'll amend later this week these to reflect the correct env var once I sent it out to Jeremy and update cloudops-infra repo too.

MihaiTabara: Just to clarify - we're not deploying this to `firefox-tv` repo until we add these missing credentials.

pushapkscript/docker.d/worker.yml
```diff
@@ -97,21 +102,34 @@ products:
...
             certificate_alias: "reference-browser"
-
+        - product_names: [ "firefox-tv" ]
+          skip_check_multiple_locales: true
+          skip_check_same_locales: true
+          skip_checks_fennec: true
+          skip_check_signature: true
+          skip_check_ordered_version_codes: true
+          override_channel_model: "single_google_app"
+          apps:
```
mitchhentges, line 117: If the channel model for the dep `firefox-tv` is `single_google_app`, then this property should change from `apps` to `app` and it should be more google-specific, like [this one](https://github.com/mozilla-releng/scriptworker-scripts/pull/51/files#diff-3ab502ad0f8a21fc4b9a368b6cbcbd8eR105)

mitchhentges (changes requested):

pushapkscript/docker.d/worker.yml
MihaiTabara: Hm, dep `firefox-tv` has no `single_google_app` channel mode.

MihaiTabara: I copy-pasted from puppet, hm.

mitchhentges: I think that puppet is wrong. You can confirm this by looking at [how `pushapkscript` parses the publish config](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/pushapkscript/pushapkscript/publish_config.py#L21) (note how it looks for an `app` property, not `apps`)

mitchhentges (approved):
If this is the same as `puppet`, feel free to land. I'll update both GCP and puppet

mitchhentges (changes requested):
Are you sure you grabbed that from the up-to-date puppet configs?
Looking at the `mobile-dep` config for `firefox-tv`, its channel model [isn't `single_google_app`](https://github.com/mozilla-releng/build-puppet/blob/master/modules/pushapk_scriptworker/manifests/settings.pp#L200). The puppet config should be correct

pushapkscript/docker.d/worker.yml
mitchhentges: I misunderstood, the dep `firefox-tv` config hasn't had `override_channel_model: "single_google_app"`, and should continue to not have that model. For some reason, I assumed that dep `firefox-tv` was like dep `reference-browser` and dep `fenix` in having that single dummy app.

Remove line 117 and this should be good to land!

mitchhentges (approved):
Mastapeece

PR #54, opened 2019-11-06
Drop Python 3.6, use 3.8 instead

rail:


escapewindow:
<s>We may want to run our unit tests on py38 before we run it in production :)</s>

Ah, you're adding 3.8, but running on 3.7. n/m :)

rail:
Time to rebase

MihaiTabara:
\o/

escapewindow (approved):
We could add the py37 and py38 support strings into setup.py, but that might not be necessary since we're not pushing these to pypi.

Looks good, 🚢 
Thanks!

PR #60, opened 2019-11-12
Fix Thunderbird Auth0 audience

rail:


bhearsum (approved):
I am unable to verify that these are the correct values, but it's not going to make anything worse.

rail:
FTR, I ran a `curl` test from https://auth0.com/docs/api/authentication?shell#client-credentials-flow, it worked.

PR #63, opened 2019-11-13
Re-land treescript fixes from #9

escapewindow:
We backed these out for GCP automation expediency; let's rebase, fix, reland.

PR #64, opened 2019-11-14
Import configloader

rail:
This should remove another temporary external dependency.

I didn't add coverage reporting, because for some reason it reports 0%, maybe because of the way how click.testing wraps things around. ¯\_(ツ)_/¯

Not sure if `/app/docker.d/configloader` is a good place to copy the package to.


escapewindow:
Do you want to take the approach in #63 instead? We're going to also have to install `../scriptworker_client`, `../vendored/mozbuild`, and possibly `../mozapkpublisher` in various Dockerfiles down the line, so moving towards installing from the root of the repo instead of copying the source into `docker.d` makes sense to me. What do you think?

rail:
Oooh, this is much better. I'll wait until it's landed and update this PR.

escapewindow (approved):

JohanLorenzo:
Hey there! I suspect this PR caused https://bugzilla.mozilla.org/show_bug.cgi?id=1600787#c10. Let's continue the discussion on Bugzilla 🙂

PR #65, opened 2019-11-14
support xpi signing

escapewindow:
Rebased on top of master.

This uses the format name as the keyid. I can change it up to use a generalized name as the keyid, and specify the keyid in the passwords file, if that's preferred.

It looks like I can't test this in puppet-land, because we dropped py36 support. I'm guessing we should eyeball it, roll it out to a new k8s pod, and test then.

### DIFF
--- a/signingscript/src/signingscript/sign.py
+++ b/signingscript/src/signingscript/sign.py
@@ -287,10 +287,6 @@ async def sign_xpi(context, orig_path, fmt):
     cert_type = task.task_cert_type(context)
     file_base, file_extension = os.path.splitext(orig_path)
 
-    a = get_autograph_config(
-        context.autograph_configs, cert_type, [fmt], raise_on_empty=True
-    )
-
     if file_extension not in (".xpi", ".zip"):
         raise SigningScriptError("Expected a .xpi")
 
--- a/signingscript/src/signingscript/sign.py
+++ b/signingscript/src/signingscript/sign.py
@@ -286,7 +286,7 @@ async def sign_xpi(context, orig_path, fmt):
     """
     file_base, file_extension = os.path.splitext(orig_path)
 
-    if file_extension != ".xpi":
+    if file_extension not in (".xpi", ".zip"):
         raise SigningScriptError("Expected a .xpi")
 
     ext_id = _extension_id(orig_path, fmt)
@@ -617,6 +617,7 @@ def _extension_id(filename, fmt):
         "languages" in manifest
         and "langpack_id" in manifest
         and LANGPACK_RE.match(manifest["applications"]["gecko"]["id"])
+        and filename.endswith(".xpi")
     ):
         raise SigningScriptError("{} is not a valid langpack".format(filename))
     return manifest["applications"]["gecko"]["id"]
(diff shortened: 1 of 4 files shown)


escapewindow:
Failing check: looks like an intermittent treescript docker bustage. Rerunning.

signingscript/src/signingscript/sign.py
```diff
@@ -286,13 +286,16 @@ def get_autograph_config(
...
-    if not file_extension == ".xpi":
+    if file_extension not in (".xpi", ".zip"):
         raise SigningScriptError("Expected a .xpi")
 
-    id = _langpack_id(orig_path)
+    ext_id = _extension_id(orig_path, fmt)
     log.info("Identified {} as extension id: {}".format(orig_path, id))
+    kwargs = {'extension_id': ext_id}
+    if 'langpack' not in fmt:
+        kwargs['keyid'] = fmt
```
catlee: Let's make this explicit in the passwords file.

Callek (approved):
This likely has some conflicts with one of my local patches (or PRs) but I can rebase over this pretty easily.

Let's still address catlee's comment.

signingscript/src/signingscript/sign.py
escapewindow: Ok. I think the latest commit does that.

signingscript/src/signingscript/sign.py
```diff
@@ -286,13 +286,14 @@ def get_autograph_config(
...
+    if file_extension not in (".xpi", ".zip"):
         raise SigningScriptError("Expected a .xpi")
 
-    id = _langpack_id(orig_path)
+    ext_id = _extension_id(orig_path, fmt)
     log.info("Identified {} as extension id: {}".format(orig_path, id))
+    kwargs = {'extension_id': ext_id}
     # Sign the appropriate inner files
-    await sign_file_with_autograph(context, orig_path, fmt, extension_id=id)
+    await sign_file_with_autograph(context, orig_path, fmt, **kwargs)
```
catlee, line 297: I think you'll need to grab the keyid from somewhere before calling sign_file_with_autograph? Or maybe modify sign_with_autograph to get the keyid from the server object if the keyid value passed into the function is None?

escapewindow: Ah, I thought that was already hooked up. Will do

escapewindow: Done. While getting 100% coverage I found some bugs; fixed.

catlee (approved):

signingscript/src/signingscript/sign.py
```diff
@@ -284,8 +284,13 @@ def get_autograph_config(
         str: the path to the signed xpi
 
     """
+    cert_type = task.task_cert_type(context)
     file_base, file_extension = os.path.splitext(orig_path)
 
+    a = get_autograph_config(
+        context.autograph_configs, cert_type, [fmt], raise_on_empty=True
+    )
+
```
catlee: what is this here for?

escapewindow: Looks like it's left over. Removed.

PR #66, opened 2019-11-14
Remove signing server support

catlee:


signingscript/Dockerfile
```diff
@@ -4,7 +4,7 @@ RUN groupadd --gid 10001 app && \
     useradd -g app --uid 10001 --shell /usr/sbin/nologin --create-home --home-dir /app app
 
 RUN apt-get update \
- && apt-get install -y zipalign default-jdk-headless osslsigncode \
+ && apt-get install -y zipalign osslsigncode \
```
escapewindow, line 7: Hm, did we need this for jar signing before, and no longer?

signingscript/config_example.json
```diff
@@ -2,12 +2,11 @@
     "signing_server_config": "server_config.json",
     "work_dir": "/tmp/work_dir",
     "artifact_dir": "/tmp/artifact_dir",
-    "signtool": "signtool",
```
escapewindow, line 5: \o/

escapewindow: We may be able to delete `ssl_cert` as well; I think that's for the signing servers.

escapewindow (approved):

signingscript/Dockerfile
catlee: I think maybe we used it for zipalign before it was available as its own package?

signingscript/config_example.json
catlee: Good catch! I've removed ssl_cert in a followup.

PR #67, opened 2019-11-15
Adds config schema and tests for pushapkscript

mitchhentges:
This is the first PR for [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1591594)


mitchhentges:
I'm going to add a schema for each script and do a single large PR

PR #69, opened 2019-11-15
Bug 1596470 - add missing Fennec prod signing for Fenix

MihaiTabara:
Turns out we missed porting https://github.com/mozilla-releng/build-puppet/blob/master/modules/signing_scriptworker/templates/passwords-mobile.json.erb#L14 over.

### DIFF
--- a/signingscript/docker.d/passwords.yml
+++ b/signingscript/docker.d/passwords.yml
@@ -233,6 +233,13 @@ in:
              ["autograph_apk"],
              "autograph"
             ]
+        project:mobile:fenix:releng:signing:cert:fennec-production-signing:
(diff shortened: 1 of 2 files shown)


signingscript/docker.d/passwords.yml
```diff
@@ -233,6 +233,13 @@ in:
              ["autograph_apk"],
              "autograph"
             ]
+        project:mobile:fenix:releng:signing:cert:fennec-production-signing:
```
rail, line 236: colons in bare strings make me nervous, but yaml knows better how to not parse them :D

rail (approved):

PR #77, opened 2019-11-18
fix version.json

escapewindow:


Callek (approved):

docker.d/generate_version_json.sh
```diff
@@ -16,4 +16,4 @@ cat > ${PROJECT_NAME}/version.json <<EOF
 }
 EOF
 
-cat ${PROJECT_NAME}/version.json
+cat version.json
```
MihaiTabara, line 19: I see we had `./version.json` before in https://github.com/mozilla-releng/scriptworker-scripts/pull/63/files#diff-c457f32136023804d3c5deaa91cfce98, but should be the same thing considering line 10.

MihaiTabara (approved):

PR #78, opened 2019-11-18
Adds config schema and tests for docker

mitchhentges:
This is for [bug 1591594](https://bugzilla.mozilla.org/show_bug.cgi?id=1591594).

I'm creating a PR before this is done so I can verify that CI is still happy as I make changes

## Merge checklist
* [x] pushapk
* [x] signing
* [x] addon
* [x] balrog
* [x] beetmover
* [x] bouncer
* [x] ~iscript~ (not set up with GCP yet)
* [x] pushsnap
* [x] shipit
* [x] tree


lundjordan:
I talked to @mitchhentges and he thinks it just needs review + land. Let's see what the review digs up and then go from there.

escapewindow:
This happened pre- #95 and #109 ; unbitrotting.

escapewindow:
This PR will add another layer of env var / schema checking for our docker deploys. It doesn't check SOPS currently, though we could potentially add that.

Ideally this adds a layer of checks, so when we change docker.d entries, we also have to think about where else to add or remove those variables. If this PR succeeds in doing that, awesome. If it adds another location to make changes, but doesn't remind us to update other locations, then it might not be as useful. @rail @JohanLorenzo @MihaiTabara what do you think?

rail:
This is much better  than we have right now. It'll create some burden, but hey that's the main idea. :)

One thing that comes to mind is that we try to cover all cases by running tests for prod, fake-prod, etc, but the list may be not exhaustive and doesn't cover all `match` cases.

BTW, should we also run validation on scriptworker start so we don't get runtime errors?

TBH, I would like to revisit the way how we generate the configs. There is a bit of boilerplate everywhere and it's a combination of Bash and Python (json, yaml, json-e). Maybe rewrite `init.sh` in python or/and use Dhall instead of json-e templates?
Well, this is a separate issue. :)

escapewindow:
> This is much better than we have right now. It'll create some burden, but hey that's the main idea. :)

Ok! I'll merge, and we can iterate. Then when we get to it, we can refactor our configs and tests to better match our needs.

> One thing that comes to mind is that we try to cover all cases by running tests for prod, fake-prod, etc, but the list may be not exhaustive and doesn't cover all `match` cases.

Someone should invent a json-e coverage tool :)

> BTW, should we also run validation on scriptworker start so we don't get runtime errors?

Probably.

> TBH, I would like to revisit the way how we generate the configs. There is a bit of boilerplate everywhere and it's a combination of Bash and Python (json, yaml, json-e). Maybe rewrite `init.sh` in python or/and use Dhall instead of json-e templates?
> Well, this is a separate issue. :)

Could do, though I'm sure there will be some debate around the specifics.

escapewindow (approved):

tomprince:
> > TBH, I would like to revisit the way how we generate the configs. There is a bit of boilerplate everywhere and it's a combination of Bash and Python (json, yaml, json-e). Maybe rewrite `init.sh` in python or/and use Dhall instead of json-e templates?
> > Well, this is a separate issue. :)
> 
> Could do, though I'm sure there will be some debate around the specifics.

Now that (almost) all the scriptworkers are running in docker, and the configuration is passed via ENV variables, we should consider pulling the relevant configuration directly from them, rather than having an intermediate step that generates a config file before running scriptworker.

escapewindow:
+1 to considering that. In some cases, the config is fairly complex, and embedding it all into env vars isn't great. We could possibly do something halfway in between, which could involve python- or otherwise computed config files with `os.environ["ENV_VAR"]` as the value.

PR #79, opened 2019-11-19
Do push checks

Callek:


shipitscript/shipitscript/script.py
```diff
@@ -61,8 +62,8 @@ def create_new_release_action(context):
         # TODO quit early, mark task as green though
         pass
     log.info('Determining most recent shippable revision')
-    shippable_revision = ship_actions.get_shippable_revision(
-        repo, last_shipped_revision
+    should_build = ship_actions.should_ship_revision(
```
lundjordan: we need to use the revision below. see how `shippable_revision` is used.

lundjordan (changes requested):

shipitscript/shipitscript/pushlog_scan.py
```diff
@@ -0,0 +1,98 @@
+from enum import Enum, auto, unique
+
+import requests
+
+URL = 'https://hg.mozilla.org/{repo}/json-pushes'
```
lundjordan: hm, will `repo` contain `releases/`?

shipitscript/shipitscript/pushlog_scan.py
```diff
@@ -0,0 +1,98 @@
...
+
+
+def push_check(func):
+    _push_checks.append(func)
+    return func
+
+
+def should_build(repo, old_rev, curr_rev):
+    # TODO redo
+    resp = requests.get(
```
lundjordan, line 34: should we be using aiohttp if this is to be used by shipitscript and scriptworker? Both of which are async right?

shipitscript/shipitscript/pushlog_scan.py
```diff
@@ -0,0 +1,98 @@
...
+            return isimportant
+    # We reached the end of our checks and
+    # we did not explicitly detect important
+    # nor unimportant.
+    # ...Our default is important
+    return Importance.IMPORTANT
+
+
+@push_check
+def is_l10n_bump(push):
```
lundjordan, line 79: can this whole method be:

```
    if push['changesets'][-1]['author'] == 'L10n Bumper Bot <release+l10nbumper@mozilla.com>':
        # This is the bumper bot, so important
        return Importance.IMPORTANT
    return Importance.MAYBE
```

shipitscript/shipitscript/pushlog_scan.py
```diff
@@ -0,0 +1,98 @@
...
+    Treat a=test-only (or a=testonly) as unimportant if present on the tip of a push.
+    """
+    # Get the tip
+    cset = push['changesets'][-1]
+    if 'a=test-only' in cset['desc'] or 'a=testonly' in cset['desc']:
+        return Importance.UNIMPORTANT
+
+
+@push_check
+def skip_version_bump(push):
```
lundjordan, line 105: similar to bumper bot push_check, do we need `if not push['user'] == 'ffxbld'` ?

shipitscript/shipitscript/pushlog_scan.py
```diff
@@ -0,0 +1,98 @@
...
+    isimportant = Importance.MAYBE
+    for check in _push_checks:
+        isimportant = check(push)
+        if not isimportant == Importance.MAYBE:
+            return isimportant
+    # We reached the end of our checks and
+    # we did not explicitly detect important
+    # nor unimportant.
+    # ...Our default is important
+    return Importance.IMPORTANT
```
lundjordan, line 65: should we log when we end up with a `Importance.MAYBE`?

also, should we initialize `isimportant` as Importance.IMPORTANT above if that's the intended default?

shipitscript/shipitscript/script.py
```diff
@@ -61,15 +62,15 @@ def create_new_release_action(context):
...
+        repo, last_shipped_revision, current_revision,
     )
-    if not shippable_revision:
+    if not should_build:
         # TODO quit early, mark task as green though
         pass
```
lundjordan: we need to do something here.

shipitscript/shipitscript/script.py
```diff
@@ -61,15 +62,15 @@ def create_new_release_action(context):
...
     log.info('create a new release')
     ship_actions.create_new_release(
-        product, repo, channel, next_version, shippable_revision, phase, shipit_config
+        product, repo, channel, next_version, current_revision, phase, shipit_config
```
lundjordan: so I'm still not sure this will work. We want to tell `create_new_release` what revision to ship from. This just takes the most tip revision every time. I get that if we have a commit in the pushlog of commits that is "important" we want to trigger a new release, but should we always take the tip rev? What if there are no builds on that rev? e.g. `NPOTB DONTBUILD`?

lundjordan (commented):
Looks good! comments below.

MihaiTabara:
@Callek FYI - some of these changes may conflict with https://github.com/mozilla-releng/scriptworker-scripts/pull/93/, specifically, the payload that we're expecting from the in-tree tasks. Do you still need `repo` and if yes, what should this do? I've updated the schema https://github.com/mozilla-releng/scriptworker-scripts/pull/93/files#diff-d067378571c66b6fc5e54677a27c3fcf

shipitscript/shipitscript/pushlog_scan.py
Callek: shipitscript uses requests in many places, decided to stick with it here (for now) -- but yes we should refactor to use async functions for this stuff.

shipitscript/shipitscript/pushlog_scan.py
Callek: adjusting so its `branch` not `repo`

MihaiTabara: Yeah, I confirm, `branch` will contain the prefix. We'll have `try` or `releases/mozilla-beta` or `projects/maple` baked in `branch`.

shipitscript/shipitscript/pushlog_scan.py
MihaiTabara: addressed them both.

shipitscript/shipitscript/pushlog_scan.py
MihaiTabara: to be addressed in a follow-up. I'll open some issues for this repo once this is over.

shipitscript/shipitscript/pushlog_scan.py
MihaiTabara: Had similar thoughts when I read this, addressed in a follow-up.

shipitscript/shipitscript/pushlog_scan.py
MihaiTabara: I believe it does.

shipitscript/shipitscript/pushlog_scan.py
MihaiTabara: I think the answer is yes because l10n bumps such as [this](https://hg.mozilla.org/releases/mozilla-beta/rev/192f60968befcde3d3174dc16a75b5ec8713fe5d) do have the user set to`ffxbld`. So it'll get to Maybe anyway, unless it's a bump in which case it'll be IMPORTANT.

shipitscript/shipitscript/script.py
MihaiTabara: done

shipitscript/shipitscript/script.py
MihaiTabara: rebased against my changes which already take care of this.

shipitscript/shipitscript/script.py
MihaiTabara: It's now taking the `shippable_revision` or it's No-op and task goes green.

shipitscript/shipitscript/pushlog_scan.py
```diff
@@ -0,0 +1,105 @@
...
+    """
+    Run through all `@push_check`s to find if this `push` is important or not.
+
+    See the help on `Importance` for more information.
+    """
+    isimportant = Importance.IMPORTANT
+    for check in _push_checks:
+        isimportant = check(push)
+        if not isimportant == Importance.MAYBE:
+            log.info(f'{push} found as MAYBE in {check}')
```
lundjordan: we want to log this outside of this condition right? just before the bottom return statement?

shipitscript/shipitscript/pushlog_scan.py
```diff
@@ -0,0 +1,105 @@
...
+        URL.format(branch=branch),
+        params={'full': '1', 'fromchange': last_shipped_rev, 'tochange': cron_rev},
+    )
+    pushlog = resp.json()
+    reversed_pushes = [pushlog[k] for k in sorted(pushlog.keys(), reverse=True)]
+    for push in reversed_pushes:
+        push_importance = is_push_important(push)
+        if push_importance == Importance.IMPORTANT:
+            # Tell caller that we want to build.
+            return push['changesets'][-1]['node']
```
lundjordan, line 44: so `node` will get revision?

Also, we are still ignoring the case where we are returning a revision that includes DONTBUILD?

lundjordan: I'm okay with this being a follow up. As per slack.

shipitscript/shipitscript/pushlog_scan.py
MihaiTabara: Moved.

shipitscript/shipitscript/pushlog_scan.py
MihaiTabara: Added another decorator, please let me know if it looks good.

lundjordan (approved):

PR #80, opened 2019-11-19
Bug 1596439 - Update dependencies for scriptworker 29.0.1 

JohanLorenzo:
2 notable things in addition to the regular bump:
 * `importlib-metadata` is not a dependency anymore because `jsonschema` removed it for python 3.8+ https://github.com/Julian/jsonschema/pull/627. Thus, `zipp` and `more-itertools` are gone too. 
 * The major changes in treescript are caused by https://github.com/mozilla-releng/scriptworker-scripts/pull/63.


escapewindow:
Hm, I wonder if the treescript requirement changes means we won’t install scriptworker at all? In which case we may need to add scriptworker back in, even if we don’t import it.

rail (changes requested):
Yeah, this will break docker images and treescript won't work. Can you explicitly add scriptworker to treescript's base requirements?

JohanLorenzo:
Great catch @escapewindow! I added `scriptworker`  back. The diff now looks about the same as the other projects 🙂

escapewindow:
I didn't think about it til I saw your diff :) Sorry I removed it.

rail (approved):

treescript/requirements/base.txt
```diff
@@ -201,9 +205,9 @@ requests==2.22.0 \
     --hash=sha256:11e007a8a2aa0323f5a921e9e6a2d7e4e67d9877e85773fba9ba6419025cbeb4 \
     --hash=sha256:9cf5292fcd0f598c671cfc1e0d7d1a7f13bb8085e9a590f48c010551dc6c4b31 \
     # via github3.py, taskcluster
-scriptworker==29.0.0 \
-    --hash=sha256:32d06bedaa7f8134cdb6319ff4c1d044c03472ba3145b971989a52527858b274 \
-    --hash=sha256:91f999f3eb60f7fd43b5e8df3f260ed79d277a930823a33de3618692d90ba89b
+scriptworker==29.0.1 \
```
escapewindow, line 208: Hm. I need 29.0.2 on treescript. (I just released that for https://bugzilla.mozilla.org/show_bug.cgi?id=1597671 ).
We can push your change, and then I can push my change to production-treescript, or we can push once with 29.0.2. I'm running pin.sh right now.

PR #85, opened 2019-11-21
Fix l10n bump

escapewindow:
This PR:

- counts the number of changes we expect to push, and the number of changesets that are outgoing, and dies if they don't match. This is important, because on Tuesday we had an issue with l10n-bumper pushing all of autoland to central. I believe I've fixed that issue, but this check helps prevent something similar from happening in case of another bug.
- adds the `strip` extension. I believe we were saved from pushing extraneous revisions by using `mq`'s `strip`, but that would potentially strip way more than expected given all the code in mozilla-unified

The other side of the l10n-bump fix is in-tree.

I'm hoping to land this PR before GTB for Thursday's betas, so Just In Case the tagging or version bumping pushes too many commits, we catch that before we push.

### DIFF
--- a/treescript/src/treescript/l10n.py
+++ b/treescript/src/treescript/l10n.py
@@ -232,7 +232,7 @@ async def l10n_bump(config, task, repo_path):
     ignore_closed_tree = get_ignore_closed_tree(task)
     l10n_bump_info = get_l10n_bump_info(task)
     revision_info = None
-    changes = False
+    changes = 0
 
     if not ignore_closed_tree:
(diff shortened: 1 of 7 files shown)


MihaiTabara (commented):
Unrelatedly to the PR itself, I was wondering, whether we should add an entry in the [CHANGELOG](https://github.com/mozilla-releng/releasewarrior-2.0/blob/master/docs/CHANGELOG.md) to mention that we've switched the home of the l10n bumper.

Callek (approved):

PR #87, opened 2019-11-21
[treescript] Don't strip after running robustcheckout.

tomprince:
robustcheckout is desigend to be a resilient tool for ensuring that a checkout
matches exactly what is in an upstream repo, in a cache friendly way. We
don't need to duplicate that worker by strip+update+purge.

### DIFF
--- a/treescript/src/treescript/mercurial.py
+++ b/treescript/src/treescript/mercurial.py
@@ -202,7 +202,6 @@ async def checkout_repo(config, task, repo_path):
         branch,
         exception=CheckoutError,
     )
-    await strip_outgoing(config, task, repo_path)
 
 
 # do_tagging {{{1


escapewindow:
If we robustcheckout to `default`, can't that include previous commits we made locally?

tomprince:
> If we robustcheckout to `default`, can't that include previous commits we made locally?

No, `robustcheckout` resolves `default` in the remote repository, and checks out whatever that resolves to, whatever local meaning is applied to that is irrelevant. (See [here](https://hg.mozilla.org/hgcustom/version-control-tools/file/f824da4501069359b208312172f071a0b5927942/hgext/robustcheckout/__init__.py#l682) for the code that handles that.)

escapewindow (approved):

PR #90, opened 2019-11-22
Bug 1598525 - [treescript] Set upstream_repo from variable.

jfx2006:
The Verbump task fails during Thunderbird releases with "repository is
unreleated".
Set config['upstream_repo'] to comm-central for Thundrbird treescript
tasks.


rail (approved):

rail:
Thank you for the patch!

rail:
it should be live in a couple of hours

PR #93, opened 2019-11-27
Shipitscript changes to support automated betas

MihaiTabara:


shipitscript/shipitscript/shipitapi.py
```diff
@@ -104,13 +105,33 @@ def getRelease(self, name, headers={}):
...
             if resp:
                 log.error(resp.content)
-                log.error('Response code: %d', resp.status_code)
+                log.error(f'Response code: {resp.status_code}')
+            raise
+
+    def get_releases(self, product, branch, status, version='', headers={}):
+        """Method to map over the GET /releases List releases API in Ship-it"""
+        resp = None
+        most_recent_params = {'product': product, 'branch': branch, 'status': status}
```
rail: A nit: you can drop "most recent" from the variable name, the function is quite generic.

shipitscript/shipitscript/script.py
```diff
@@ -38,38 +37,37 @@ def mark_as_shipped_action(context):
...
-    next_version = ship_actions.get_next_release_version(
-        product, channel, shipit_config
+        shipit_config, product, branch,
     )
-    log.info('Ensuring next version is a new version and not a buildnum increment')
-    if get_buildnum_from_version(next_version) != 1:
-        # TODO quit early, mark task as green though
-        pass
+    if not last_shipped_revision:
+        log.info("No valid most recent shipped revisison found, silent exit ...")
```
rail: Hmm, if we don't have anything shipped for this branch product, probably something is broken under the sun. Maybe we should exit with an error here?

shipitscript/shipitscript/ship_actions.py
```diff
@@ -33,36 +33,83 @@ def get_shippable_revision(repo, last_shipped_revision):
...
+    log.info(f'Product {product} and {branch} is enabled. Continuing ...')
+    return False
+
+
+def start_new_release(shipit_config, product, branch, version, revision, phase):
+    # compute the build_number for the to-be-created release
+    build_number = calculate_build_number(shipit_config, product, branch, version)
+
+    # safeguard to avoid creating releases if they have been disabled from UI
+    if releases_are_disabled(shipit_config, product, branch):
```
rail, line 86: I'd move this before you calculate the build number, so you don't even waste time.

shipitscript/shipitscript/ship_actions.py
```diff
@@ -33,36 +33,83 @@ def get_shippable_revision(repo, last_shipped_revision):
...
+    release_details = release_api.create_new_release(
+        product, branch, version, build_number, revision, headers=headers
     )
 
+    # grab the release name from the Ship-it response
+    release_name = release_details['name']
+
+    # avoid race conditions in between creating the release and triggering the
+    # specific `phase`
+    if releases_are_disabled(shipit_config, product, branch):
```
rail, line 104: You never know, maybe they will be disabled after you check here and before you trigger a phase? :P

shipitscript/shipitscript/ship_actions.py
```diff
@@ -33,36 +33,83 @@ def get_shippable_revision(repo, last_shipped_revision):
...
+    except KeyError:
+        log.error('No `revision` key present in the most recent release')
+        return
+
+
+def calculate_build_number(shipit_config, product, branch, version):
     release_api, headers = get_shipit_api_instance(shipit_config)
 
+    log.info('Call Ship-it to retrieve all releases matching criteria ...')
+    all_releases = release_api.get_releases(
```
rail: I'd use `all_builds`, because it's limited to a particular version.

shipitscript/shipitscript/ship_actions.py
```diff
@@ -33,36 +33,83 @@ def get_shippable_revision(repo, last_shipped_revision):
...
     release_api, headers = get_shipit_api_instance(shipit_config)
-    release_name = ""
 
+    log.info('Call Ship-it to check for disabled products across branches')
+    disabled_products = release_api.get_disabled_products(headers=headers)
+
+    if product in disabled_products:
+        if branch in disabled_products[product]:
+            log.info(f'Product {product} and {branch} are currently disabled')
+            return True
```
rail: I'd just used something like `return branch in disabled_products.get(product, [])` unless you want extra logging.

shipitscript/shipitscript/shipitapi.py
```diff
@@ -123,32 +144,83 @@ def update_status(self, name, status, rebuild_product_details=True, headers={}):
...
+            )
+            return resp.json()
+        except Exception:
+            log.error('Caught error while creating the release', exc_info=True)
+            if resp:
+                log.error(resp.content)
+                log.error(f'Response code: {resp.status_code}')
+            raise
+
+    def trigger_release_phase(self, release_name, phase, headers={}):
```
rail, line 203: Aren't `headers` mandatory? Maybe don't use the default `{}` here?

rail (approved):
In overall it looks great! I have some nit picks. :)

shipitscript/shipitscript/shipitapi.py
MihaiTabara: Ah, thanks for catching this. I actually had a different function name before and then when I added the logic for build_number I realized I can re-use this one and I tweaked it a bit, forgetting of this variable name. I've addressed this in a follow-up.

shipitscript/shipitscript/shipitapi.py
MihaiTabara: I believe they are mandatory everywhere because of https://github.com/mozilla-releng/scriptworker-scripts/blob/master/shipitscript/shipitscript/utils.py#L32. If so, I believe I should address this param to be mandatory for all of the functions, not just `trigger_release_phase`, wouldn't I?

shipitscript/shipitscript/ship_actions.py
MihaiTabara: I've enriched the comment to say we're attempting to minimize the possiblity of a race condition. True that it can still occur. Is there a better way to solve this?

shipitscript/shipitscript/ship_actions.py
MihaiTabara: Sounds good. Done.

shipitscript/shipitscript/ship_actions.py
MihaiTabara: Yeah, I initially had it like that but then I thought some logging won't hurt. But afterall, we can catch that after returning the function. Will do.

shipitscript/shipitscript/ship_actions.py
MihaiTabara: Good catch

shipitscript/shipitscript/script.py
MihaiTabara: I was thinking of the possiblity we're running these checks on a newly created branch (maybe we resurrect a new `aurora` or we run this a new project branch. But yeah, better to exit and notify the user this is at least happening. Will address it.

rail:
:shipit:

PR #94, opened 2019-11-28
Bug 1600104 - Expose `fennec-nightly-signing` scope to fenix

JohanLorenzo:
`AUTOGRAPH_FENNEC_NIGHTLY_*` is available per https://github.com/mozilla-releng/scriptworker-scripts/blob/d64a7f9120f63ae534ec143b1484fc2e53fe5452/signingscript/docker.d/passwords.yml#L147-L148

We use `autograph_apk` instead `autograph_apk_fennec_sha1` of because of https://github.com/mozilla-releng/build-puppet/pull/591#discussion_r320742427

### DIFF
--- a/signingscript/docker.d/passwords.yml
+++ b/signingscript/docker.d/passwords.yml
@@ -202,6 +202,12 @@ in:
              {"$eval": "AUTOGRAPH_FENIX_PASSWORD"},
              ["autograph_apk"]
             ]
+        project:mobile:fenix:releng:signing:cert:fennec-nightly-signing:
+          - ["https://autograph-external.prod.autograph.services.mozaws.net",
+             {"$eval": "AUTOGRAPH_FENNEC_NIGHTLY_USERNAME"},
+             {"$eval": "AUTOGRAPH_FENNEC_NIGHTLY_PASSWORD"},
+             ["autograph_apk"]
+            ]
         project:mobile:fenix:releng:signing:cert:fennec-production-signing:
           - ["https://autograph-external.prod.autograph.services.mozaws.net",
              {"$eval": "AUTOGRAPH_FENNEC_RELEASE_USERNAME"},
--- a/signingscript/docker.d/init_worker.sh
+++ b/signingscript/docker.d/init_worker.sh
@@ -144,6 +144,8 @@ case $ENV in
         test_var_set 'AUTOGRAPH_WIDEVINE_USERNAME'
         ;;
       mobile)
+        test_var_set 'AUTOGRAPH_FENNEC_NIGHTLY_PASSWORD'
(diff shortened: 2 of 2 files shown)


signingscript/docker.d/passwords.yml
```diff
@@ -202,6 +202,12 @@ in:
              {"$eval": "AUTOGRAPH_FENIX_PASSWORD"},
              ["autograph_apk"]
             ]
+        project:mobile:fenix:releng:signing:cert:fennec-nightly-signing:
```
MihaiTabara, line 205: It'd be good to fail-fast in case we forgot to set them, so let's add them in [here](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/signingscript/docker.d/init_worker.sh#L147) too.

MihaiTabara (approved):
This LGTM with a caveat that we should extend the fail-safe checks in the script so that it doesn't fail silently for lacking the proper credentials.

Leftovers once this PR is pushed:
1. Update SOPS to ensure the `firefoxci-mobile-3` has the new values. I've already taken care of that
2. We need to redeploy this change of scriptworker-scripts to the signing workers to make sure they pick up this image + thew new credentilals from the infrastructure. That means pushing either to `production` (updates all workers) or `production-signingscript` (updates only this worker) within this repository.

signingscript/docker.d/passwords.yml
JohanLorenzo: Thanks for the pointer, TIL 🙂👍

JohanLorenzo:
Thanks for the details, Mihai! On my end, I took care of the deployment https://github.com/mozilla-releng/scriptworker-scripts/commits/production.

signingscript/docker.d/passwords.yml
escapewindow, line 205: Should we remove the trailing `:`?

JohanLorenzo: That's a good question. I see the other entries have it too. Is it causing any trouble?

escapewindow: It may cause confusion when we add it to the restricted scopes [here](https://github.com/escapewindow/scriptworker/blob/master/src/scriptworker/constants.py#L389-L404).

JohanLorenzo: Okay. As far as I understand, the trailing comma indicates `project:mobile:fenix:releng:signing:cert:fennec-nightly-signing` is a key of a YAML object defined here https://github.com/mozilla-releng/scriptworker-scripts/blob/c94fc477ddddbb1984ea8f6c0d93398980c92157/signingscript/docker.d/passwords.yml#L14

I'm sorry, I'm not too sure to understand where the confusion may come from.

PR #96, opened 2019-11-29
Shipitscript - bail if buildo is not 1

MihaiTabara:


rail (approved):
The commit message needs some love ;)

PR #103, opened 2019-12-05
Override the default master doc in RTD

MihaiTabara:


MihaiTabara:
I broke Readthedocs engine because the default `conf.py` file expects a different master file.  Overriding that should help overcoming this error:

```
Sphinx error:
master file /home/docs/checkouts/readthedocs.org/user_builds/scriptworker-scripts/checkouts/latest/docs/contents.rst not found
```

JohanLorenzo (approved):

PR #113, opened 2019-12-09
Do not override python dependencies

rail:
Ideally we should install configloader isolated, I'll tackle this in another PR. This PR uses the approach we used before I added the problematic `pip install -r requirements/base.txt`


rail:
#114 is better

PR #114, opened 2019-12-09
Install configloader in venv

rail:


JohanLorenzo (approved):
I see `PyYAML==3.13` is not overriden anymore in https://firefoxci.taskcluster-artifacts.net/FNaBgWX7REWYY8bJHiyU5Q/0/public/logs/live_backing.log. Having a dedicated virtualenv makes sense to me. I wonder how `configloader_venv` gets used, though.

rail:
I'm going to deploy this to pushsnap only first, verify and then merge to production.

PR #119, opened 2019-12-11
Add merge_day action to treescript

srfraser:
Some rework of do_actions has been required to keep flake8 happy about the complexity. 

I'd like someone with solid mercurial knowledge to look over the process and how it's changed from [gecko_migration.py](https://searchfox.org/mozilla-central/source/testing/mozharness/scripts/merge_day/gecko_migration.py#449)


treescript/src/treescript/merge_config.py
```diff
@@ -0,0 +1,99 @@
+"""Configuration for repository merges."""
+
+NEW_ESR_BRANCH = "esr68"
+
+merge_configs = {
```
tomprince: I'm not sure if this config belongs in treescript, or in-tree (particularly as it is also used by `mach try release` to generate the changes required for a staging release.

treescript/src/treescript/versionmanip.py
```diff
@@ -126,7 +155,7 @@ def get_version(file_, parent_directory=None):
                 next_version.is_beta,
             )
         ):
-            next_version = VersionClass.parse("{}esr".format(bump_info["next_version"]))
+            next_version = VersionClass.parse("{}esr".format(next_version))
```
tomprince, line 145: I guess the intent is for this to saved_next_version? In any case, I would be inclined to spell this `next_version = attr.evolve(next_version, is_esr=True)` rather than parsing again.

tomprince (commented):
It would probably make it slightly easier to review if this were rebased on master; there are a lot of whitespace changes that show up in github.

This is not a complete review, and I don't have any overall thoughts, but just a couple (rough) thoughts I had while skimming this at my EoD.

treescript/src/treescript/merge_config.py
srfraser: I kept it in-tree because it's then in version control, rather than being in the hook that will call the task. I agree that it would be nicer from the scriptworker's point of view to have this configuration outside, but I'm reluctant to do that without an audit trail of what it's been told to do.

treescript/src/treescript/versionmanip.py
srfraser: Either `next_version` should work at that stage. I didn't know about `attr.evolve`, it does seem more efficient, possibly less clear depending on familiar people are with it. Will have a think.

srfraser:
> It would probably make it slightly easier to review if this were rebased on master; there are a lot of whitespace changes that show up in github.

The only thing different from master is the dirschema commit from the 12th. I'll update again.
 
> This is not a complete review, and I don't have any overall thoughts, but just a couple (rough) thoughts I had while skimming this at my EoD.

I understood Jordan had spoken with you about looking over the mercurial commands to make sure they're reasonable. Do they look ok?

treescript/src/treescript/mercurial.py
```diff
@@ -347,7 +353,8 @@ def _count_outgoing(output):
         PushError: on failure
 
     """
-    source_repo = get_source_repo(task)
+    if not source_repo:
+        source_repo = get_source_repo(task)
     source_repo_ssh = source_repo.replace("https://", "ssh://")
     ssh_username = config.get("hg_ssh_user")
     ssh_key = config.get("hg_ssh_keyfile")
```
tomprince: We should use the `ffxbld-merge`/`trybld-merge`/`tbirdbld-merge` keys for merge day things.

I think we probably want to have a new scope for which user to use and corresponding field in the payload.

tomprince:
I rebased this on tip, and ran black+isort on it. (Linting was not running in CI, see #125, so the errors did not appear here).

treescript/src/treescript/merge_config.py
tomprince: It would be in mozilla-central, not in the hook.

treescript/src/treescript/merges.py
```diff
@@ -0,0 +1,180 @@
...
+from treescript.exceptions import TaskVerificationError
+from treescript.mercurial import get_revision, run_hg_command
+from treescript.merge_config import merge_configs
+from treescript.task import get_merge_flavor
+from treescript.versionmanip import do_bump_version, get_version
+
+log = logging.getLogger(__name__)
+
+
+def replace(file_name, from_, to_):
```
tomprince, line 16: I wonder if would make sense to handle multiple changes at once?

treescript/src/treescript/merges.py
```diff
@@ -0,0 +1,180 @@
...
+    for line in contents.splitlines():
+        line = line.strip()
+        if line.startswith("#") or line == "":
+            new_contents += f"{line}\n"
+    new_contents += "Merge day clobber"
+    with open(clobber_file, "w") as f:
+        f.write(new_contents)
+
+
+async def apply_rebranding(config, repo_path, merge_config):
```
tomprince, line 44: This name seems misleading.

treescript/src/treescript/merges.py
```diff
@@ -0,0 +1,180 @@
...
+    """
+    flavor = get_merge_flavor(task)
+
+    if flavor not in merge_configs:
+        raise TaskVerificationError("Unknown configuration for merge day flavor {}".format(flavor))
+
+    from_branch = merge_configs[flavor].get("from_branch")
+    to_branch = merge_configs[flavor].get("to_branch")
+
+    log.info("hg pull %s", repo_path)
```
tomprince: If we are logging commands like this, it would probably make sense to have a helper that does it, rather than requiring the logging be kept in sync with the commands.

treescript/src/treescript/merges.py
```diff
@@ -0,0 +1,180 @@
...
+    await run_hg_command(config, "pull", repo_path=repo_path)
```
tomprince: We should be explicit were we are pulling from, and not depend on `hg config paths.default` being mozilla-unified

treescript/src/treescript/merges.py
```diff
@@ -0,0 +1,180 @@
...
+
+    log.info("hg up -C %s", from_branch)
```
tomprince: We probably need to be careful here to ensure that we are picking up the right bookmark. If there has somehow been local divergence, pulling will create bookmarks like `central@default` (where default is the path alias). I think we should enable the `firefoxtree` extension, which will ensure that the names refer to the appropriate branches.

treescript/src/treescript/merges.py
```diff
@@ -0,0 +1,180 @@
...
+    await run_hg_command(config, "up", "-C", from_branch, repo_path=repo_path)
+
+    base_from_rev = await get_revision(config, repo_path)
```
tomprince: It would be better to do `hg identify -r <from_branch> -T {node}` rather than relying on the working directory being checked out to this version.

treescript/src/treescript/script.py
```diff
@@ -24,11 +51,14 @@
...
-        attempts (int, optional): the number of attempts to perform,
-            retrying on error.
-
     """
     await checkout_repo(config, task, repo_path)
+
+    # Split the action selection up due to complexity in do_actions
+    # caused by different push behaviour, and action return values.
+    if MERGE_ACTION_WORD in actions:
+        await perform_merge_actions(config, task, actions, repo_path)
```
tomprince, line 58: I do not think that this case should fall through to the rest of this function. I lean *slightly* towards thinking the rest this function should be moved into its own function (and then, I'd consider inlining it into it's caller, but that looking at the callsite, I'm not sure that is possible).

treescript/src/treescript/versionmanip.py
```diff
@@ -68,6 +68,27 @@ def get_version(file_, parent_directory=None):
...
+    Returns:
+        int: the number of commits created.
+
+    """
+    bump_info = get_version_bump_info(task)
+
+    return await do_bump_version(config, repo_path, bump_info["files"], bump_info["next_version"], get_dontbuild(task), commit=True)
+
+
+async def do_bump_version(config, repo_path, files, next_version, dontbuild, commit=True):
```
tomprince: Instead of passing `commit` (and `dontbuild`) as a parameter, `bump_version` above should handle making the commit.

treescript/src/treescript/merges.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+async def apply_rebranding(config, repo_path, merge_config):
+    """Apply changes to repo required for merge/rebranding."""
+    log.info("Rebranding %s to %s", merge_config.get("from_branch"), merge_config.get("to_branch"))
+    if merge_config.get("version_files"):
+        current_version = get_version("browser/config/version.txt", repo_path)
+        next_version = f"{current_version.major_number}.{current_version.minor_number}"
+
+        await do_bump_version(config, repo_path, merge_config["version_files"], next_version, dontbuild=True, commit=False)
```
tomprince: We should do some work to be able to call this once, rather than different times with different version numbers. I think it would also make sense to specify the suffix via `attr.evolve`[1] rather than a string.

[1] Maybe something like specifying `{'beta': 1}` that can be passed to that.

treescript/src/treescript/merges.py
```diff
@@ -0,0 +1,180 @@
...
+    if merge_config.get("version_files_suffix"):
+        current_version = get_version("browser/config/version.txt", repo_path)
+        next_version = f"{current_version.major_number}.{current_version.minor_number}{merge_config.get('version_suffix')}"
+
+        await do_bump_version(config, repo_path, merge_config["version_files_suffix"], next_version, dontbuild=True, commit=False)
+
+    for f in merge_config.get("copy_files", list()):
```
tomprince, line 59: The only use of `copy_files` is for removing `bX` from the version number. I wonder if we should handle that as part of version number bumping instead, and drop support for copying files.

treescript/src/treescript/merges.py
```diff
@@ -0,0 +1,180 @@
...
+        shutil.copyfile(os.path.join(repo_path, f["src"]), os.path.join(repo_path, f["dst"]))
+
+    for f, from_, to in merge_config.get("replacements", list()):
+        replace(os.path.join(repo_path, f), from_, to)
+
+    if merge_config.get("remove_locales"):
```
tomprince: I don't see this set anywhere, so we should delete this.

treescript/src/treescript/merge_config.py
tomprince: Also, are we going to handle the `bump_central` case?

treescript/src/treescript/merges.py
```diff
@@ -0,0 +1,180 @@
...
+    log.info("hg up -C %s", to_branch)
+    await run_hg_command(config, "up", "-C", to_branch, repo_path=repo_path)
+
+    log.info("Adding end tag")
+    end_tag = merge_configs[flavor].get("end_tag")  # tag the end of the to repo
+    if end_tag:
+        to_fx_major_version = get_version("browser/config/version.txt", repo_path).major_number
+        base_to_rev = await get_revision(config, repo_path)
+        end_tag = end_tag.format(major_version=to_fx_major_version)
+        await run_hg_command(config, "tag", "-m", end_tag, "-r", base_to_rev, "-u", config["hg_ssh_user"], "-f", end_tag, repo_path=repo_path)
```
tomprince: Note that tags are tracked in a file in the reposiory, and so `hg tag` actually creates a commit that modifies that file. This has a couple of implications:
- the `-m` argument is actually a commit message, so needs to be a valid commit message. See [here](https://searchfox.org/mozilla-central/rev/723e61e5f6fe36e69592b7742bc10338392e53e4/testing/mozharness/mozharness/mozilla/repo_manipulation.py#122-127) for where the gecko_migration script handles that.
- To push the tag, we need to push this new commit, not the rev that was saved above.

treescript/src/treescript/data/treescript_task_schema.json
```diff
@@ -44,6 +44,9 @@
                         "next_version": {
                             "type": "string"
                         },
+                        "commit": {
```
tomprince: I don't think this should be here.

treescript/src/treescript/script.py
```diff
@@ -8,11 +8,38 @@
 from treescript.exceptions import CheckoutError, PushError, TreeScriptError
 from treescript.l10n import l10n_bump
 from treescript.mercurial import checkout_repo, do_tagging, log_mercurial_version, log_outgoing, push, strip_outgoing, validate_robustcheckout_works
+from treescript.merges import do_merge
 from treescript.task import is_dry_run, task_action_types
 from treescript.versionmanip import bump_version
 
 log = logging.getLogger(__name__)
 
+MERGE_ACTION_WORD = "merge_day"
```
tomprince: I do not think making this a constant is useful. We could turn actions into an enum, and use that consistently, but I think that is out of scope for this PR.

treescript/src/treescript/merge_config.py
```diff
@@ -0,0 +1,76 @@
...
+            (
+                "build/mozconfig.common",
+                "# Disable enforcing that add-ons are signed by the trusted root",
+                "# Enable enforcing that add-ons are signed by the trusted root",
+            ),
+        ],
+        "from_branch": "central",
+        "to_branch": "beta",
+        "push_repositories": {"from": "https://hg.mozilla.org/mozilla-central", "to": "https://hg.mozilla.org/releases/mozilla-beta"},
+        "require_debugsetparents": True,
```
tomprince: We should name this something descriptive of what it is doing, rather than how it is implemented. Maybe `merge_old_head`?

treescript/src/treescript/data/treescript_task_schema.json
```diff
@@ -108,7 +121,7 @@
                     "uniqueItems": true,
                     "items": {
                         "type": "string",
-                        "enum": ["tag", "version_bump", "l10n_bump", "push"]
+                        "enum": ["tag", "version_bump", "l10n_bump", "merge_day", "push"]
```
tomprince: We should adjust this to enforce that `merge_day` is not specified with anything other than push.

I think specifying the array as `{'oneOf': [{..., "enum: ["tag", "version_bump", "push"]}, {..., "enum": ["l10n_bump", "push"], {..., "enum": ["merge_day", "push"]}]}` would handle that.

treescript/src/treescript/data/treescript_task_schema.json
```diff
@@ -87,6 +90,16 @@
                         "required": ["path", "name", "platform_configs", "version_path"]
                     }
                 },
+                "merge_info": {
```
tomprince, line 233: https://json-schema.org/understanding-json-schema/reference/conditionals.html would allow us to require this be specified iff `merge_day` is a requested action.

treescript/src/treescript/merges.py
tomprince: I also don't think we should pass `-u` anywhere to commits. We currently set a username [here](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/treescript/src/treescript/data/hgrc#L2). There is an argument to be made that we should set the author to ffxbld etc, but if we do that, I think it should be done by making the commit user linked above configurable, rather than using `hg_ssh_user` for that.

Everything but dropping the `-u` here is out of scope of this PR.

treescript/src/treescript/merges.py
```diff
@@ -0,0 +1,180 @@
...
+    for f in merge_config.get("copy_files", list()):
+        shutil.copyfile(os.path.join(repo_path, f["src"]), os.path.join(repo_path, f["dst"]))
+
+    for f, from_, to in merge_config.get("replacements", list()):
+        replace(os.path.join(repo_path, f), from_, to)
+
+    if merge_config.get("remove_locales"):
+        log.info("Removing locales")
+        remove_locales(os.path.join(repo_path, "browser/locales/shipped-locales"), merge_config["remove_locales"])
+    touch_clobber_file(repo_path)
```
tomprince, line 65: We should consider doing the initial l10n bump as part of the merge, now that the code for both lives here.

treescript/src/treescript/mercurial.py
```diff
@@ -335,7 +341,7 @@ def _count_outgoing(output):
 
 
 # push {{{1
-async def push(config, task, repo_path):
+async def push(config, task, repo_path, source_repo=None, revision=None):
```
tomprince: The argument to this should be `target_repo` (see #89) and required; given that there is only a single other caller it should be a required argument and that call-site can handle the default value you set below.

treescript/src/treescript/script.py
```diff
@@ -37,23 +67,44 @@
...
+
+    Returns:
+        bool
+    """
     if is_dry_run(task):
         log.info("Not pushing changes, dry_run was forced")
     elif "push" in actions:
-        if num_changes:
-            await push(config, task, repo_path)
+        if num_changes or MERGE_ACTION_WORD in actions:
```
tomprince: I think the handling of `num_changes` should be handled outside this function. I've done this refactoring in #129.

tomprince (changes requested):
This looks good, thanks! I've left a bunch of comments, some minor and some slight longer. The comments were written over an extended period of time, so might not all exactly hang together. I'm happy to chat about any thing here, if it is unclear.

treescript/src/treescript/merges.py
srfraser: Why? I was trying to think of a coherent name for the set of changes that get performed, and during discussions we settled on this, but I'm happy to change it if there's a better option.

treescript/src/treescript/data/treescript_task_schema.json
srfraser: I believe that was the plan for this year, as part of wider changes to introduce valid sets of operations.

treescript/src/treescript/mercurial.py
srfraser: I can change it to target - it was called that because it gets the default value from `metadata.source`, and I felt source_repo was clearer, but I can see arguments both ways.

treescript/src/treescript/script.py
srfraser: I'm not clear on what you mean, here. If it's "Use should_push instead" then great. If not then the linked PR looks the same as the original line in this part of the diff

treescript/src/treescript/merges.py
srfraser: > This looks good, thanks! I've left a bunch of comments, some minor and some slight longer. The comments were written over an extended period of time, so might not all exactly hang together. I'm happy to chat about any thing here, if it is unclear.

Do you have any comments about the mercurial instructions?

> It would be in mozilla-central, not in the hook.

I couldn't reply to this comment in-line and can't see what it's a reply to, but I'm guessing it's about the version control of the configuration. I'm guessing this is because we'd be using an action task to then create the actual merge task (Grumbles about overcomplication saved for later). I'm ok as long as the configuration is under vcs somewhere, although I worry that it'll be harder to find.

treescript/src/treescript/mercurial.py
srfraser: The merge day instance only seemed to know about the first of those. Hmm. The existing treescript key is set up with puppet, right? Or since you mention scopes would you prefer those to be taskcluster secrets?

treescript/src/treescript/merge_config.py
srfraser: There's already a version bump in treescript, would that be sufficient?

treescript/src/treescript/data/treescript_task_schema.json
srfraser: I like this idea, although I'd want to add it to the other payload options as well; perhaps in the same PR as the 'valid action sets' so that we don't end up contradicting it.

treescript/src/treescript/merges.py
srfraser: Probably not in the same task, though. I was imagining that the hook for the merge would result in all the required tasks being produced

treescript/src/treescript/merge_config.py
tomprince: Possibly, with a bit of work. There is at least the xpcom and weave version number bump that isn't handled, see [here](https://searchfox.org/mozilla-central/source/testing/mozharness/scripts/merge_day/gecko_migration.py#305). I've haven't investigated beyond that.

treescript/src/treescript/mercurial.py
tomprince: I guess I misremembered the key names (and imagined the `tbird` one). The second one I was thinking of is `stage-ffxbld-merge` (see [here](https://github.com/mozilla-releng/build-puppet/blob/master/modules/mergeday/manifests/install.pp#L27)). They definitely should not be taskcluster secrets.

My suggestion about scopes was to have something like the `:bucket:*` scopes [here](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/beetmoverscript/SCOPES.md) to control access to the different keys.

treescript/src/treescript/script.py
tomprince: The relevant difference to what you implemented here was that `should_push` there doesn't look at `num_changes` (that is now handled in the caller), and so `should_push` does not need to special case the  `merge` action.

So, you should be able to use `should_push` from that PR as-is.

tomprince (commented):
> > This looks good, thanks! I've left a bunch of comments, some minor and some slight longer. The comments were written over an extended period of time, so might not all exactly hang together. I'm happy to chat about any thing here, if it is unclear.
> 
> Do you have any comments about the mercurial instructions?

I commented on them in an number of places. It looks like a bunch of them are hidden behind `13 hidden conversations`: 
image


> 
> > It would be in mozilla-central, not in the hook.
> 
> I couldn't reply to this comment in-line and can't see what it's a reply to, but I'm guessing it's about the version control of the configuration. I'm guessing this is because we'd be using an action task to then create the actual merge task (Grumbles about overcomplication saved for later). I'm ok as long as the configuration is under vcs somewhere, although I worry that it'll be harder to find.

It may be hard to find for people in releng (though I'm not sure). I suspect that having in mozilla-central would make it easier to find for firefox devs.

treescript/src/treescript/mercurial.py
srfraser: Hm, they probably shouldn't be in the payload, if they're a scope, as the duplication won't add much. Is it likely there'll be someone who can use `stage-ffxbld-merge` and not the production one? 
We should really get these keys added to the usual place.

treescript/src/treescript/merges.py
srfraser: The only repo we know about is mozilla-unified, so does it add anything?

treescript/src/treescript/merges.py
srfraser: I'm not clear on what you mean here: It's a fresh clone, so local divergence doesn't seem to be an issue. `firefoxtree` has a lot of dependencies that will vendoring and make deploying treescript a lot more awkward.

treescript/src/treescript/versionmanip.py
srfraser: We'd then need to duplicate those commits inside `bump_version` and twice in `apply_rebranding` - the contents of the old `bump_version` were drawn out for re-use so that it could be called from the task payload and also in a merge.

treescript/src/treescript/merges.py
srfraser: It seemed simpler this way than having a special case, but if you've a suggestion about how to restructure it I'll happily try it out. I admit I think `attr.evolve` is less clear here than a string, and it doesn't address restructuring to combine two distinct set of replacements

treescript/src/treescript/merges.py
srfraser: Fair

treescript/src/treescript/merges.py
srfraser: >     * the `-m` argument is actually a commit message, so needs to be a valid commit message. See [here](https://searchfox.org/mozilla-central/rev/723e61e5f6fe36e69592b7742bc10338392e53e4/testing/mozharness/mozharness/mozilla/repo_manipulation.py#122-127) for where the gecko_migration script handles that.

Thanks, I was doing that in the earlier tagging in that function, I'd missed it there.

treescript/src/treescript/script.py
srfraser: Hmm, yes I can see the benefit. There's only one value at the moment but an enum would make the intent clearer.

treescript/src/treescript/data/treescript_task_schema.json
srfraser: hm, you're probably right. I was imagining a situation where the version bumping during a merge would be handled by the payload, and so we'd want to hold off on a commit until it was ready, but that's now hidden away inside a function.

treescript/src/treescript/merge_config.py
srfraser: In any case, it was already out of scope for this PR. We'll do it in the next one.

treescript/src/treescript/versionmanip.py
tomprince: The call in `apply_rebranding` always passes `commit=False` though. I'm suggesting moving the last `if changed and commit` block into the one caller that currently passes `commit=True`. I don't think there needs to be any duplication.

treescript/src/treescript/merges.py
tomprince: We don't do a fresh clone. The worker keeps a single clone, and then reuses it between multiple tasks.

Regarding vendoring, I think we could add 
```
hg clone -r db888e6d3d88172f4d3518d0aa46c3a587a774c0 https://hg.mozilla.org/hgcustom/version-control-tools /app/version-control-tools
```
to the docker file and
```
[extensions]
/app/version-control-tools/version-control-tools/hgext/firefoxtree
```
to the hgrc. That way we can depend on the upstream vendoring, rather than needing to duplicate it.

treescript/src/treescript/mercurial.py
tomprince: > Hm, they probably shouldn't be in the payload, if they're a scope, as the duplication won't add much.

My general thought is that scopes should not control the behavior of a task, but only limit what they can do. I've been (slowly) pushing to move that kind of configuration in the payloads, and so when defining new things, we should put them in the payload to begin with.

> Is it likely there'll be someone who can use `stage-ffxbld-merge` and not the production one?

Yes. My inclination would to make the key usable by anyone pushing to try:
- it is a level-1 key, so it doesn't grant any additional permissions.
- it make it easier for non-relengers to test changes (particularly to the in-tree config.

treescript/src/treescript/merge_config.py
tomprince: > In any case, it was already out of scope for this PR. We'll do it in the next one.

Definitely. It is worth considering if that future change will have impact on how the current one is structured, and I was just asking about it ot ensure that that had been done, and it sounds like it has.

treescript/src/treescript/merges.py
tomprince: I generally prefer to be explicit when interacting with things in automation. 

`hg robustcheckout` doesn't set `hg config paths.default`, except on the initial clone. If some other action were to were to clone mozilla-central rather than mozilla-unified and that action happened to run on this worker first, this would do the wrong thing.

Additionally, the following code depends on all the mozilla-* bookmarks/firefoxtree heads being pulled, so I think being explicit here helps people figuring out what is going on. (Note that for thunderbird usages of treescript, the repo will be a clone of comm-central (since there is a comm-unified). That doesn't affect this code, but does mean that people looking at this code might not assume the repo has a `paths.default = .../mozilla-unified`).

treescript/docker.d/extensions.rc
```diff
@@ -0,0 +1,2 @@
+[extensions]
+firefoxtree=/app/version-control-tools
```
tomprince: nit: newline

treescript/src/treescript/merges.py
```diff
@@ -0,0 +1,138 @@
...
+    base_from_rev = await get_revision(config, repo_path, branch=from_branch)
+
+    base_tag = merge_config["base_tag"].format(major_version=get_version("browser/config/version.txt", repo_path).major_number)
+
+    tag_message = f"No bug - tagging {os.path.basename(repo_path)} with {base_tag} a=release DONTBUILD CLOSED TREE"
+    await run_hg_command(config, "tag", "-m", '"{}"'.format(tag_message), "-r", base_from_rev, "-f", base_tag, repo_path=repo_path)
+
+    # TODO This shouldn't be run on esr, according to old configs.
+    # perhaps: hg push -r bookmark("release") esrNN
+    # Perform the kludge-merge.
```
tomprince, line 108: I think this comment could be improved somewhat. The idea is we want a single head in each repository, so we merge (in the DAG) the old head, with the new head from central or beta, but in such a way that we keep the contents from central or beta.

The reason we don't do this on esr, is that we are not merging in changes from another repo, so there isn't an old head to merge in with a new one.

treescript/src/treescript/merges.py
```diff
@@ -0,0 +1,138 @@
...
+    with open(path, "w") as fh:
+        fh.write(diff_output)
+
+    await run_hg_command(config, "commit", "-m", "Update configs. IGNORE BROKEN CHANGESETS CLOSED TREE NO BUG a=release ba=release", repo_path=repo_path)
+    push_revision_to = await get_revision(config, repo_path, branch=to_branch)
+
+    # Do we need to perform multiple pushes for the push stage? If so, return
+    # what to do.
+    if merge_config.get("push_repositories"):
+        return [(merge_config["push_repositories"]["from"], base_from_rev), (merge_config["push_repositories"]["to"], push_revision_to)]
```
tomprince: Note that `base_from_rev` here needs to be changed to point to the commit that created the tag (rather than the commit that was tagged). As it is, nothing will get pushed to the "from" repository.

tomprince (approved):
One last change (the revision to push for the base tag), otherwise this looks good. I will do another pass for the tests, but that isn't blocking.

treescript/src/treescript/merges.py
srfraser: I'm not clear on why esr is different, here, but I'm happy to update the comment if you could help form one. The mercurial commands are just deep magic to me.

treescript/src/treescript/mercurial.py
```diff
@@ -222,6 +222,11 @@ def build_hg_environment():
     return tags
 
 
+async def get_revision(config, repo_path, branch):
```
tomprince, line 225: `s/branch/revset/` here, I think. This can be any [revset](https://www.mercurial-scm.org/repo/hg/help/revsets).

treescript/src/treescript/merges.py
```diff
@@ -0,0 +1,140 @@
...
+    await run_hg_command(config, "up", "-C", from_branch, repo_path=repo_path)
+
+    base_from_rev = await get_revision(config, repo_path, branch=from_branch)
+
+    base_tag = merge_config["base_tag"].format(major_version=get_version("browser/config/version.txt", repo_path).major_number)
+
+    tag_message = f"No bug - tagging {os.path.basename(repo_path)} with {base_tag} a=release DONTBUILD CLOSED TREE"
+    await run_hg_command(config, "tag", "-m", '"{}"'.format(tag_message), "-r", base_from_rev, "-f", base_tag, repo_path=repo_path)
+
+    tagged_from_rev = await get_revision(config, repo_path, branch=from_branch)
```
tomprince: This will need to be `branch="."` (or based on my other comment `revset="."`); branch names we are using here are static. `.` refers to the current commit.

treescript/src/treescript/merges.py
```diff
@@ -0,0 +1,140 @@
...
+    await apply_rebranding(config, repo_path, merge_config)
+
+    diff_output = await run_hg_command(config, "diff", repo_path=repo_path, return_output=True)
+    path = os.path.join(config["artifact_dir"], "public", "logs", "{}.diff".format(to_branch))
+    makedirs(os.path.dirname(path))
+    with open(path, "w") as fh:
+        fh.write(diff_output)
+
+    await run_hg_command(config, "commit", "-m", "Update configs. IGNORE BROKEN CHANGESETS CLOSED TREE NO BUG a=release ba=release", repo_path=repo_path)
+    push_revision_to = await get_revision(config, repo_path, branch=to_branch)
```
tomprince: Same here `revset="."`.

treescript/src/treescript/mercurial.py
srfraser: If we don't know the revision, how would we pass it in? I think the use-case here is when we're given a bookmark

treescript/src/treescript/merges.py
```diff
@@ -0,0 +1,147 @@
...
+
+        await do_bump_version(config, repo_path, merge_config["version_files_suffix"], next_version)
+
+    for f in merge_config.get("copy_files", list()):
+        shutil.copyfile(os.path.join(repo_path, f[0]), os.path.join(repo_path, f[1]))
+
+    for f, from_, to in merge_config.get("replacements", list()):
+        replace(os.path.join(repo_path, f), from_, to)
+
+    touch_clobber_file(repo_path)
```
jfx2006, line 65: Is it possible to add a parameter in the task description to skip running "touch_clobber_file()"? Thunderbird repos won't have that.

treescript/src/treescript/merges.py
jfx2006: The way we do Thunderbird merges right now, I would need something like copy_files. (It's also possible that what we are doing could be done another way.)

treescript/src/treescript/merges.py
srfraser: This PR isn't intended for Thunderbird work - it doesn't cover all of the manual Firefox tasks yet. Could we add those when we add the Thunderbird tasks?

jfx2006: Of course.

srfraser: Thank you for being understanding!

PR #121, opened 2019-12-12
bug 1562412 - Notarization poller worker

escapewindow:
This is the poller worker, as described in [the user story for bug 1562412](https://bugzilla.mozilla.org/show_bug.cgi?id=1562412#user-story). We need to be able to claim and track multiple concurrent tasks, poll Apple for each of their statuses, and resolve tasks when Apple's ready.

It looks like we could potentially abstract away some of this logic into a base python worker module, and share it with scriptworker. I've explicitly decided not to do that this time around.


notarization_poller/src/notarization_poller/config.py
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env python
+"""Config for scriptworker.
```
tomprince: Not scriptworker (this occurs a few times throughout)

notarization_poller/src/notarization_poller/constants.py
```diff
@@ -0,0 +1,63 @@
...
+    DEFAULT_CONFIG (frozendict): the default config for scriptworker.  Running configs
+        are validated against this.
+    STATUSES (dict): maps taskcluster status (string) to exit code (int).
+
+"""
+import os
+
+from frozendict import frozendict
+
+STATUSES = frozendict(
```
tomprince: This seems like something that should live in taskcluster-python maybe?

escapewindow:
Moving to review-ready before I've fully tested, since I'm a ways away from in-tree patches. I'm hoping to avoid the situation where I'm all ready to land, but am missing review during the holidays. I can try to keep track of the interdiff between the reviewed revision and any new changes.

@mitchhentges do you have time / headspace for giving this a review pass? If not, possibly @JohanLorenzo ?

mitchhentges:
Oof, this is a big patch. I'll look at this later this week if I have time :)

escapewindow:
Yeah. @tomprince said he could take a look too.

notarization_poller/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+WORKDIR /app
+
+COPY . /app
+
+RUN python -m venv /app \
+ && cd notarization_poller \
+ && /app/bin/pip install -r requirements/base.txt \
+ && /app/bin/pip install . \
+ && cd /app \
+ && ./bin/pip install https://github.com/rail/configloader/archive/8719e33f6ba65e79e188355efabbcf6c9e1f1780.tar.gz
```
JohanLorenzo: Nit: Configloader is now installed differently. See https://github.com/mozilla-releng/scriptworker-scripts/pull/114

notarization_poller/README.md
```diff
@@ -0,0 +1,15 @@
+# `notarization_poller`
+
+This is a worker in its own right, and does not run under scriptworker.
```
JohanLorenzo, line 3: 👍 to this precision!

notarization_poller/src/notarization_poller/config.py
```diff
@@ -0,0 +1,110 @@
...
+from frozendict import frozendict
+from yaml import safe_load
+
+from notarization_poller.constants import DEFAULT_CONFIG
+from notarization_poller.exceptions import ConfigError
+
+log = logging.getLogger(__name__)
+
+
+def update_logging_config(config, log_name=None, file_name="worker.log"):
```
JohanLorenzo: Potential followup: Do you think we should replace [_init_logging()](https://github.com/mozilla-releng/scriptworker-scripts/blob/fcf3a030bdc87e5d1d964ad6c8990c6f2ded6fa0/scriptworker_client/src/scriptworker_client/client.py#L145) with this more complete function?

notarization_poller/src/notarization_poller/config.py
```diff
@@ -0,0 +1,110 @@
...
+    else:
+        # Avoid using WatchedFileHandler during scriptworker unittests
+        handler = logging.FileHandler(path)
+    handler.setFormatter(formatter)
+    top_level_logger.addHandler(handler)
+    top_level_logger.addHandler(logging.NullHandler())
+
+
+# read_config {{{1
+def read_config(config_path="poller.yaml"):
```
JohanLorenzo: Potential followup: do you think we should make `[_init_config()](https://github.com/mozilla-releng/scriptworker-scripts/blob/fcf3a030bdc87e5d1d964ad6c8990c6f2ded6fa0/scriptworker_client/src/scriptworker_client/client.py#L128) use this function?

notarization_poller/src/notarization_poller/constants.py
JohanLorenzo: Nit: How about re-using https://github.com/mozilla-releng/scriptworker-scripts/blob/fcf3a030bdc87e5d1d964ad6c8990c6f2ded6fa0/scriptworker_client/src/scriptworker_client/constants.py#L10-L19 ?

notarization_poller/src/notarization_poller/exceptions.py
```diff
@@ -0,0 +1,38 @@
...
+
+    Attributes:
+        exit_code (int): this is set to 5 (internal-error).
+
+    """
+
+    exit_code = STATUSES["internal-error"]
+
+
+class RetryError(WorkerError):
```
JohanLorenzo, line 21: Nit: we can reuse https://github.com/mozilla-releng/scriptworker-scripts/blob/fcf3a030bdc87e5d1d964ad6c8990c6f2ded6fa0/scriptworker_client/src/scriptworker_client/exceptions.py#L64

I wonder if we should move the other exceptions there too. What do you think?

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,316 @@
...
+    def task_credentials(self):
+        """Return the temporary credentials returned from [re]claimWork."""
+        return self._reclaim_task.get("credentials", self.claim_task["credentials"])
+
+    async def reclaim_task(self):
+        """Try to reclaim a task from the queue.
+
+        This is a keepalive.  Without it the task will expire and be re-queued.
+
+        A 409 status means the task has been resolved.
```
JohanLorenzo: 👍 to this contextual comment!

JohanLorenzo (approved):
Looks great to me! I like how multiple tasks are handled in a single worker. I haven't found any major issue. 

I haven't looked too closely at the project configuration (tox, flake8 and all), because I assumed it's a copy-pasta of the other projects. 

👍 👍 👍

notarization_poller/src/notarization_poller/config.py
escapewindow: Hm. I think we avoided formatting the logging from the script too much, to avoid double timestamps and the like (`worker_timestamp worker_level - worker_msg` : the `worker_msg` is the task output, so with full script log formatting we would get `worker_timestamp worker_level - script_timestamp script_level - script_msg`.  Also, we don't have to worry about rotating file logs in a script, because it's not long-running.

notarization_poller/src/notarization_poller/config.py
escapewindow: Hm, possibly. The default isn't good, but we could remove the default or rename it. We could also pass in the default config instead of hardcoding DEFAULT_CONFIG. I might try using `_init_config()` here.

[EDIT] yup, I did it.

notarization_poller/src/notarization_poller/exceptions.py
escapewindow: Hm. This is a bit different in that we're setting a `resource-unavailable` status. Other than that, possibly.

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+    uuids = None
+
+    def __init__(self, config, claim_task, event_loop=None):
+        """Initialize Task."""
+        self.config = config
+        self.task_id = claim_task["status"]["taskId"]
+        self.run_id = claim_task["runId"]
+        self.claim_task = claim_task
+        self.event_loop = event_loop or asyncio.get_event_loop()
+        self.task_dir = os.path.join(config["work_dir"], self.task_id)
```
tomprince: It would probably be useful to include the run-id too.

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+
+        """
+        while True:
+            log.debug("waiting %s seconds before reclaiming..." % self.config["reclaim_interval"])
+            await asyncio.sleep(self.config["reclaim_interval"])
+            log.debug("Reclaiming task %s %s", self.task_id, self.run_id)
+            try:
+                async with aiohttp.ClientSession() as session:
+                    temp_queue = Queue(options={"credentials": self.task_credentials, "rootUrl": self.config["taskcluster_root_url"]}, session=session)
+                    self.reclaim_task = await temp_queue.reclaimTask(self.task_id, self.run_id)
```
tomprince: Based on `__init__`, this should be `_reclaim_task`

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+        A 409 status means the task has been resolved.
+
+        Raises:
+            TaskclusterRestFailure: on non-409 status_code from
+                `taskcluster.aio.Queue.reclaimTask`
+
+        """
+        while True:
+            log.debug("waiting %s seconds before reclaiming..." % self.config["reclaim_interval"])
+            await asyncio.sleep(self.config["reclaim_interval"])
```
tomprince, line 109: Rather than being hard-coded in the config, this should probably be based on `takenUntil` from `claimWork`/`reclaimTask`

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+        """Run the task, creating a task-specific log file."""
+        self.status = 0
+        username = self.config["notarization_username"]
+        password = self.config["notarization_password"]
+
+        await self.download_uuids()
+        done = False
+        while not done:
+            for uuid in [u for u in self.uuids if not self.uuids[u]]:
+                base_cmd = ["xcrun", "altool", "--notarization-info", uuid, "-u", username, "--password"]
```
tomprince: (possible followup) Looking at [this](https://developer.apple.com/documentation/xcode/notarizing_macos_software_before_distribution/customizing_the_notarization_workflow#3087732), it appears that it is possible to store the password in the OSX keychain. I wonder if it would be possible to use that to avoid having the password in-process here?

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+                            raise RetryError("Bad status {}".format(resp.status))
+
+    async def complete_task(self):
+        """Submit task status to Taskcluster."""
+        reversed_statuses = get_reversed_statuses()
+        args = [self.task_id, self.run_id]
+        try:
+            async with aiohttp.ClientSession() as session:
+                temp_queue = Queue(options={"credentials": self.task_credentials, "rootUrl": self.config["taskcluster_root_url"]}, session=session)
+                if self.status == 0:
```
tomprince: I'd be inclined to use something symbolic, rather than `0` and `1` here.

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+    log.debug("Calling claimWork for {}/{}...".format(config["worker_group"], config["worker_id"]))
+    payload = {"workerGroup": config["worker_group"], "workerId": config["worker_id"], "tasks": num_tasks}
+    try:
+        return await worker_queue.claimWork(config["provisioner_id"], config["worker_type"], payload)
+    except (taskcluster.exceptions.TaskclusterFailure, aiohttp.ClientError) as exc:
+        log.warning("{} {}".format(exc.__class__, exc))
+
+
+# create_temp_creds {{{1
+def create_temp_creds(client_id, access_token, start=None, expires=None, scopes=None, name=None):
```
tomprince: This appears unused.

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+                self.task_log("Polling response (status %d)", status, worker_log=False)
+                for line in contents.splitlines():
+                    self.task_log(" %s", line, worker_log=False)
+                if status == 0:
+                    m = NOTARIZATION_POLL_REGEX.search(contents)
+                    if m is not None:
+                        if m["status"] == "invalid":
+                            self.status = STATUSES["failure"]
+                            self.task_log("Apple believes UUID %s is invalid!", uuid, level=logging.CRITICAL)
+                            done = True
```
tomprince: This variable seems superfluous with the `break`.

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+            self.uuids[uuid] = False
+        self.task_log("UUIDs: %s", self.uuids)
+
+    async def run_task(self):
+        """Run the task, creating a task-specific log file."""
+        self.status = 0
+        username = self.config["notarization_username"]
+        password = self.config["notarization_password"]
+
+        await self.download_uuids()
```
tomprince, line 230:
```suggestion
        uuids = await self.download_uuids()
        pending_uuids = set(uuids)
```

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+        done = False
+        while not done:
+            for uuid in [u for u in self.uuids if not self.uuids[u]]:
```
tomprince:
```suggestion
            for uuid in sorted(pending_uuids):
```

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+                            self.task_log("Apple believes UUID %s is invalid!", uuid, level=logging.CRITICAL)
+                            done = True
+                            break
+                        # There are only two possible matches with the regex
+                        # Adding `pragma: no branch` to be explicit in our
+                        # checks, but still avoid testing an unreachable code
+                        # branch
+                        if m["status"] == "success":  # pragma: no branch
+                            self.task_log("UUID %s is successful", uuid)
+                            self.uuids[uuid] = True
```
tomprince:
```suggestion
                            pending_uuids.remove(uuid)
```

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+            if all(self.uuids.values()):
+                self.task_log("All UUIDs are successfully notarized: %s", self.uuids)
```
tomprince, line 266:
```suggestion
                self.task_log("All UUIDs are successfully notarized: %s", uuids)
```

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+                break
+            await asyncio.sleep(self.config["poll_sleep_time"])
+        await self.stop(kill_run_task=False)
+
+
+# claim_work {{{1
+async def claim_work(config, worker_queue, num_tasks=1):
```
tomprince: It feels like this should live in `worker.py`, or `RunTasks` here.

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+
+    async def stop(self, status=None, kill_run_task=True):
+        """Stop the task."""
+        if status is not None:
+            self.status = status
+        log.info("Stopping task %s %s with status %s", self.task_id, self.run_id, self.status)
+        self.reclaim_fut and self.reclaim_fut.cancel()
+        if kill_run_task and self.task_fut:
+            self.task_fut.cancel()
+        if self.process:
```
tomprince: This doesn't appear to ever be set.

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+        password = self.config["notarization_password"]
+
+        await self.download_uuids()
+        done = False
+        while not done:
+            for uuid in [u for u in self.uuids if not self.uuids[u]]:
+                base_cmd = ["xcrun", "altool", "--notarization-info", uuid, "-u", username, "--password"]
+                log_cmd = base_cmd + ["********"]
+                rm(self.poll_log_path)
+                status = await run_command([base_cmd + [password]], log_cmd=log_cmd, log_path=self.poll_log_path)
```
tomprince: This is probably something that needs to be handled in `run_command`, but something should handle `CanceledError` (per the [`Task.cancel`](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancel) documentation), and cleanup the process when that is received.

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+        except TaskError:
+            self.status = STATUSES["malformed-payload"]
+            self.task_log(traceback.format_exc(), level=logging.CRITICAL)
+            await self.stop()
+        self.uuids = {}
+        for uuid in uuids:
+            self.uuids[uuid] = False
+        self.task_log("UUIDs: %s", self.uuids)
+
+    async def run_task(self):
```
tomprince, line 224: I think the body of `run_task` and `download_uuids` should be factored out of `Task` and `Task` left as an object  that knows how to run a generic task, given a function that takes a task, and returns some sort of status + logs.

notarization_poller/src/notarization_poller/task.py
tomprince: It might be useful to expand this to indicate that this is typically when the claim has expired, or the deadline reached.

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+        rm(self.task_dir)
+        makedirs(self.task_dir)
+        self.reclaim_fut = event_loop.create_task(self.reclaim_task())
+        self._reclaim_task = {}
+        self.task_fut = event_loop.create_task(self.run_task())
+
+    @property
+    def task_credentials(self):
+        """Return the temporary credentials returned from [re]claimWork."""
+        return self._reclaim_task.get("credentials", self.claim_task["credentials"])
```
tomprince, line 92: It think it might make sense to construct a queue here. Or, perhaps even better, rather than store both the the `claimTask` and `reclaimTask` responses, create a queue from them and store the queue as an attribute. This would probably necessitate sharing a session object, but it seems reasonable to share a session anyway.

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+                Defaults to ``True``.
+
+        """
+        with open(self.log_path, "a") as log_fh:
+            print("{} {} - {}".format(arrow.utcnow().format(self.config["log_datefmt"]), logging._levelToName.get(level, str(level)), msg % args), file=log_fh)
+            worker_log and log.log(level, "%s:%s - {}".format(msg), self.task_id, self.run_id, *args)
+
+    async def download_uuids(self):
+        """Download the UUID manifest."""
+        url = self.claim_task["task"]["payload"]["uuid_manifest"]
```
tomprince: Should this be a URL, or a task reference?

I'm also curious if these UUIDs are at all sensitive; if they are, they should probably be in a private artifact, and not logged.

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+        self.task_id = claim_task["status"]["taskId"]
+        self.run_id = claim_task["runId"]
+        self.claim_task = claim_task
+        self.event_loop = event_loop or asyncio.get_event_loop()
+        self.task_dir = os.path.join(config["work_dir"], self.task_id)
+        self.log_path = os.path.join(self.task_dir, "live_backing.log")
+        self.poll_log_path = os.path.join(self.task_dir, "polling.log")
+        rm(self.task_dir)
+        makedirs(self.task_dir)
+        self.reclaim_fut = event_loop.create_task(self.reclaim_task())
```
tomprince: My experience suggests that constructors should not trigger async stuff.

notarization_poller/src/notarization_poller/worker.py
```diff
@@ -0,0 +1,128 @@
...
+        log.info("SIGTERM received; shutting down")
+        nonlocal done
+        done = True
+        await running_tasks.cancel()
+
+    async def _handle_sigusr1():
+        """Stop accepting new tasks."""
+        log.info("SIGUSR1 received; no more tasks will be taken")
+        nonlocal done
+        done = True
```
tomprince: I don't think this does anything, as it appears that `RunTasks.invoke` runs forever.

notarization_poller/src/notarization_poller/worker.py
```diff
@@ -0,0 +1,128 @@
...
+                        options={
+                            "credentials": {"accessToken": self.config["taskcluster_access_token"], "clientId": self.config["taskcluster_client_id"]},
+                            "rootUrl": self.config["taskcluster_root_url"],
+                        },
+                        session=session,
+                    )
+                    new_tasks = await self._run_cancellable(claim_work(self.config, queue, num_tasks=num_tasks_to_claim))
+                self.last_claim_work = arrow.utcnow()
+                for claim_task in new_tasks.get("tasks", []):
+                    new_task = Task(self.config, claim_task)
```
tomprince, line 78: If `Task` has a `.start` method that returns a future, you can add a `done_callback` to remove it from running_tasks, rather than periodically pruning it.

notarization_poller/src/notarization_poller/worker.py
```diff
@@ -0,0 +1,128 @@
...
+                    self.running_tasks.append(new_task)
+            await self.prune_running_tasks()
+            sleep_time = self.last_claim_work.timestamp + self.config["claim_work_interval"] - arrow.utcnow().timestamp
+            sleep_time > 0 and await self._run_cancellable(sleep(sleep_time))
```
tomprince, line 83: I *think* if the body of this functions is wrapped in an `asyncio.Task`, then you don't need `_run_cancellable`, and cancelling the task will propagate to the awaited futures here.

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,321 @@
...
+                    temp_queue = Queue(options={"credentials": self.task_credentials, "rootUrl": self.config["taskcluster_root_url"]}, session=session)
+                    self.reclaim_task = await temp_queue.reclaimTask(self.task_id, self.run_id)
+            except taskcluster.exceptions.TaskclusterRestFailure as exc:
+                if exc.status_code == 409:
+                    log.warning("Stopping task %s %s after receiving 409 response from reclaim_task: %s %s", self.task_id, self.run_id)
+                    exit_status = STATUSES["superseded"]
+                else:
+                    log.exception("reclaim_task unexpected exception: %s %s", self.task_id, self.run_id)
+                    exit_status = STATUSES["internal-error"]
+                await self.stop(status=exit_status)
```
tomprince: i *think* it makes sense to instead cancel the `run_task` future here, and handle `CancelledError` there to handle uploading logs, and resolving the task.

tomprince (changes requested):
I've left a bunch of comments on the implementation with suggestions on some restructuring that I think would make the code easier to reason about, and in one case make the notarization code clearly separate from the taskcluster worker code.

It is probably worth chatting about this before addressing the comments.

notarization_poller/src/notarization_poller/constants.py
escapewindow: Possibly. `taskcluster-python` doesn't directly deal with workers. No tc client library currently contains exit code to taskcluster task status maps. I'm inclined to leave it in scriptworker_client and scriptworker until we either a) create a shared python taskcluster worker library, or b) consensus is we should have an exit code to taskcluster task status map in each of the taskcluster client libraries.

notarization_poller/src/notarization_poller/task.py
escapewindow: No, we should have a config option that is significantly less than the `takenUntil`, or a slightly delayed `reclaimTask` will result in a `claim-expired` task that gets rerun. This is the same model as scriptworker.

notarization_poller/src/notarization_poller/task.py
escapewindow: OSX Keychains are a PITA to deal with. I think it's possible, and we decided it wasn't worth it. It's possible someone can put in the time to figure out how to make it seamless with the right steps. That work has not been done yet, nor do I think it's that important. If someone can get on these boxes to observe the password, it's already game over.

notarization_poller/src/notarization_poller/task.py
escapewindow: I'm not sure what the benefit of that is, since we're using 0 and 1 for the statuses. Unless you mean `STATUSES['success']` instead of 0 ?

notarization_poller/src/notarization_poller/task.py
escapewindow: I believe we use it in scriptworker integration tests, and I left it in case we use them in notarization poller integration tests.

notarization_poller/src/notarization_poller/task.py
escapewindow: We have a `while` with a `for` nested. We want to break out of the outer loop as well as the inner loop.

notarization_poller/src/notarization_poller/task.py
escapewindow: Hm. I see where you're coming from here. Making `uuids` a property of the object allows us to inspect progress externally, should we want to (f/e, if we wanted to log a message if the number of pending UUIDs across all tasks for the worker exceeded a certain threshold). That doesn't mean we can't break the dictionary up into `self.uuids` and `self.pending_uuids` if we want to allow for external inspection without the dictionary.

notarization_poller/src/notarization_poller/task.py
tomprince: I agree that the time used should be less than `takenUntil`, not equal to it. But, since we ignore the value that taskcluster returns, we could in fact have an interval that is past `takenUntil` time. Presumably this would require a change to the taskcluster configuration, but there isn't any reason we can't be resilient to changes to that. If we do have a config, it should be for how long before `takenUntil` we should reclaim, rather than how long between reclaims, or something along those lines.

notarization_poller/src/notarization_poller/task.py
escapewindow: <s>I'm under the impression sets are already sorted. No?</s>

Looks like no:

```python
>>> a = [9, 2, 1]
>>> b = set(a)
>>> for c in b:
...     print(c)
...
9
2
1
```

tomprince: On recent versions of python, sets and dictionaries preserve insertion order.

notarization_poller/src/notarization_poller/task.py
escapewindow: I thought about that. However, in [this comment](https://github.com/mozilla-releng/scriptworker-scripts/pull/121#issue-352187022), I explicitly stated "It looks like we could potentially abstract away some of this logic into a base python worker module, and share it with scriptworker. I've explicitly decided not to do that this time around." The whole question around how many tasks to claim and whether to run them concurrently or serially is itself fairly complex if we want to make it support any type of behavior. I'm wary that making it generic before we actually have a real use case for it is likely YAGNI.

notarization_poller/src/notarization_poller/task.py
escapewindow: I avoided creating a session and queue and passing it around, because it took a lot of effort in scriptworker to do so, and I'm not sure we have a clear win. Aiohttp wants to force us to use `async with aiohttp.ClientSession()` which means a long lived generator or having to hack around those efforts (we took the latter approach in scriptworker, and have had to update our hacks every time aiohttp changes their rules).

notarization_poller/src/notarization_poller/task.py
escapewindow: Afaik they're not sensitive, and can't be accessed without our Apple account credentials.

In the taskgraph patches, I used the task reference to create an artifact url.

notarization_poller/src/notarization_poller/task.py
escapewindow: How should we do this? Create a future in `RunTasks.invoke` after creating the Task?

notarization_poller/src/notarization_poller/task.py
tomprince: I definitely don't think we should try to make it re-usable. I do think having a separation of concerns makes both parts easier to reason about, though.

notarization_poller/src/notarization_poller/worker.py
escapewindow: We cancel the task, and want to prevent the `while not done` from looping again. It's possible that while porting it from scriptworker I pruned too much or too little, but I'm not seeing the problem currently.

notarization_poller/src/notarization_poller/task.py
escapewindow: The current config-driven reclaim interval mirrors what we do in scriptworker, which is what the taskcluster team told me to do when I first wrote it. I'm inclined to leave that in place for now.

notarization_poller/src/notarization_poller/worker.py
tomprince: I don't see anything that cancels a task in `_handle_sigusr1` (I do in `_handle_sigterm`). I also don't see anything in `RunTasks.invoke` that requires the loop here. I think this function should probably call something on `RunTasks` that causes it to break out of the loop, and the outer loop in `main` can be removed.

notarization_poller/src/notarization_poller/task.py
tomprince: Looking at the taskcluster maintained workers:
- docker-worker [reclaims](https://github.com/taskcluster/docker-worker/blob/b089a1dbf5aed5b9cd1f70fcf25f1e92ff05a15a/src/lib/task.js#L183) at 3/4 of the time between now and `takenUntil` (`reclaimDivisor` is 1.3 and 1/1.3 is approximately 76%)
- generic-worker [reclaims](https://github.com/taskcluster/generic-worker/blob/5fd8e48d63d2f7a7eeb6c9a65d4c4d1b41d535b7/taskstatus.go#L340) 3m before `takenUntil`, with some protection from hitting the queue too often

I think we should do something similar here (and eventually update scriptworker). I don't know if the advice changed, or maybe taskcluster didn't used to report claim times and so picking an interval made sense when scriptworker was originally written.

notarization_poller/src/notarization_poller/constants.py
escapewindow: Resolving, since this is in `scriptworker_client` now, and we can still potentially move it to the tc-client.py later.

notarization_poller/src/notarization_poller/config.py
escapewindow: I ended up calling `_init_logging` in this function :)

notarization_poller/src/notarization_poller/task.py
escapewindow: Fixed.

notarization_poller/src/notarization_poller/worker.py
escapewindow: I think I've done this in the latest working patchset.

notarization_poller/src/notarization_poller/task.py
escapewindow: +1.

escapewindow:
Tom's wip [here](https://github.com/mozilla-releng/scriptworker-scripts/commit/fae37f8e48fae5e9beebe51c7b017fc5e6383324). I'll push up my latest changes here after I have a chance to clean them up. I haven't split the queue interactions from the apple interactions, but I may do so to support integration tests better (since we can easily "mock" apple interactions out by instantiating a different class for the task logic).

notarization_poller/src/notarization_poller/task.py
escapewindow: Tracking followup in https://github.com/mozilla-releng/scriptworker/issues/424 .

notarization_poller/src/notarization_poller/worker.py
escapewindow: Addressed in https://github.com/escapewindow/scriptworker-scripts/commit/bc304c110186ed77ea392408e1037e472142636a .

notarization_poller/src/notarization_poller/task.py
escapewindow: Removed.

notarization_poller/src/notarization_poller/task.py
escapewindow: Removed.

notarization_poller/src/notarization_poller/task.py
escapewindow: Went with `STATUSES["success"]` (incoming revision)

notarization_poller/src/notarization_poller/task.py
escapewindow: I went with an exception, which breaks us out of both loops :)

notarization_poller/src/notarization_poller/task.py
escapewindow: I did end up using `self.pending_uuids`.

notarization_poller/src/notarization_poller/task.py
escapewindow: added a `sorted()` in an incoming revision.

notarization_poller/src/notarization_poller/task.py
escapewindow: This assumes we're using a function and not a class. We can still go that route if wanted later, but not in this PR.

notarization_poller/src/notarization_poller/task.py
escapewindow: Moved.

notarization_poller/src/notarization_poller/task.py
escapewindow: This is probably fodder for followup.

notarization_poller/src/notarization_poller/task.py
escapewindow: I think the refactored code is better and more easily reasoned about. I haven't split the object into two.

notarization_poller/src/notarization_poller/worker.py
escapewindow: I haven't done this. The current approach seems to be working. Not sure if this was a blocker?

escapewindow:
I'm planning on merging and deploying this. We can follow up with any needed changes.

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,267 @@
...
+            self.status = STATUSES["resource-unavailable"]
+            self.task_log(traceback.format_exc(), level=logging.CRITICAL)
+        except (DownloadError, RetryError):
+            self.status = STATUSES["intermittent-task"]
+            self.task_log(traceback.format_exc(), level=logging.CRITICAL)
+        except TaskError:
+            self.status = STATUSES["malformed-payload"]
+            self.task_log(traceback.format_exc(), level=logging.CRITICAL)
+        log.info("Stopping task %s %s with status %s", self.task_id, self.run_id, self.status)
+        self.reclaim_fut and self.reclaim_fut.cancel()
```
tomprince: I don't think the first part of `future and future.cancel()` has any effect. It is set unconditionally above, is set nowhere else, and `Future` does not appear to have a `__nonzero__`.  That is, you can drop the `future and` part here.

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,267 @@
...
+        self.task_fut and self.task_fut.cancel()
```
tomprince: I *think* this is superfluous, as we've already awaited `task_fut`. I think there is no way the future is not complete here.

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,267 @@
...
+                    temp_queue = Queue(options={"credentials": self.task_credentials, "rootUrl": self.config["taskcluster_root_url"]}, session=session)
+                    self._reclaim_task = await temp_queue.reclaimTask(self.task_id, self.run_id)
+            except taskcluster.exceptions.TaskclusterRestFailure as exc:
+                if exc.status_code == 409:
+                    log.warning("Stopping task after receiving 409 response from reclaim_task: %s %s", self.task_id, self.run_id)
+                    self.status = STATUSES["superseded"]
+                else:
+                    log.exception("reclaim_task unexpected exception: %s %s", self.task_id, self.run_id)
+                    self.status = STATUSES["internal-error"]
+                self.task_fut and self.task_fut.cancel()
```
tomprince, line 122: I'd be slightly inclined to switch the creation of `task_fut` and `reclaim_fut` above, and drop the `self.task_fut and` here, but probably not important.

notarization_poller/src/notarization_poller/task.py
```diff
@@ -0,0 +1,267 @@
...
+        self.main_fut = self.event_loop.create_task(self.async_start())
+
+    async def async_start(self):
+        """Async start the task."""
+        self.reclaim_fut = self.event_loop.create_task(self.reclaim_task())
+        self.task_fut = self.event_loop.create_task(self.run_task())
+
+        try:
+            await self.task_fut
+        except Download404:
```
tomprince, line 70: I think this probably should deal with `asyncio.CancelledError` which will be raised (at least) if reclaim fails.

notarization_poller/src/notarization_poller/worker.py
tomprince: I still see `_run_cancellable` here. It exists to propagate cancellation to the `sleep` and `claim_work` calls. If you don't have that function, and cancel the future from `invoke`, it should automatically propagate the cancellation to the `sleep` and `claim_work` calls.

That said, this seems like something that can be left for a followup.

notarization_poller/src/notarization_poller/worker.py
```diff
@@ -0,0 +1,152 @@
...
+    async def _handle_sigterm():
+        log.info("SIGTERM received; shutting down")
+        nonlocal done
+        done = True
+        await running_tasks.cancel()
+
+    async def _handle_sigusr1():
+        """Stop accepting new tasks."""
+        log.info("SIGUSR1 received; no more tasks will be taken")
+        running_tasks.is_stopped = True
```
tomprince, line 139: There is nothing async and `add_signal_handler` doesn't do anything with a returned future, so this function may as well be sync.

notarization_poller/src/notarization_poller/worker.py
```diff
@@ -0,0 +1,152 @@
...
+def main(event_loop=None):
+    """Notarization poller entry point: get everything set up, then enter the main loop.
+
+    Args:
+        event_loop (asyncio.BaseEventLoop, optional): the event loop to use.
+            If None, use ``asyncio.get_event_loop()``. Defaults to None.
+
+    """
+    event_loop = event_loop or asyncio.get_event_loop()
+    done = False
```
tomprince: `done` appears to be set an never looked at, so can be removed.

notarization_poller/src/notarization_poller/worker.py
```diff
@@ -0,0 +1,152 @@
...
+    log.info("Worker FQDN: {}".format(socket.getfqdn()))
+    rm(config["work_dir"])
+    makedirs(config["work_dir"])
+    running_tasks = RunTasks(config)
+
+    async def _handle_sigterm():
+        log.info("SIGTERM received; shutting down")
+        nonlocal done
+        done = True
+        await running_tasks.cancel()
```
tomprince, line 134: Since `add_signal_handler` does handle futures, I think it would be better if `_handle_sigterm` were sync and handled anything that needed to be async itself.

That said, it looks like it async so that `running_tasks.cancel` can wait on the task futures. Since nothing is waiting on the result of that, I don't think that await actually has any effect. Thus, I think this function and `running_tasks.cancel` can both be sync.

https://github.com/mozilla-releng/scriptworker-scripts/commit/fae37f8e48fae5e9beebe51c7b017fc5e6383324#diff-5fb318e79005ec63cfed70064fa1e861R71 is where I handled waiting on all of those tasks to finish in my sketch

escapewindow: I mainly copied over [these coroutines](https://github.com/mozilla-releng/scriptworker/blob/master/src/scriptworker/worker.py#L236). I'm almost leaning towards tearing out this code, since the worst that can happen with `notarization_poller` dying is we wait a bit longer on tasks that just poll and wait. Do you have a preference between tearing these out and leaving them as-is?

notarization_poller/src/notarization_poller/worker.py
escapewindow: Done

notarization_poller/src/notarization_poller/task.py
escapewindow: Done

notarization_poller/src/notarization_poller/task.py
escapewindow: Done

notarization_poller/src/notarization_poller/task.py
escapewindow: Done

notarization_poller/src/notarization_poller/worker.py
escapewindow: This is also likely leave as-is or tear out.

notarization_poller/src/notarization_poller/task.py
```diff
@@ -76,9 +76,12 @@ def start(self):
         except TaskError:
             self.status = STATUSES["malformed-payload"]
             self.task_log(traceback.format_exc(), level=logging.CRITICAL)
+        except asyncio.CancelledError:
+            # We already dealt with self.status in reclaim_task
+            self.task_log(traceback.format_exc(), level=logging.CRITICAL)
```
tomprince, line 81: (followup) We may want to consider not logging here, since we'll have already logged elsewhere. It probably isn't worth worrying about until after this has been deployed and is in production for a while though.

notarization_poller/src/notarization_poller/task.py
tomprince: I still see `self.task_fut.cancel()` here. I think it is impossible for `self.task_fut` to not be resolved here (we `await`ed above, so this code won't execute until it has been resolved (either succesfully or by raising)), in which case calling `.cancel` on it does nothing.

notarization_poller/src/notarization_poller/worker.py
tomprince: I'm fine either way. My main concerns are:
1) the code is more complicated than it needs to be
2) the code makes it look like it will wait for the futures to complete when it likely won't. (Whether they will or not depends on the order of callback in the event loop, and may also depend on how much (if any) work they do in being cancelled)

Long-term, I think we should improve the code here and in scriptworker to handle this case well. But, I'm fine landing without this code or as-is.

notarization_poller/src/notarization_poller/worker.py
tomprince: The change here is removing the `async` on the function def here, and the `ensure_future` (and possibly the `lambda`) below. I think handling graceful restarts is valuable enough to keep.

notarization_poller/src/notarization_poller/worker.py
```diff
@@ -0,0 +1,149 @@
...
+        """Prune any complete tasks from ``self.running_tasks``."""
+        for task in self.running_tasks:
+            if task.complete:
+                self.running_tasks.remove(task)
+
+    async def _run_cancellable(self, coroutine: typing.Awaitable):
+        if not self.is_cancelled:
+            self.future = asyncio.ensure_future(coroutine)
+            result = await self.future
+            self.future = None
```
tomprince: I think this may want to be in a `finally` block (with the `await` in the try` block).

notarization_poller/src/notarization_poller/worker.py
```diff
@@ -0,0 +1,149 @@
...
+        self.last_claim_work = arrow.get(0)
+        self.is_stopped = False
+        self.is_cancelled = False
+        self.future = None
+
+    async def invoke(self):
+        """Claims and processes Taskcluster work."""
+        while not self.is_cancelled and not self.is_stopped:
+            num_tasks_to_claim = min(self.config["max_concurrent_tasks"] - len(self.running_tasks), MAX_CLAIM_WORK_TASKS)
+            if num_tasks_to_claim > 0 and arrow.utcnow().timestamp - self.last_claim_work.timestamp >= self.config["claim_work_interval"]:
```
tomprince: Is the timestamp part of this ever going be false? We always sleep for an amount of time that will make the timestamp part true.

notarization_poller/src/notarization_poller/worker.py
```diff
@@ -0,0 +1,149 @@
...
+                        session=session,
+                    )
+                    new_tasks = await self._run_cancellable(claim_work(self.config, queue, num_tasks=num_tasks_to_claim)) or {}
+                self.last_claim_work = arrow.utcnow()
+                for claim_task in new_tasks.get("tasks", []):
+                    new_task = Task(self.config, claim_task)
+                    new_task.start()
+                    self.running_tasks.append(new_task)
+            await self.prune_running_tasks()
+            sleep_time = self.last_claim_work.timestamp + self.config["claim_work_interval"] - arrow.utcnow().timestamp
```
tomprince, line 82: (potential followup) Since we don't do anything async[1] between setting `last_claim_work` and now, I suspect we could use `claim_work_interval` directly here, but it doesn't hurt to do this calculation. (It would be slightly more interesting if we took the time *before* calling claim work, so it was the time between calls to claim work)

[1] We await `prune_running_tasks`, so it may take several event loop iterations, but we won't wait on I/O or anything.

tomprince (approved):
One minor comment (about handling graceful shutdown well), and a number of possible followups.

notarization_poller/src/notarization_poller/worker.py
escapewindow: RIght. My main concerns are: I have done a significant amount of testing with the code as is. Making non-trivial changes has the likelihood of introducing new errors, just as I'm about to roll out.

notarization_poller/src/notarization_poller/task.py
escapewindow: Done

PR #122, opened 2019-12-12
add `notarize_{1,3}_behavior`s

escapewindow:
This is probably safe to merge as-is, since it shouldn't break backwards compatibility. I'd like to get in some testing before that.

@nthomas-mozilla do you want to take a preliminary look?


escapewindow:
Moving to review-ready before I've fully tested, since I'm a ways away from in-tree patches. I'm hoping to avoid the situation where I'm all ready to land, but am missing review during the holidays. I can try to keep track of the interdiff between the reviewed revision and any new changes.

iscript/tests/test_mac.py
```diff
@@ -995,6 +995,103 @@ def fake_get_notarization_status_from_log(path):
     await mac.notarize_behavior(config, task)
 
 
+# notarize_1_behavior {{{1
+@pytest.mark.parametrize("notarize_type,use_langpack", zip(("multi_account", "single_account", "single_zip"), (False, True)))
```
nthomas-mozilla: I suspect this was supposed to be `itertools.product()` over the behaviors and langpack on/off, to test all the combinations of parameters. Originally landed in 8ab0366 to test `notarize_behavior` and doesn't currently cover the `single_zip` case. Do you agree ?

iscript/tests/test_mac.py
```diff
@@ -995,6 +995,103 @@ def fake_get_notarization_status_from_log(path):
...
+        mocker.patch.object(mac, "sign_langpacks", new=noop_async)
+        task["payload"]["upstreamArtifacts"].append({"taskId": "task3", "formats": ["autograph_langpack"], "paths": ["public/build3/target.langpack.xpi"]})
+
+    mocker.patch.object(os, "listdir", return_value=[])
+    mocker.patch.object(mac, "run_command", new=noop_async)
+    mocker.patch.object(mac, "unlock_keychain", new=noop_async)
+    mocker.patch.object(mac, "get_bundle_executable", return_value="bundle_executable")
+    mocker.patch.object(mac, "poll_notarization_uuid", new=noop_async)
+    mocker.patch.object(mac, "get_app_dir", return_value=os.path.join(work_dir, "foo/bar.app"))
+    mocker.patch.object(mac, "get_notarization_status_from_log", return_value=None)
```
nthomas-mozilla: Nit, probably no need to mock `poll_notarization_uuid` or `get_notarization_status_from_log` for this _1 behavior.

nthomas-mozilla (approved):
Looks good to me, just a couple of minor tweaks.

escapewindow:
Thanks!

iscript/src/iscript/mac.py
```diff
@@ -1031,6 +1053,100 @@ def get_notarization_status_from_log(log_path):
...
+
+    Args:
+        config (dict): the running configuration
+        task (dict): the running task
+
+    Raises:
+        IScriptError: on fatal error.
+
+    """
+    all_paths = get_app_paths(config, task)
```
nthomas-mozilla, line 1128: Initially I didn't click that `all_paths` has a different meaning here than for `notarize_behavior` and `notarize_1_behavior`, notably it has separate entries for pkg and apps rather than just apps, which then leads to the changes to `_get_pkg_name_from_tarball` and so on. A comment pointing this out might be useful for the future.

nthomas-mozilla (approved):

iscript/src/iscript/mac.py
escapewindow: Done.

escapewindow:
Thanks!

PR #123, opened 2019-12-12
Add dirschema app config for configloader

rail:


rail:
Not quite sure what to do with `Dockerfile`, which is unused here.

bhearsum:
If Dockerfile is unused, then this doesn't meet the `app` spec. Are you sure it's not a tool or a lib instead?

rail:
Oh, I should switch the schema then

bhearsum (approved):

PR #126, opened 2019-12-30
Fix l10n_bump when the tree is closed (bug 1604476)

jcristau:
do_actions expects l10n_bump to return an int, not None.

### DIFF
--- a/signingscript/tests/test_sign.py
+++ b/signingscript/tests/test_sign.py
@@ -6,6 +6,7 @@ import os.path
 import re
 import shutil
 import subprocess
+import sys
 import tarfile
 import zipfile
 from contextlib import contextmanager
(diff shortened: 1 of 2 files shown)


treescript/src/treescript/l10n.py
```diff
@@ -220,7 +220,7 @@ def build_commit_message(name, locale_map, dontbuild=False, ignore_closed_tree=F
     if not ignore_closed_tree:
```
MihaiTabara, line 216: Nit: we should also perhaps move some of these assignments after this if, to avoid un-necessary operations if we have to return fast. Lines 214, 216, 217 and 218 can come afterwards.

MihaiTabara: Even though this isn't much of a time-saver, I believe it's better for sanity check.
Hijacked @jcristau's PR to address this. Mea culpa for this.

signingscript/tests/test_sign.py
```diff
@@ -6,6 +6,7 @@
 import re
```
MihaiTabara, line 6: Backported @tomprince's solution to fix this from https://github.com/mozilla-releng/scriptworker-scripts/pull/125/ in order to unblock the l10n bumper fixing.

Callek (approved):

PR #129, opened 2020-01-03
[treescript] Change how actions are handled.

tomprince:
- Remove the deprecated support for specifying actions by scope. This is not used on esr68, so can be removed.
- Don't treat `push` as an action. See [the commit message](https://github.com/mozilla-releng/scriptworker-scripts/commit/ec98cf61d4146653b63100e5952f92b64fa8580d) for the rational.
- Treat actions as a set rather than a list; we currently sort the actions in a specific order, so that when we loop over the list of actions, they run in the right order. This changes it to just run each action in the right order, if the action is specified in the set of actions.

### DIFF
--- a/treescript/src/treescript/script.py
+++ b/treescript/src/treescript/script.py
@@ -30,17 +30,12 @@ async def do_actions(config, task, actions, repo_path):
     """
     await checkout_repo(config, task, repo_path)
     num_changes = 0
-    for action in actions:
-        if action in ["tagging", "tag"]:
-            num_changes += await do_tagging(config, task, repo_path)
-        elif "version_bump" == action:
-            num_changes += await bump_version(config, task, repo_path)
-        elif "l10n_bump" == action:
-            num_changes += await l10n_bump(config, task, repo_path)
-        elif "push" == action:
-            pass  # handled after log_outgoing
-        else:
-            raise NotImplementedError("Unexpected action")
+    if "tag" in actions:
+        num_changes += await do_tagging(config, task, repo_path)
+    if "version_bump" in actions:
+        num_changes += await bump_version(config, task, repo_path)
+    if "l10n_bump" in actions:
+        num_changes += await l10n_bump(config, task, repo_path)
     num_outgoing = await log_outgoing(config, task, repo_path)
     if num_outgoing != num_changes:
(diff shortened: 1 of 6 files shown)


treescript/src/treescript/task.py
```diff
@@ -193,39 +186,35 @@ def get_ignore_closed_tree(task):
...
-        log.warning("Scopes-based actions are deprecated! Use task.payload.actions instead.")
-        actions = [s.split(":")[-1] for s in task["scopes"] if s.startswith(config["taskcluster_scope_prefix"] + "action:")]
-        if len(actions) < 1:
-            raise TaskVerificationError("Need at least one valid action specified in scopes")
+    actions = set(task["payload"].get("actions", []))
     log.info("Action requests: %s", actions)
-    invalid_actions = set(actions) - set(VALID_ACTIONS)
+    invalid_actions = actions - VALID_ACTIONS
     if len(invalid_actions) > 0:
         raise TaskVerificationError("Task specified invalid actions: {}".format(invalid_actions))
```
escapewindow, line 206: It looks like we no longer have coverage for this line; could you please add it?

treescript/src/treescript/script.py
```diff
@@ -30,29 +30,20 @@
...
-            num_changes += await do_tagging(config, task, repo_path)
-        elif "version_bump" == action:
-            num_changes += await bump_version(config, task, repo_path)
-        elif "l10n_bump" == action:
-            num_changes += await l10n_bump(config, task, repo_path)
-        elif "push" == action:
-            pass  # handled after log_outgoing
-        else:
-            raise NotImplementedError("Unexpected action")
+    if "tag" in actions:
```
escapewindow, line 33: I think this is a good set of changes overall.

We had been discussing moving treescript to behaviors rather than actions, which makes sense because certain combinations of actions aren't intended to run in the same task. We should move the logic of which operations to run, and in which order, to the behavior function. I think these changes can land in the meantime, though. Please be mindful of the merge day PR.

escapewindow (approved):

treescript/src/treescript/script.py
tomprince: Yeah, this (partly) grew out of changes I was thinking of suggesting on that PR.

PR #131, opened 2020-01-08
Bug 1607266 - set backgroundRate and forceFallBack as optional

MihaiTabara:
We make the new payload vars optional: https://hg.mozilla.org/releases/mozilla-beta/rev/0b1cde02d80a#l2.71

But we expect them to exist: https://github.com/mozilla-releng/scriptworker-scripts/pull/116/files#diff-c94ab011a5da8654fdb1da7508514dbdR101

### DIFF
--- a/balrogscript/src/balrogscript/script.py
+++ b/balrogscript/src/balrogscript/script.py
@@ -98,8 +98,8 @@ def schedule(task, config, auth0_secrets):
         task["payload"]["version"],
         task["payload"]["build_number"],
         task["payload"]["publish_rules"],
-        task["payload"]["release_eta"] or None,  # Send None if release_eta is ''
         task["payload"].get("force_fallback_mapping_update", False),
+        task["payload"]["release_eta"] or None,  # Send None if release_eta is ''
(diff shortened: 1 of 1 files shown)


balrogscript/src/balrogscript/script.py
```diff
@@ -98,9 +98,9 @@ def schedule(task, config, auth0_secrets):
         task["payload"]["version"],
         task["payload"]["build_number"],
         task["payload"]["publish_rules"],
-        task["payload"]["force_fallback_mapping_update"] or False,
         task["payload"]["release_eta"] or None,  # Send None if release_eta is ''
-        task["payload"]["background_rate"] or None,
+        task["payload"].get("force_fallback_mapping_update", False),
+        task["payload"].get("background_rate"),
```
bhearsum, line 103: Looks like you're changing the order these args are passed - that's probably going to break something :)

MihaiTabara: You're right, mea culpa. Fixed now.

bhearsum (approved):

PR #132, opened 2020-01-14
Bump dependencies to use scritpworker 29.1.0

JohanLorenzo:
Follows https://github.com/mozilla-releng/scriptworker/pull/420 up.

I ran `./maintenance/pin.sh` to get this result.


escapewindow (approved):
Thanks! We need to track down the test bustage. I’ll try to take a closer look later.

escapewindow:
Hm. Addonscript tests are green for me locally. Digging deeper...

PR #138, opened 2020-01-20
Bug 1610291 - skip commits outside of default for automated betas.

MihaiTabara:
hg.mozilla.org retrieves commits from all branches when queried `from` - `to`. For some repositories (such as esr68) that includes relbranches commits which "pollute" what shipitscript should analyze in order to start a proper beta.

Earlier today we accidentally promoted a Fennec beta release based off a wrong revision. This is to fix this for future.

### DIFF
--- a/shipitscript/src/shipitscript/pushlog_scan.py
+++ b/shipitscript/src/shipitscript/pushlog_scan.py
@@ -33,7 +33,9 @@ def get_shippable_revision_build(branch, last_shipped_rev, cron_rev):
     # TODO use redo here to automatically retry
     resp = requests.get(URL.format(branch=branch), params={"full": "1", "fromchange": last_shipped_rev, "tochange": cron_rev})
     pushlog = resp.json()
-    reversed_pushes = [pushlog[k] for k in sorted(pushlog.keys(), reverse=True)]
+    # we strip out all the pushes that contain other branches' commits (e.g. relbranch)
+    reversed_pushes = [pushlog[k] for k in sorted(pushlog.keys(), reverse=True)
+                        if pushlog[k]['changesets'][0]['branch'] == 'default']
     for push in reversed_pushes:
         push_importance = is_push_important(push)
         if push_importance == Importance.IMPORTANT:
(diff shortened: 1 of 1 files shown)


shipitscript/src/shipitscript/pushlog_scan.py
```diff
@@ -33,7 +33,9 @@ def get_shippable_revision_build(branch, last_shipped_rev, cron_rev):
     # TODO use redo here to automatically retry
     resp = requests.get(URL.format(branch=branch), params={"full": "1", "fromchange": last_shipped_rev, "tochange": cron_rev})
```
MihaiTabara: Note: including `branch=default` or something in the params doesn't work, I've already tried in:

https://hg.mozilla.org/releases/mozilla-esr68/json-pushes?full=1&branch=default&fromchange=ba35b4e926ed8da2b7796272948d9d86cfec8f8d&tochange=615d4ace089d57ddf11facdb678188fb41c0981b

MihaiTabara, line 34: Note: including `branch=default` or something in the params doesn't work, I've already tried in:

https://hg.mozilla.org/releases/mozilla-esr68/json-pushes?full=1&branch=default&fromchange=ba35b4e926ed8da2b7796272948d9d86cfec8f8d&tochange=615d4ace089d57ddf11facdb678188fb41c0981b

shipitscript/src/shipitscript/pushlog_scan.py
```diff
@@ -33,7 +33,9 @@ def get_shippable_revision_build(branch, last_shipped_rev, cron_rev):
...
     pushlog = resp.json()
-    reversed_pushes = [pushlog[k] for k in sorted(pushlog.keys(), reverse=True)]
+    # we strip out all the pushes that contain other branches' commits (e.g. relbranch)
+    reversed_pushes = [pushlog[k] for k in sorted(pushlog.keys(), reverse=True)
+                        if pushlog[k]['changesets'][0]['branch'] == 'default']
```
MihaiTabara: I've tested locally against beta as well and works smooth, no changes, as expected.

shipitscript/src/shipitscript/pushlog_scan.py
rail: Yeah, according to [the docs](https://hg.mozilla.org/hgcustom/version-control-tools/file/tip/docs/hgmo/pushlog.rst) `branch` is not a valid parameter.

rail (approved):

Callek:
@MihaiTabara I should note, here, that it is possible to have a push where we have both default and relbranch csets in the same push. Its not common for our use case, but certainly possible.

I don't necessarily think its important to handle though.

Callek (approved):

shipitscript/src/shipitscript/pushlog_scan.py
MihaiTabara: TIL about those docs, thanks for pointing it out 👍

PR #142, opened 2020-01-30
signingscript: Add adhoc signing support

JohanLorenzo:


MihaiTabara (approved):
LGTM! We need to add these definitions as well in SOPS and cloudops infra in order to be able to use these workers.

JohanLorenzo:
Sounds good! Please let me know whenever SOPS is updated! 🙂

signingscript/docker.d/passwords.yml
```diff
@@ -133,6 +134,20 @@ in:
              ["autograph_authenticode_ev"]
             ]
 
+      # dep-passwords-adhoc.json
+      '(ENV == "dev" || ENV == "fake-prod") && COT_PRODUCT == "adhoc"':
```
escapewindow, line 138: Hm. (Brainstorming for long-term.)

We could create single file with a bunch of sections with product and cert level. For example, maybe this means we have `project:adhoc:releng:signing:cert:firefox-dep-signing` and `project:adhoc:releng:signing:cert:mobile-release-signing`? That might mean creating a `cert-prefix` that defaults to `""` but is `firefox-` or `mobile-` in those two cases for adhoc. Or maybe we move to having the product in the cert name in general, across all products.

OR,

Instead of creating a `passwords.yml`, maybe we could create a `passwords-PRODUCT-TRUSTLEVEL.yml` file as appropriate. We could allow for a list of these, and search each of them for the appropriate cert(s), or allow specifying which one to use in the task definition or something?

rail: +1, `passwords.yml` needs some refactoring

MihaiTabara: Yeah, I think we can definitely do a better job in `passwords.yml`. Main focus initially was to import existing logic from the previous puppet confgs.

PR #145, opened 2020-02-17
Bug 1572316 - Treescript: Translate Thunderbird repos to treestatus trees

jfx2006:
check_treestatus in l10n.py fails for Thunderbird repositories because
the names in Treestatus are different than expected.

The get_treestatus_name function is based on how Treeherder determines
TreeStatus names for Thunderbird repos.

### DIFF
--- a/treescript/src/treescript/l10n.py
+++ b/treescript/src/treescript/l10n.py
@@ -139,6 +139,24 @@ def build_commit_message(name, locale_map, dontbuild=False, ignore_closed_tree=F
     return message
 
 
+# get_treestatus_name {{{1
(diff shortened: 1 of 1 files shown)


jfx2006:
How Treeherder handles this:
https://github.com/mozilla/treeherder/blob/65b7f4ab451fcb8bba49f5e41963c81be2fa9cdb/ui/models/treeStatus.js#L4-L11

tomprince:
I chatted with @jfx2006 about this briefly. I think the plan is to change the treestatus tree names, so that the comm trees are no longer exceptional. Feel free to re-open if that is not the case, or this is still required.

PR #146, opened 2020-02-19
Py38

escapewindow:
I'm kind of leaning towards always including the `pyVER` in the `requirements/*.txt` files. The patchset doesn't currently include that, but we could. One downside is we'd need to change the Dockerfile to point at the right `requirements/base.pyVER.txt` for 3.9, but we already have to modify that file.

I'm also happy to leave the PR as is. What do you think @rail ?


escapewindow:
Maybe we leave the PR as is, and fix minor version upgrades in taskgraph.

escapewindow:
I'm also not 100% sure about these patches, so we may want some staging runs. Or roll out at a quiet time when we can monitor and roll back.

rail (approved):
LGTM! Let's test it in staging! Also we can deploy the workers one by one in production.

bhearsum (approved):
I don't have strong feelings about always including the python version in the requirements files names - there's a good case to be made both ways. r+ either way :)

escapewindow:
Thanks!

PR #162, opened 2020-03-03
Bump balrogclient version

bhearsum:
This picks up https://github.com/mozilla-releng/balrog/pull/1197, which should help debugging in some scenarios.


MihaiTabara (approved):
Let's do this!

PR #163, opened 2020-03-04
Xpi shipit

escapewindow:


rail (commented):
XPI uses a slightly different [API end point](https://github.com/mozilla-releng/shipit/blob/bed967693e55cac02c8e4f6ddd5597ae7f1e6407/api/src/shipit_api/admin/api.yml#L492)... Shipitscript calls it from [here](https://github.com/mozilla-releng/scriptworker-scripts/blob/b40bda25e4f2eeaad9c62a10a926bd95ed2d63d7/shipitscript/src/shipitscript/shipitapi.py#L83). Also it explicitly tries to regenerate product details [later](https://github.com/mozilla-releng/scriptworker-scripts/blob/b40bda25e4f2eeaad9c62a10a926bd95ed2d63d7/shipitscript/src/shipitscript/shipitapi.py#L85).

I wonder if we should unify the API endpoints for simplicity (we may need to pass product name explicitly) and let shipit regenerate product details implicitly or have a special case for XPI in shipitscript.

What do you think?

escapewindow:
Hm. I don't think we should special case in shipitscript's code. We should either add config to shipitscript to specify the endpoint paths, and use it for all products, or we should have a single endpoint and pass product information in. The former is less ship-it work, but we should probably think about how we want ship-it to look going forward. Do we want to add a set of endpoints per product, or have a single set of endpoints for all products? @rail

rail:
I totally agree with that. I have no good solution to unify all end points for now, but it should be relatively easy to do for a this case. We can revisit the current approach at some point.

I'm going to change the API and submit a PR. No need to change anything here now.

rail (approved):

PR #164, opened 2020-03-05
Support ev cert sig algorithm

Callek:


Callek:
Huh, I thought I had landed all my outstanding patches, but I just found #43 I think I'll update that approach since it was what we previously talked about....

signingscript/src/signingscript/sign.py
```diff
@@ -1300,7 +1300,10 @@ def verify_mar_signature(cert_type, fmt, mar, keyid=None):
 
     infile = orig_path
     outfile = orig_path + "-new"
-    digest_algo = "sha1"
+    if "authenticode_ev" in fmt:
```
escapewindow, line 1303: This check may eventually be worth putting elsewhere, e.g.:

- `build_filelist_dict` could add this into the `filelist_dict`, for `**kwargs`, and
- rather than hardcode this check in the code, we could have a `script_config.yaml` setting, constant map, or similar, that specifies the `authenticode_digest_algo` map for formats. This would allow us to find the config to tweak without having to dig into code later on.

This doesn't block, but we may want to add a followup issue if we don't address.

escapewindow (approved):

signingscript/src/signingscript/sign.py
Callek: I'll create a new issue for this.

PR #167, opened 2020-03-06
scriptworker 32.1.0

escapewindow:


MihaiTabara (approved):
LGTM. Ah, `json-e` is bumping to 4.0.0, I wonder if we should expect any breakages  changes because of that.

escapewindow:
https://github.com/taskcluster/json-e/commit/8cf234d2f872516470bcba327a8f8b855ffb5f8a doesn't look too bad.

PR #172, opened 2020-03-09
Preserve tags when doing a merge.

srfraser:
I didn't flag this  option out as none of the existing calls seem to turn it off, as long as they use debugsetparents


.taskcluster.yml
```diff
@@ -150,8 +150,6 @@ tasks:
             # Build docker image and (optionally) push to docker hub
             'run_tests == "1" && dockerhub_repo != ""':
               taskId: '${as_slugid(project_name + "docker_build_and_push")}'
-              dependencies:
-                - '${as_slugid(project_name + python_version)}'
```
tomprince: I think this is left over from testing?

treescript/src/treescript/mercurial.py
```diff
@@ -163,6 +165,7 @@ def build_hg_environment():
     await run_hg_command(
         config, "robustcheckout", source_repo, repo_path, "--sharebase", share_base, "--upstream", upstream_repo, "--branch", branch, exception=CheckoutError
     )
+    await strip_outgoing(config, task, repo_path)
```
tomprince: I think this is unnecessary, and is left-over from before #178 enabled the `firefoxtree` extension.

treescript/docker.d/worker.yml
```diff
@@ -3,6 +3,7 @@ artifact_dir: { "$eval": "ARTIFACTS_DIR" }
 verbose: { "$eval": "VERBOSE == 'true'" }
 taskcluster_scope_prefix: { "$eval": "TASKCLUSTER_SCOPE_PREFIX" }
 treestatus_base_url: https://treestatus.mozilla-releng.net
+use_hgrc: false
```
tomprince: I think this (and the other hgrc related changes) are left-over from https://github.com/mozilla-releng/scriptworker-scripts/pull/172/commits/cb451a6f8003207c2fdba0ca704544e670e9972c which was my first iteration of #178 we should back them out in favor of the changes that landed.

treescript/src/treescript/merges.py
```diff
@@ -115,6 +115,28 @@ def touch_clobber_file(repo_path):
             "Merge old head via |hg debugsetparents {} {}| CLOSED TREE DONTBUILD a=release".format(tagged_from_rev, to_branch),
             repo_path=repo_path,
         )
+        # Preserve tags.
+        patch_file = os.path.join(os.path.dirname(repo_path), "patch_file")
+        tag_diff = await run_hg_command(config, "diff", "-r", to_branch, os.path.join(repo_path, ".hgtags"), "-U9", return_output=True, repo_path=repo_path)
+        with open(patch_file, "w") as fh:
+            fh.write(tag_diff)
+        await run_command(["patch", "-R", "-p1", "-i", patch_file], cwd=repo_path)
```
tomprince: This is equivalent to `hg revert -r <to_branch> .hg_tags`, which avoids the need for a temporary file

treescript/src/treescript/merges.py
```diff
@@ -115,6 +115,28 @@ def touch_clobber_file(repo_path):
             "Merge old head via |hg debugsetparents {} {}| CLOSED TREE DONTBUILD a=release".format(tagged_from_rev, to_branch),
             repo_path=repo_path,
         )
+        # Preserve tags.
+        patch_file = os.path.join(os.path.dirname(repo_path), "patch_file")
+        tag_diff = await run_hg_command(config, "diff", "-r", to_branch, os.path.join(repo_path, ".hgtags"), "-U9", return_output=True, repo_path=repo_path)
```
tomprince: If you are not apply the patch (see below), the `-U0` probably makes more sense. `-U` specifies the number of lines of context to include; since we are only looking at additions, we don't need to generate context.

treescript/src/treescript/merges.py
```diff
@@ -115,6 +115,28 @@ def touch_clobber_file(repo_path):
...
+            fh.write(tag_diff)
+        await run_command(["patch", "-R", "-p1", "-i", patch_file], cwd=repo_path)
+        os.unlink(patch_file)
+        with open(os.path.join(repo_path, ".hgtags"), "a") as fh:
+            # Skip four header lines
+            for line in tag_diff.splitlines()[4:]:
+                # We only care about additions.
+                if not line.startswith("+"):
+                    continue
+                line = line.replace("+", "")
```
tomprince: We should strip the first character, rather than replace, in case there are tags with `+` in them.

treescript/src/treescript/merges.py
```diff
@@ -115,6 +115,28 @@ def touch_clobber_file(repo_path):
             "Merge old head via |hg debugsetparents {} {}| CLOSED TREE DONTBUILD a=release".format(tagged_from_rev, to_branch),
             repo_path=repo_path,
         )
+        # Preserve tags.
```
tomprince: I think this logic for generating the new `.hgtags` likely deserves its own function.

I also agree with the comment in gecko_migrations about thinking there should be a better way than generating patches. I think we should preserve the sentiment from that comment.

tomprince (approved):

PR #174, opened 2020-03-12
fix adhoc signing

escapewindow:
The `docker.d/init.sh` patch fixes dev-signingscript. The widevine block isn't needed to fix, but we'll need it down the line, and I had that change locally, so I let it ride along. We can separate the two if wanted.


nthomas-mozilla (approved):

escapewindow:
Thanks!

PR #178, opened 2020-03-14
[treescript] Set mercurial config via files.

tomprince:
This make the logging clearer, because we are not passing all the config on
every call to hg. This also properly enable the firefoxtree extension, as we
don't set `HGRCPATH` (which was overriding the system-level config file in
production).

This also switches to using the robustcheckout from the version-control-tools
checkout we have in production.


treescript/src/treescript/mercurial.py
```diff
@@ -35,15 +34,7 @@ def build_hg_command(config, *args):
...
-    robustcheckout_args = [
-        "--config",
-        "extensions.robustcheckout={}".format(ROBUSTCHECKOUT_PATH),
-        "--config",
-        "extensions.purge=",
-        "--config",
-        "extensions.strip=",
-    ]
-    return hg + [*robustcheckout_args, *args]
+    return hg + [*args]
```
srfraser, line 37: nit: Can just do `hg + args`

srfraser (approved):
Looks like the CI errors are just `isort` so that's easily fixable. Wonder if we should fix that upstream. Will 'Approve' for now as I'm on PTO next week so may not respond quickly once the import order has changed.

Thank you for your help with this! I appreciate it a lot.

treescript/src/treescript/mercurial.py
tomprince: At least the tests pass a tuple for `args`, and you can't add a tuple to a list. The `[*args]` syntax turns the argument into a list.

tomprince:
> Looks like the CI errors are just `isort` so that's easily fixable. Wonder if we should fix that upstream. Will 'Approve' for now as I'm on PTO next week so may not respond quickly once the import order has changed.

It looks like I messed up the exclusion for the vendored file. I think fixing that should make the tests happy.

Issue #171, opened 2020-03-09
[meta] refactor scripts to stop importing scriptworker, start importing scriptworker_client


escapewindow:
Part of the rationale of having scriptworker call a script from the commandline, rather than import a python module and call a function, was to make the script independent of scriptworker. In theory this would:

1. let us upgrade scriptworker without affecting scripts, as long as that commandline api stayed consistent,
2. allow us to test scripts standalone, rather than requiring the worker,
3. allow for a mixture of languages and python versions

(3) is currently no longer an issue. We've standardized on python 3. However, we still have the option of supporting other languages, especially if some other team wants to start using scriptworkers.

(2) has been theoretical, and still has some issues with user friendliness and discoverability. [The `testing iscript` wiki page](https://github.com/mozilla-releng/scriptworker-scripts/wiki/Testing-iscript) shows it's possible to test a script without scriptworker, especially when using the [`create_test_workdir`](https://github.com/mozilla-releng/scriptworker-scripts/wiki/Testing-iscript#directory-tree) script from scriptworker to populate the `work_dir` with a known task's `task.json` and CoT upstreamArtifacts.

The main thing we're solving here is (1). Because many scripts import scriptworker, upgrading scriptworker may still break those scripts, even if the commandline api stays consistent. Let's import `scriptworker_client` instead.

Doing so isn't just a simple drop-in replacement, however. `scriptworker_client` removes the usage of `context` objects, so this may be a sizeable patch set per script. But once we do this, we'll once more be able to upgrade scriptworker without fear of breaking the script, and we will be able to make all script-related changes in this repo rather than across two repos. (And the vocal opponents to the `context` objects will be appeased. :)

escapewindow:
* [X] balrogscript
* [X] treescript
* [X] iscript
* [X] notarization_poller
* [ ] addonscript
* [ ] signingscript
* [ ] pushapkscript

The other \*scripts not listed could also import `scriptworker_client` for consistency, but don't currently import `scriptworker.client`.

PR #180, opened 2020-03-21
Switch to using async_main from scriptworker_client.

Callek:
This actually has the added benefit of *starting* down the path of async support, more consistent logging, and increasing code coverage for non-submitter/ code.


escapewindow:
Is there a reason why you decided to import `scriptworker.client` rather than `scriptworker_client` per https://github.com/mozilla-releng/scriptworker-scripts/issues/171 ? Just to avoid tearing out `context` usage?

Callek:
Switched to scriptworker_client now, I'm a bit sad to have to switch to not testing with pinned hashes though .. (I'm open to ideas on how to fix tox for that)

escapewindow (approved):
Thanks!

PR #183, opened 2020-03-25
Add "merge_day_clobber_file" to Treescript worker config.

jfx2006:
Needs a test case for when the clobber file is "".


treescript/src/treescript/merges.py
```diff
@@ -25,20 +25,23 @@ def replace(file_name, from_, to_):
...
+        new_contents = ""
+        for line in contents.splitlines():
+            line = line.strip()
+            if line.startswith("#") or line == "":
+                new_contents += f"{line}\n"
+        new_contents += "Merge day clobber"
+        with open(clobber_file, "w") as f:
+            f.write(new_contents)
+    else:
+        log.info("merge_day_clobber_file not set, skipping clobber file update")
```
srfraser: Minor preference: To me it looks better if we do `if not config[...` and return straight away as the `else` here ends up being quite far from the `if`. Not a blocking comment, though.

treescript/Dockerfile.test
```diff
@@ -7,7 +7,7 @@ WORKDIR /app
 COPY MANIFEST.in setup.py tox.ini /app/
 COPY requirements/ /app/requirements/
 
-RUN pip install -r requirements/local.txt
+RUN pip install -r requirements/local.py37.txt
```
srfraser: I believe releng is mostly on py38 now

srfraser (commented):
Looks good, thank you for submitting it! If we make sure it's up to date with master I can merge and deploy it.

treescript/Dockerfile.test
jfx2006: I don't recall why I had changed that. I've been working with py38.

treescript/src/treescript/merges.py
jfx2006: Changed.

srfraser (approved):

PR #187, opened 2020-03-26
Fix new use of async_main in balrogscript

### LINKED ISSUE #186: bug 1625053 - [balrogscript] post-async_main bustage
I landed https://github.com/mozilla-releng/scriptworker-scripts/pull/185 to fix, but we're still hitting

```
  File "/app/lib/python3.8/site-packages/balrogscript/script.py", line 187, in update_config
    config["api_root"] = config["server_config"][server]["api_root"]
TypeError: 'immutabledict' object does not support item assignment
```

Easy enough to fix: `config = dict(config)` before that line should do it. I think we should do a dev run before pushing to production again though, and potentially put in some `test_script` tests that don't mock quite as much.

Callek:
This fixes #186

### DIFF
--- a/balrogscript/src/balrogscript/script.py
+++ b/balrogscript/src/balrogscript/script.py
@@ -4,7 +4,6 @@ import json
 import logging
 import os
 import sys
-from copy import deepcopy
 
 from immutabledict import immutabledict
 from redo import retry  # noqa: E402
@@ -183,7 +182,7 @@ def set_readonly(task, config, auth0_secrets):
 
 # update_config {{{1
 def update_config(config, server="default"):
-    config = deepcopy(dict(config))
+    config = dict(config)
 
     config["api_root"] = config["server_config"][server]["api_root"]
     auth0_secrets = dict(
--- a/balrogscript/src/balrogscript/script.py
+++ b/balrogscript/src/balrogscript/script.py
@@ -6,6 +6,7 @@ import os
 import sys
 from copy import deepcopy
 
+from immutabledict import immutabledict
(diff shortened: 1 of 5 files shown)


balrogscript/src/balrogscript/script.py
```diff
@@ -182,7 +183,7 @@ def set_readonly(task, config, auth0_secrets):
 
 # update_config {{{1
 def update_config(config, server="default"):
-    config = deepcopy(config)
+    config = deepcopy(dict(config))
```
escapewindow: I don't think we need the deepcopy if we `dict()`.

balrogscript/tests/conftest.py
```diff
@@ -28,61 +29,73 @@ def release_manifest():
...
+            }
+        )
     finally:
         shutil.rmtree(tmpdir)
 
 
+@pytest.fixture(scope="function")
+def nightly_task(config):
+    with open(NIGHTLY_TASK_PATH, "r") as fh:
+        return json.load(fh)
```
escapewindow, line 81: Nice. Are we able to use this fixture to fix this todo? https://github.com/mozilla-releng/scriptworker-scripts/blob/55b7bbfe5a026b8ef352859a2dae9e0663ba1c75/balrogscript/tests/test_script.py#L379-L380

escapewindow (approved):
Thanks! Have we tested in dev, or are you confident in this?

balrogscript/tests/conftest.py
Callek: Looks like yes!

PR #188, opened 2020-03-26
Switch to using new behavior payload entry, support existing scopes

Callek:


balrogscript/src/balrogscript/script.py
```diff
@@ -232,7 +232,7 @@ def get_default_config():
 
 # main {{{1
 async def async_main(config, task):
-    action = get_task_action(task, config)
+    action = get_task_behavior(task, config)
```
escapewindow: Non-blocking nit: we could also rename this variable.

balrogscript/src/balrogscript/task.py
```diff
@@ -24,21 +24,24 @@ def validate_task_schema(script_config, task_definition, action):
...
-    actions = [s.split(":")[-1] for s in task["scopes"] if s.startswith(script_config["taskcluster_scope_prefix"] + "action:")]
-    if actions:
-        log.info("actions: %s", actions)
-        if len(actions) != 1:
-            raise ValueError("Only one action can be used")
-        action = actions[0]
+def get_task_behavior(task, script_config):
+    """Extract task behavior from payload, fallback on scopes"""
+    behavior = task.get("payload", {}).get("behavior")
+    behaviors = {s.split(":")[-1] for s in task["scopes"] if s.startswith(script_config["taskcluster_scope_prefix"] + "action:")}
```
escapewindow, line 31: I'm guessing we're deprecating using the action scope, and are moving towards using the payload. Is that accurate? If so, a comment may be in order.

escapewindow (approved):
Thanks!

PR #192, opened 2020-03-31
Bump central action for treescript / merge automation

srfraser:
Have reordered a bit of the merge action so that it can support the central bump and tag. bump_central doesn't fit too cleanly into the existing separate tag and bump actions as it determines the version & revision to use and the others need it specifying in the task. Maybe we want to rework that later so something outside the task knows what targets there are? 

BashFormatter is there because some of the replacement strings are bash variables, and so contain `{}` which shouldn't be formatted. Using the formatter here lets us keep the single `{variable}` strings in config, making them easier to grep, and also lets us avoid KeyError when formatting.


treescript/src/treescript/mercurial.py
```diff
@@ -110,7 +110,7 @@ def build_hg_environment():
         config (dict): the running config.
 
     """
-    await run_hg_command(config, "-v", "version", return_output=True)
```
JohanLorenzo, line 113: Question: Out of curiosity, what did `return_output` do in this context?

treescript/src/treescript/merges.py
```diff
@@ -44,22 +59,35 @@ def touch_clobber_file(repo_path):
...
         shutil.copyfile(os.path.join(repo_path, f[0]), os.path.join(repo_path, f[1]))
 
+    format_options = {
+        "current_major_version": current_major_version,
+        "next_major_version": version.major_number,
+        "current_weave_version": current_major_version + 2,
+        "next_weave_version": current_major_version + 3,  # current_weave_version + 1
+    }
+
+    fmt = BashFormatter()
```
JohanLorenzo, line 95: Question: What does this custom formatter achieve? I'm trying to understand what it does compared to https://docs.python.org/3/library/shlex.html#shlex.quote

treescript/tests/test_merges.py
```diff
@@ -188,3 +203,29 @@ def noop_replace(*arguments, **kwargs):
 
     assert len(called_args) == expected_calls
     assert result == expected_return
+
+
+@pytest.mark.asyncio
+async def test_bump_central(mocker, config, task, repo_context, merge_bump_info):
```
JohanLorenzo, line 209: 👍 for testing this case \o/

treescript/tests/test_merges.py
```diff
@@ -188,3 +203,29 @@ def noop_replace(*arguments, **kwargs):
...
+    async def noop_apply_rebranding(*arguments, **kwargs):
+        called_args.append("apply_rebranding")
+
+    mocker.patch.object(merges, "run_hg_command", new=mocked_run_hg_command)
+    mocker.patch.object(merges, "get_revision", new=mocked_get_revision)
+    mocker.patch.object(merges, "apply_rebranding", new=noop_apply_rebranding)
+
+    result = await merges.do_merge(config, task, repo_context.repo)
+
+    assert len(called_args) == 6
```
JohanLorenzo: Question: It's great to count the number of times `hg` was called. Do you think some calls are worth to be asserted? On top of my head, I think the tags and the commit ones are interesting and would be a good way to test their messages.

treescript/src/treescript/merges.py
```diff
@@ -44,22 +59,35 @@ def touch_clobber_file(repo_path):
...
     """Apply changes to repo required for merge/rebranding."""
     log.info("Rebranding %s to %s", merge_config.get("from_branch"), merge_config.get("to_branch"))
-    if merge_config.get("version_files"):
-        current_version = get_version("browser/config/version.txt", repo_path)
-        next_version = f"{current_version.major_number}.{current_version.minor_number}"
 
+    version = get_version("browser/config/version.txt", repo_path)
+    current_major_version = version.major_number
+    if merge_config.get("incr_major_version", False):
+        version = version.bump("major_number")
```
JohanLorenzo, line 71: I'm glad `mozilla-version` is useful in this case!

treescript/src/treescript/merges.py
```diff
@@ -44,22 +59,35 @@ def touch_clobber_file(repo_path):
...
+    if merge_config.get("version_files"):
+        next_version = f"{version.major_number}.{version.minor_number}"
         await do_bump_version(config, repo_path, merge_config["version_files"], next_version)
-    if merge_config.get("version_files_suffix"):
-        current_version = get_version("browser/config/version.txt", repo_path)
-        current_version = attr.evolve(current_version, is_esr=False, beta_number=None, is_nightly=False)
-        next_version = f"{current_version}{merge_config.get('version_suffix')}"
 
+    if merge_config.get("version_files_suffix"):
+        version = attr.evolve(version, is_esr=False, beta_number=None, is_nightly=False)
```
JohanLorenzo, line 78: TIL `attr.evolve()`

JohanLorenzo (approved):
These changes make sense to me! I have a couple of questions, but none of them are blocking this PR 🙂 

Like I said in the phabricator patch, I'm looking forward to trying this 👍

treescript/src/treescript/mercurial.py
srfraser: `run_hg_command` without that will print the results to stdout, so it ends up in the logs. With `return_output=True` it also returns it to the caller so it can be used - we were just throwing it away here

treescript/src/treescript/merges.py
srfraser: In order to deal with formatting like `{next_major_version}` we have to add string formatting, but some of the file placements contain bash variables, such  as this one: `"MOZ_REQUIRE_SIGNING=${MOZ_REQUIRE_SIGNING-0}"` 
The default `.format()` will throw a `KeyError` there. `shlex` is about making things shell-safe

So what we need is a formatter that will fill in variables it has been told about, such as `next_major_version` but not error if it hasn't been told about `MOZ_REQUIRE_SIGNING` because we want that passed through literally.

My first thought was to simply double `{{}}` the bash variables but that would mean that anyone looking for matching strings with grep would miss these - I can imagine a situation where we rename `MOZ_REQUIRE_SIGNING` and someone does `grep -r '${MOZ_REQUIRE_SIGNING'` to find other references to it, and the `{{` `}}` would mean the file replacement entry is missed. 

Since `.format` can't be told to ignore missing keys, we need a formatter that can be told what to do, and this does the job: If it's not passed a value for the `{key}` then it returns the key, letting the value be unchanged.

Phew that was long, but I hope it helps

treescript/src/treescript/merges.py
srfraser: It is, thank you!

treescript/src/treescript/merges.py
srfraser: Tom taught me that one.

treescript/tests/test_merges.py
srfraser: Yes, I think you're probably right, I will add a few more detailed checks

treescript/src/treescript/merges.py
JohanLorenzo: It helps a lot! Thank you very much for this thorough answer! 

Let's add a code comment to include this piece of knowledge alongside the class. This way, future readers will know right away 😃

PR #193, opened 2020-03-31
Bug 1634757 - V2 backend

Callek:
This PR for now only showcases v2-submit-toplevel and will need to be tested in a stage environment before utilized.

I also would love to see this refactored a bunch, though I suspect that refactor is not worthwhile until the unversioned backend api's are no longer used.


balrogscript/src/balrogscript/script.py
```diff
@@ -238,14 +239,19 @@ def get_default_config():
     server = get_task_server(task, config)
     auth0_secrets, config = update_config(config, server)
 
+    backend_version = 1
```
escapewindow, line 245: Are we planning on obsoleting backend version 1 if this goes well?
If/when we make that decision, let's add a comment.

balrogscript/src/balrogscript/submitter/cli.py
```diff
@@ -143,24 +148,50 @@ def generate_data(self, appVersion, productName, version, buildNumber, updateCha
...
+            try:
+                resp.raise_for_status()
+                return
+            except requests.HTTPError as excp:
+                logging.error("Caught HTTPError: %s", excp.response.content)
                 raise
+            finally:
+                stats = {"timestamp": time.time(), "method": "PUT", "url": url, "status_code": resp.status_code, "elapsed_secs": time.time() - before}
+                logging.debug("REQUEST STATS: %s", json.dumps(stats))
+            return
```
escapewindow: I'm half-wanting two helper functions, but that doesn't block. Especially if we're planning to tear out support for api version 1.

escapewindow (dismissed):

balrogscript/src/balrogscript/submitter/cli.py
```diff
@@ -143,24 +148,50 @@ def generate_data(self, appVersion, productName, version, buildNumber, updateCha
...
-                current_data, data_version = {}, None
-            else:
+
+        if self.backend_version == 2:
+            log.info("Using backend version 2...")
+            blob["schema_version"] = self.schemaVersion
+            blob["hashFunction"] = hashFunction
+            blob["name"] = name
+            data = {"product": productName, "blob": blob}
+            headers = {"Accept-Encoding": "application/json", "Accept": "application/json", "Content-Type": "application/json", "Referer": self.api_root}
```
bhearsum: Are you sure you need to set these? In my curl tests, only `Content-Type` is needed, and if you pass a dictionary to `requests_api.put`, it should set that for you, eg: `requests_api.put(url, json=data)`.

balrogscript/src/balrogscript/submitter/cli.py
bhearsum: I'd support this, if for no other reason than it makes it much easier to rip out v1 support. I'm happy to defer to @Callek on how he wants to proceed though.

bhearsum (commented):
This seems fine overall, just a couple of small comments.

balrogscript/src/balrogscript/submitter/cli.py
Callek: I wasn't sure these were needed, no.

Callek:
So I just updated the PR, I haven't yet tested this anywhere.

It looks like we'll need GET support to land to make this work, and this has not been tested yet at all.

I certainly want to improve this logic (and coverage) sooner than later, but I'm not blocking myself on it today. I don't mind if anyone wants to polish this more, or whatever. The Nightly and non-nightly paths are different here as well.

balrogscript/src/balrogscript/script.py
bhearsum: Yep, v1 will be dying.

bhearsum:
> It looks like we'll need GET support to land to make this work, and this has not been tested yet at all.

This got merged to master yesterday; I should have it stage sometime next week.

balrogscript/src/balrogscript/submitter/cli.py
```diff
@@ -261,6 +306,81 @@ def update_latest():
...
+        url = self.api_root + "/v2/" + name
+        existing_release = do_request(url)
+        existing_locale_data = existing_release["blob"].get(build_type, {}).get("locales", {}).get(locale)
+        update_data(url, existing_release, existing_locale_data)
+
+        name = get_nightly_blob_name(productName, branch, build_type, "latest", self.dummy)
+        url = self.api_root + "/v2/" + name
+        existing_release = do_request(url)
+        existing_locale_data = existing_release["blob"].get(build_type, {}).get("locales", {}).get(locale)
+        update_data(url, existing_release, existing_locale_data)
```
bhearsum: Minor nit, this could be a loop over (buildID, "latest") instead of repeating the entire block AFAICT.

balrogscript/src/balrogscript/submitter/cli.py
```diff
@@ -261,6 +306,81 @@ def update_latest():
...
+            skip_submission = bool(
+                existing_locale_data
+                and existing_locale_data.get("completes") == data.get("completes")
+                and all(p in existing_locale_data.get("partials", []) for p in data.get("partials", []))
+            )
+            if skip_submission:
+                log.warning("Dated data didn't change, skipping update")
+                return
+            # explicitly pass data version
+            blob = {"blob": {"platforms": {build_target: {"locales": {locale: data}}}}, "old_data_versions": {"platforms": {build_target: {"locales": {}}}}}
```
bhearsum: Nit: `blob` as the variable name is confusing when `blob` is also a top level key in the dict. I'd suggest calling the variable `data` instead.

balrogscript/src/balrogscript/submitter/cli.py
```diff
@@ -261,6 +306,81 @@ def update_latest():
...
+            if existing_release.get("old_data_versions", {}).get("platforms", {}).get(build_target, {}).get("locales", {}).get(locale):
+                blob["old_data_versions"]["platforms"][build_target]["locales"][locale] = existing_release["old_data_versions"]["platforms"][build_target][
+                    "locales"
+                ][locale]
+            do_request(url=url, method="post", data=json.dumps(blob))
```
bhearsum: You can pass this as `json` instead of dumping to `data`. This will also ensure proper headers are set.

balrogscript/src/balrogscript/submitter/cli.py
```diff
@@ -261,6 +306,81 @@ def update_latest():
...
+        headers = {"Accept-Encoding": "application/json", "Accept": "application/json", "Content-Type": "application/json", "Referer": self.api_root}
+
+        def do_request(url, method="get", data=None):
+            logging.debug("Balrog request to %s via %s", url, method.upper())
+            logging.debug("Data sent: %s", None)
+            before = time.time()
+            resp = requests_api.request(method=method, url=url, headers=headers, data=data)
+            try:
+                resp.raise_for_status()
+                return json.loads(resp.content)
```
bhearsum: You can use `resp.json()` instead of loading this by hand.

balrogscript/src/balrogscript/submitter/cli.py
```diff
@@ -261,6 +306,81 @@ def update_latest():
...
+        self, platform, buildID, productName, branch, appVersion, locale, hashFunction, extVersion, schemaVersion, isOSUpdate=None, **updateKwargs
+    ):
+        log.info("Using backend version 2...")
+        requests_api = get_balrog_api(auth0_secrets=self.auth0_secrets)
+        headers = {"Accept-Encoding": "application/json", "Accept": "application/json", "Content-Type": "application/json", "Referer": self.api_root}
+
+        def do_request(url, method="get", data=None):
+            logging.debug("Balrog request to %s via %s", url, method.upper())
+            logging.debug("Data sent: %s", None)
+            before = time.time()
```
bhearsum: You don't need to do this by hand. You can look at `resp.elapsed.total_seconds()` to get this.

balrogscript/src/balrogscript/submitter/cli.py
```diff
@@ -261,6 +306,81 @@ def update_latest():
...
 
+    def run_backend2(
+        self, platform, buildID, productName, branch, appVersion, locale, hashFunction, extVersion, schemaVersion, isOSUpdate=None, **updateKwargs
+    ):
+        log.info("Using backend version 2...")
+        requests_api = get_balrog_api(auth0_secrets=self.auth0_secrets)
+        headers = {"Accept-Encoding": "application/json", "Accept": "application/json", "Content-Type": "application/json", "Referer": self.api_root}
+
+        def do_request(url, method="get", data=None):
+            logging.debug("Balrog request to %s via %s", url, method.upper())
```
bhearsum: Rather than doing all of this manual logging ourselves I wonder if we should increase the `requests` or `urllib3` verbosity? Up to you if you want to try it.

balrogscript/src/balrogscript/submitter/cli.py
```diff
@@ -342,9 +463,36 @@ def run(self, platform, productName, appVersion, version, build_number, locale,
...
+            # XXX Check for existing data_version for this locale
+            blob = {
+                "blob": {"platforms": {build_target: {"locales": {locale: data}}}},
+                # XXX old_data_versions here is currently required but shouldn't be
+                "old_data_versions": {"platforms": {build_target: {"locales": {}}}},
+            }
+            headers = {"Accept-Encoding": "application/json", "Accept": "application/json", "Content-Type": "application/json", "Referer": self.api_root}
+            requests_api = get_balrog_api(auth0_secrets=self.auth0_secrets)
+            url = self.api_root + "/v2/" + name
+            logging.debug("Balrog request to {url} via POST", url)
```
bhearsum: This code appears to be nearly identical or identical as an above block; it should probably be refactored. Do you think it belongs in `balrogclient`? I'm open to that.

bhearsum (changes requested):
This looks fine overall, and like it will work with the new API. I have some minor suggestions and requests for improvements noted below.

balrogscript/src/balrogscript/submitter/cli.py
Callek: I mostly copied the manual logging from the upstream api we used in v1 for this. I didn't feel like playing with requests/urllib3 verbosity here at this point.

balrogscript/src/balrogscript/submitter/cli.py
Callek: Followup fodder? Since this is how the original v1 api does it?

balrogscript/src/balrogscript/submitter/cli.py
Callek: This does probably belong in balrogclient, I updated my apiv2 code to include that logging/timing fodder, which was culled from balrogclient from v1 code anyway.

balrogscript/src/balrogscript/submitter/cli.py
bhearsum: :man_shrugging: Your call

balrogscript/src/balrogscript/apiv2.py
```diff
@@ -0,0 +1,42 @@
+# This should live in balrogclient in the balrog repo, placed here to facilitate easy testing
```
bhearsum: Thanks for writing most of this! I'll tidy it up a bit and get it into `balrogclient`.

balrogscript/src/balrogscript/submitter/cli.py
```diff
@@ -261,6 +290,59 @@ def update_latest():
...
+        balrog_api = functools.partial(do_balrog_req, auth0_secrets=self.auth0_secrets)
+
+        targets = buildbot2updatePlatforms(platform)
+        build_target = targets[0]
+        alias = None
+        if len(targets) > 1:
+            alias = targets[1:]
+            log.debug("alias entry of %s ignored...", json.dumps(alias))
+        data = {"buildID": buildID, "appVersion": appVersion, "platformVersion": extVersion, "displayVersion": appVersion}
+        if isOSUpdate:
```
bhearsum: Please remove this; it was only ever used for B2G. It's a good time to kill the cruft.

balrogscript/src/balrogscript/submitter/cli.py
```diff
@@ -261,6 +290,59 @@ def update_latest():
...
+            data["isOSUpdate"] = isOSUpdate
+
+        data.update(self._get_update_data(productName, branch, **updateKwargs))
+
+        if "old-id" in platform:
```
bhearsum: And this. It was used for some old Android migration. It's dead code now.

balrogscript/src/balrogscript/submitter/cli.py
```diff
@@ -342,9 +425,21 @@ def run(self, platform, productName, appVersion, version, build_number, locale,
 
         data.update(self._get_update_data(productName, version, build_number, **updateKwargs))
 
-        api = SingleLocale(name=name, build_target=build_target, locale=locale, auth0_secrets=self.auth0_secrets, api_root=self.api_root)
-        current_data, data_version = api.get_data()
-        api.update_build(data_version=data_version, product=productName, hashFunction=hashFunction, buildData=json.dumps(data), schemaVersion=9)
+        if self.backend_version == 2:
+            log.info("Using backend version 2...")
+            # XXX Check for existing data_version for this locale
+            blob = {
```
bhearsum: Use `data` or something else that's not `blob` for this variable, please.

bhearsum (changes requested):
Just a few other nits, but this looks good. I like the new structure with the v2 and v1 code separated.

Callek:
@mozbhearsum can you please give the final changes a once over, this is from testing "make sure... VALID_ACTIONS" to the tip of this PR.

https://github.com/mozilla-releng/scriptworker-scripts/compare/722ecc8...8211a68

Callek:
I'll note for others, that testing failed a bit for nightly because there is no 'toplevel' submit used presently for nightly, and we rely on complicated backend logic that auto-creates the main blob for every locale submission, and logic in the balrogscript that always passes aliases in.

The reason this won't work for current v2 code is that a PUT will overwrite the entire blob, rather than amend it, and we'd need to PUT to create it in the first place. At the same time any locale/platform could finish and race with each other in nightlies, causing this PUT to happen incorrectly and race each other, ending up with a situation where we may be missing balrog entries for locales and have no visible way to say why.

Callek:
This is now ready for review @mozbhearsum   New stuff you haven't reviewed is in https://github.com/mozilla-releng/scriptworker-scripts/compare/4138fa4...0f2cde1

EDIT: does not yet fully support Nightly releases as we discussed.

bhearsum:
> This is now ready for review @mozbhearsum New stuff you haven't reviewed is in [4138fa4...0f2cde1](https://github.com/mozilla-releng/scriptworker-scripts/compare/4138fa4...0f2cde1)
> 
> EDIT: does not yet fully support Nightly releases as we discussed.

I didn't look in great depth (I'll do that when it's ready for final review, but everything in the diff you linked to looks sane to me.

bhearsum (approved):
I tried this out for both releases and nightlies on try -- it worked fine. I'd like to get this pushed to prod after we uplift central to beta. @escapewindow - do you want to review this, too?

escapewindow (approved):

PR #196, opened 2020-04-02
Bug 1626933 Add support for bump_esr action in treescript

srfraser:
Have refactored the payload so it can cope with the various different styles of version replacement, and this let us remove copy_files, as that only existed to cope with a version number update.

`merge_flavor` is now called `behavior` in the action, as ideally the action parameter should have been `merge-flavor` but this caused errors in taskcluster.

### DIFF
--- a/treescript/src/treescript/merges.py
+++ b/treescript/src/treescript/merges.py
@@ -72,17 +72,16 @@ def create_new_version(version_config, repo_path):
             {
                 "filename": mandatory path,
                 "new_suffix": string, default is to keep original.
-                "bump_major_number": bool, optional, default false
-                "bump_minor_number": bool, optional, default false
+                "version_bump": string, optional, enum 'major', 'minor'
             }
 
     Returns:
         string: new version string for file contents.
     """
     version = get_version(version_config["filename"], repo_path)
-    if version_config.get("bump_major_number"):
+    if version_config.get("version_bump") == "major":
         version = version.bump("major_number")
-    if version_config.get("bump_minor_number"):
+    elif version_config.get("version_bump") == "minor":
         version = version.bump("minor_number")
     if "new_suffix" in version_config:  # '' is a valid entry
         version = attr.evolve(version, is_esr=False, beta_number=None, is_nightly=False)
@@ -99,6 +98,13 @@ async def apply_rebranding(config, repo_path, merge_config):
 
     # Must collect this before any bumping.
     version = get_version("browser/config/version.txt", repo_path)
+    # Used in file replacements, further down.
+    format_options = {
+        "current_major_version": version.major_number,
+        "next_major_version": version.major_number + 1,
+        "current_weave_version": version.major_number + 2,
+        "next_weave_version": version.major_number + 3,  # current_weave_version + 1
+    }
 
     if merge_config.get("version_files"):
         for version_config in merge_config["version_files"]:
@@ -106,13 +112,6 @@ async def apply_rebranding(config, repo_path, merge_config):
 
     for f in merge_config.get("copy_files", list()):
         shutil.copyfile(os.path.join(repo_path, f[0]), os.path.join(repo_path, f[1]))
-
-    format_options = {
-        "current_major_version": version.major_number,
-        "next_major_version": version.major_number + 1,
-        "current_weave_version": version.major_number + 2,
-        "next_weave_version": version.major_number + 3,  # current_weave_version + 1
-    }
 
     # Cope with bash variables in strings that we don't want to
     # be formatted in Python. We do this by ignoring {vars} we
--- a/treescript/src/treescript/merges.py
+++ b/treescript/src/treescript/merges.py
@@ -64,32 +64,54 @@ def touch_clobber_file(config, repo_path):
             f.write(new_contents)
 
 
+def create_new_version(version_config, repo_path):
+    """Create the new version string used in file manipulation.
+
+    Arguments:
+        version_config (dict):
+            {
+                "filename": mandatory path,
+                "new_suffix": string, default is to keep original.
+                "bump_major_number": bool, optional, default false
+                "bump_minor_number": bool, optional, default false
+            }
+
+    Returns:
+        string: new version string for file contents.
+    """
+    version = get_version(version_config["filename"], repo_path)
+    if version_config.get("bump_major_number"):
+        version = version.bump("major_number")
+    if version_config.get("bump_minor_number"):
+        version = version.bump("minor_number")
+    if "new_suffix" in version_config:  # '' is a valid entry
+        version = attr.evolve(version, is_esr=False, beta_number=None, is_nightly=False)
+        version = f"{version}{version_config['new_suffix']}"
+    else:
+        version = f"{version}"
+    log.info("New version is %s", version)
+    return version
+
+
 async def apply_rebranding(config, repo_path, merge_config):
     """Apply changes to repo required for merge/rebranding."""
     log.info("Rebranding %s to %s", merge_config.get("from_branch"), merge_config.get("to_branch"))
 
+    # Must collect this before any bumping.
     version = get_version("browser/config/version.txt", repo_path)
-    current_major_version = version.major_number
-    if merge_config.get("incr_major_version", False):
-        version = version.bump("major_number")
 
     if merge_config.get("version_files"):
-        next_version = f"{version.major_number}.{version.minor_number}"
-        await do_bump_version(config, repo_path, merge_config["version_files"], next_version)
-
-    if merge_config.get("version_files_suffix"):
-        version = attr.evolve(version, is_esr=False, beta_number=None, is_nightly=False)
-        next_version = f"{version}{merge_config.get('version_suffix')}"
-        await do_bump_version(config, repo_path, merge_config["version_files_suffix"], next_version)
+        for version_config in merge_config["version_files"]:
+            await do_bump_version(config, repo_path, [version_config["filename"]], create_new_version(version_config, repo_path))
 
     for f in merge_config.get("copy_files", list()):
         shutil.copyfile(os.path.join(repo_path, f[0]), os.path.join(repo_path, f[1]))
 
     format_options = {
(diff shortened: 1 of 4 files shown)


treescript/src/treescript/merges.py
```diff
@@ -64,32 +64,54 @@ def touch_clobber_file(config, repo_path):
...
+
+    Returns:
+        string: new version string for file contents.
+    """
+    version = get_version(version_config["filename"], repo_path)
+    if version_config.get("bump_major_number"):
+        version = version.bump("major_number")
+    if version_config.get("bump_minor_number"):
+        version = version.bump("minor_number")
+    if "new_suffix" in version_config:  # '' is a valid entry
```
JohanLorenzo, line 86: Thanks for this useful comment 👍  I was actually wondering why we couldn't use `version_config.get()`

treescript/src/treescript/merges.py
```diff
@@ -64,32 +64,54 @@ def touch_clobber_file(config, repo_path):
...
+                "bump_minor_number": bool, optional, default false
+            }
+
+    Returns:
+        string: new version string for file contents.
+    """
+    version = get_version(version_config["filename"], repo_path)
+    if version_config.get("bump_major_number"):
+        version = version.bump("major_number")
+    if version_config.get("bump_minor_number"):
```
JohanLorenzo: Potential bug: I wonder if we should bail out when both `bump_major_number` and `bump_minor_number` are true. I hardly see a case where we do want to achieve this behavior. However, I can see cases where we don't want this behavior and we end up triggering it. For instance:

1. I duplicate the line `bump-minor-number: true` in `taskcluster/ci/config.yml`, without noticing it.
2. In a rush (because mergedays are important and time-constrained), I ask for a super quick review => The reviewer accepts it
3. We trigger the merge but end up with wrong version numbers.

What do you think?

treescript/src/treescript/merges.py
```diff
@@ -64,32 +64,54 @@ def touch_clobber_file(config, repo_path):
...
+        version = version.bump("minor_number")
+    if "new_suffix" in version_config:  # '' is a valid entry
+        version = attr.evolve(version, is_esr=False, beta_number=None, is_nightly=False)
+        version = f"{version}{version_config['new_suffix']}"
+    else:
+        version = f"{version}"
+    log.info("New version is %s", version)
```
JohanLorenzo, line 91: Good thinking on logging this here 👍

treescript/src/treescript/merges.py
```diff
@@ -64,32 +64,54 @@ def touch_clobber_file(config, repo_path):
...
+    version = get_version(version_config["filename"], repo_path)
+    if version_config.get("bump_major_number"):
+        version = version.bump("major_number")
+    if version_config.get("bump_minor_number"):
+        version = version.bump("minor_number")
+    if "new_suffix" in version_config:  # '' is a valid entry
+        version = attr.evolve(version, is_esr=False, beta_number=None, is_nightly=False)
+        version = f"{version}{version_config['new_suffix']}"
+    else:
+        version = f"{version}"
```
JohanLorenzo, line 90: Question: I'm sorry, I'm not sure to follow. What does converting a `mozilla-version` into a string achieve here?

treescript/src/treescript/versionmanip.py
```diff
@@ -23,6 +23,8 @@
     "mobile/android/config/version-files/beta/version_display.txt",
     "mobile/android/config/version-files/release/version.txt",
     "mobile/android/config/version-files/release/version_display.txt",
+    "mobile/android/config/version-files/nightly/version.txt",
```
JohanLorenzo, line 26: 😱 I'm surprised these files weren't there in the first place.

Thank you for adding them!

treescript/src/treescript/merges.py
```diff
@@ -64,32 +64,54 @@ def touch_clobber_file(config, repo_path):
...
-        version = attr.evolve(version, is_esr=False, beta_number=None, is_nightly=False)
-        next_version = f"{version}{merge_config.get('version_suffix')}"
-        await do_bump_version(config, repo_path, merge_config["version_files_suffix"], next_version)
+        for version_config in merge_config["version_files"]:
+            await do_bump_version(config, repo_path, [version_config["filename"]], create_new_version(version_config, repo_path))
 
     for f in merge_config.get("copy_files", list()):
         shutil.copyfile(os.path.join(repo_path, f[0]), os.path.join(repo_path, f[1]))
 
     format_options = {
```
JohanLorenzo: Nit: I think we should now move this line right after https://github.com/mozilla-releng/scriptworker-scripts/pull/196/files#diff-335c841df8c4f66756b08ed519cd72ddR101. `version` is not used anywhere else and I was confused to see it assigned but not used anymore in the scope of `if merge_config.get("version_files")`

treescript/src/treescript/merges.py
```diff
@@ -64,32 +64,54 @@ def touch_clobber_file(config, repo_path):
...
     if merge_config.get("version_files"):
-        next_version = f"{version.major_number}.{version.minor_number}"
-        await do_bump_version(config, repo_path, merge_config["version_files"], next_version)
-
-    if merge_config.get("version_files_suffix"):
-        version = attr.evolve(version, is_esr=False, beta_number=None, is_nightly=False)
-        next_version = f"{version}{merge_config.get('version_suffix')}"
-        await do_bump_version(config, repo_path, merge_config["version_files_suffix"], next_version)
+        for version_config in merge_config["version_files"]:
+            await do_bump_version(config, repo_path, [version_config["filename"]], create_new_version(version_config, repo_path))
```
JohanLorenzo, line 111: To be honest, I love this new logic much more than what the previous one 👍 

Thank you for making it simpler!

treescript/tests/test_merges.py
```diff
@@ -178,6 +182,37 @@ def noop_replace(*arguments, **kwargs):
     assert called_args[0] == expected
 
 
+@pytest.mark.asyncio
```
JohanLorenzo, line 185: Yay for more tests! 🎉 😃

treescript/tests/test_merges.py
```diff
@@ -178,6 +182,37 @@ def noop_replace(*arguments, **kwargs):
...
+@pytest.mark.parametrize(
+    "version_config,current_version, expected",
+    (
+        # central-to-beta
+        ({"filename": "config/milestone.txt", "new_suffix": ""}, FirefoxVersion.parse("75.0a1"), "75.0"),
+        ({"filename": "browser/config/version_display.txt", "new_suffix": "b1"}, FirefoxVersion.parse("75.0a1"), "75.0b1"),
```
JohanLorenzo, line 191: I like the fact that this suite is exhaustive! The number of files is small and this test doesn't seem too expensive to run, so I agree we can be as exhaustive as possible.

Nit: I found an additional case:

```suggestion
        ({"filename": "browser/config/version.txt", "new_suffix": ""}, FirefoxVersion.parse("75.0a1"), "75.0"),
        ({"filename": "browser/config/version_display.txt", "new_suffix": "b1"}, FirefoxVersion.parse("75.0a1"), "75.0b1"),
```

JohanLorenzo (approved):
\o/ 

Like I said in https://phabricator.services.mozilla.com/D69425#2111485, this is wonderful work!

I found what I think is a potential bug that might bite us in the future. I don't think it blocks this PR from being merge as is. My comments can be done in followups.

treescript/src/treescript/merges.py
srfraser: I think this is a good layer of protection to have.

Could throw an error, or for now just do: 

    -    if version_config.get("bump_minor_number"):
    +    elif version_config.get("bump_minor_number"):

treescript/src/treescript/merges.py
srfraser: The existing bump_version function takes strings, and combined with the new suffix change it seemed easier this way. I can change the wrapper's call at https://github.com/mozilla-releng/scriptworker-scripts/blob/master/treescript/src/treescript/versionmanip.py#L87 as well as `do_bump_version`

`version = f"{version}{version_config['new_suffix']}"` though - do you think it'd be ok to generate the new version that way and just `.parse` it? I was trying to come up with a way of replacing the suffix without having to know about the various options. What do you think?

treescript/src/treescript/merges.py
srfraser: Good call

treescript/tests/test_merges.py
srfraser: I think that's the same replacement operation as `({"filename": "config/milestone.txt", "new_suffix": ""}, FirefoxVersion.parse("75.0a1"), "75.0"),` isn't it? Since the test mocks out reading the actual file, which doesn't exist.

treescript/src/treescript/merges.py
srfraser: I like it more as well

treescript/src/treescript/merges.py
JohanLorenzo: I'd prefer the worker to raise an error so we don't end up with a false-green task. It can be done in a followup with the `elif` logic landed now 🙂

treescript/src/treescript/merges.py
JohanLorenzo: Discussed over Slack. 

Good call `f"{version}{version_config['new_suffix']}"` may end up in a broken version number and `mozilla-version` can help in capturing this error. We could implement a function similar to `bump` that takes care of "promoting" a nightly version into a beta an so on. 

That said, making this change is not doable before Monday, so let's keep this string conversion as is, and iterate later on.

treescript/tests/test_merges.py
JohanLorenzo: True. I was thinking on making this case explicit so the expected behavior is documented by the test. I agree with you, from a functional perspective, it doesn't test anything more than the previous case.

treescript/src/treescript/merges.py
srfraser: New commit incoming where it's not possible to set both

PR #198, opened 2020-04-08
log notarization response

escapewindow:
I'm wondering if Apple is sending us information about the lost UUIDs in the response, and we're dropping it. Let's log the response so we can debug when we see this again.

I don't *think* we're receiving anything sensitive in the response...

### DIFF
--- a/iscript/src/iscript/mac.py
+++ b/iscript/src/iscript/mac.py
@@ -595,10 +595,12 @@ def get_uuid_from_log(log_path):
     regex = re.compile(r"RequestUUID = (?P<uuid>[a-zA-Z0-9-]+)")
     try:
         with open(log_path, "r") as fh:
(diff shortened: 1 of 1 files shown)


nthomas-mozilla (approved):
Agree this is a good idea. The other place that might help is to add `--verbose` to the altool invocation for upload - happy to wait and see if that's necessary. I know it automatically switches to verbose when the service is down but it might be missing an error case.

nthomas-mozilla:
Whoops, `--verbose` is only available in altool 4.0 and higher, and we have 1.1.1138.

escapewindow:
Thanks! I think I'd like to see what `--verbose` does before adding it by default -- by the blog post you linked, it looks like it's super verbose.

escapewindow:
Plus upgrading XCode first :)

PR #199, opened 2020-04-10
poller - catch TimeoutError during claimWork

escapewindow:
I was thinking about landing this with the rest of the lost UUID changes. Now that we know the root cause of at least some of the lost UUIDs, we may not need those changes, so let's land this earlier.


nthomas-mozilla (approved):
Ah whoops, I missed that https://github.com/mozilla-releng/scriptworker/pull/451 didn't touch the right code. Still good to have that fix too.

escapewindow:
It's one PR for scriptworker, one PR for notarization poller... we need both.

PR #200, opened 2020-04-10
bug 1620697 - Raise exception if `ERROR` is in the response; also notarize pkg in single_zip

escapewindow:
I debated putting in notarization retries around this error, but I
haven't, yet, because:

 - I'm leaning towards moving to `single_zip` notarization rather than
   `multi_account`. Overall `single_zip` is a simpler setup, simpler
   workflow, and should automatically reduce our concurrent notarization
   requests to less than the 20/min max Apple is enforcing. It's
   possible that `single_zip` will resolve bug 1620697 by itself.
   Retrying around `multi_account` will keep the complexity, and hammer
   Apple with even more requests.

- I want these errors to be very visible for a period of time. I believe
  that `single_zip` will resolve the upload limit issue, but it's
  possible they could respond with other `ERROR` messages that we would
  automatically retry without noticing. This patch throws separate
  errors around `ERROR ITMS-10004` (`ThrottledNotarization`) and an
  unknown `ERROR` (`UnknownNotarizationError`) which should let us
  decide how to behave in each case.

https://bugzilla.mozilla.org/show_bug.cgi?id=1620697#c18 has an example
response we need to detect.

### DIFF
--- a/iscript/tests/test_mac.py
+++ b/iscript/tests/test_mac.py
@@ -13,7 +13,7 @@ import pexpect
 import pytest
 
 import iscript.mac as mac
-from iscript.exceptions import InvalidNotarization, IScriptError, TimeoutError, UnknownAppDir
+from iscript.exceptions import InvalidNotarization, IScriptError, ThrottledNotarization, TimeoutError, UnknownAppDir, UnknownNotarizationError
 from scriptworker_client.utils import makedirs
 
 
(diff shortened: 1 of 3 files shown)


nthomas-mozilla (approved):
:+1:

iscript/tests/test_mac.py
```diff
@@ -439,25 +439,34 @@ def test_get_bundle_id(mocker, counter):
...
+def test_get_uuid_from_log(tmpdir, uuid, raises, extra):
     """``get_uuid_from_log`` returns the correct uuid from the logfile if present.
     It raises if it has problems finding the uuid in the log.
 
     """
     log_path = os.path.join(str(tmpdir), "log")
     if raises != "missing file":
         with open(log_path, "w") as fh:
-            fh.write("foo\nbar\nbaz\n RequestUUID = {} \nblah\n".format(uuid))
+            fh.write(f"foo\nbar\nbaz\n RequestUUID = {uuid} \n{extra}\nblah\n")
```
nthomas-mozilla, line 464: Nit, this could be closer to reality if the output written depended on the exception we expect to raise (aka no RequestUUID would be present when throttled).

escapewindow: Hm. I think we do get a UUID when throttled, which is part of the problem. (If we didn't get a RequestUUID when throttled, an unpatched `get_uuid_from_log` wouldn't find a UUID, and we would raise an `IScriptError` during notarization-part-1. For some reason they are sending a RequestUUID and `altool` exits 0. The only sign that there was an issue is the `ERROR ITMS-10004`. This is why we're only realizing that the UUID was throttled in the poller, many hours later.)

escapewindow:
Thanks! I'm going to roll this out to the pool as-is. If I'm mistaken about https://github.com/mozilla-releng/scriptworker-scripts/pull/200#discussion_r407581683 let's revisit :)

iscript/tests/test_mac.py
escapewindow: Hm. We did log output with the `ERROR ITMS-10004` but no `RequestUUID`, though, didn't we? Now I'm confused where we get the UUID from.

escapewindow: Aha! `2020-04-10 02:19:54,437 - scriptworker_client.utils - INFO - [2020-04-10 02:19:53 GMT] <main> DBG-X:   parameter RequestUUID = 60707656-fd64-430b-ae00-1c4c8fc69463` From https://firefoxci.taskcluster-artifacts.net/VDm1wKy3TK-6VkzvNsqa3Q/0/public/logs/live_backing.log 

My original thought was right: we're getting both `RequestUUID` and `ERROR ITMS-10004` in the same response.

nthomas-mozilla: Ah, that makes sense now. Lets disregard my comment.

PR #202, opened 2020-04-23
Implement pypiscript.

tomprince:


jcristau:
Going to close this.  We've moved most/all releng projects that are published to pypi over to using https://docs.pypi.org/trusted-publishers/ with github actions instead of manual uploads, so I don't think we'll need/want something like this in the foreseeable future.

PR #205, opened 2020-04-30
Unify Dockerfiles, do not upgrade pip

rail:
Upgrading pip causes some permission issues on new base images


MihaiTabara (approved):
Thanks a lot for fixing these @rail!

PR #211, opened 2020-05-10
Implement a PyPI script in rust.

tomprince:


escapewindow:
Some early thoughts:

1. We should block landing on an RFC where the team agrees we want to make Rust a Releng-team-supported language. We should also start conversations as to how this affects our [best practices](https://github.com/mozilla-releng/releng-rfcs/blob/master/rfcs/0022-app-best-practices.md), e.g. how we pin and update dependencies for non-python projects.
2. Thanks for adding tests!
3. I'm not sure we want a top-level module named `script`, as that could be confused with scripts that we run for maintenance or python. Not sure if we should rename, move everything into `python/` and `rust/` subdirectories, or implement some other solution.

tomprince:
>     1. We should block landing on an RFC where the team agrees we want to make Rust a Releng-team-supported language. We should also start conversations as to how this affects our [best practices](https://github.com/mozilla-releng/releng-rfcs/blob/master/rfcs/0022-app-best-practices.md), e.g. how we pin and update dependencies for non-python projects.

Per https://trello.com/c/wd6VJByU/192-languages#comment-5e825b272cb0146c889d397f my understanding was that this could land (being an experiment and not critical path) but that further usage would require a broader discussion.

>     2. Thanks for adding tests!

Note that I setup linting and running tests, but it is currently running 0 tests.
 
>     3. I'm not sure we want a top-level module named `script`, as that could be confused with scripts that we run for maintenance or python. Not sure if we should rename, move everything into `python/` and `rust/` subdirectories, or implement some other solution.

I don't have strong feelings on this.

tomprince:
@mozilla-releng/releng I think this is ready for people to look at. Please ask about anything that isn't clear.

escapewindow (changes requested):
> > ```
> > 1. We should block landing on an RFC where the team agrees we want to make Rust a Releng-team-supported language. We should also start conversations as to how this affects our [best practices](https://github.com/mozilla-releng/releng-rfcs/blob/master/rfcs/0022-app-best-practices.md), e.g. how we pin and update dependencies for non-python projects.
> > ```
> 
> Per https://trello.com/c/wd6VJByU/192-languages#comment-5e825b272cb0146c889d397f my understanding was that this could land (being an experiment and not critical path) but that further usage would require a broader discussion.

I didn't gather that from https://trello.com/c/wd6VJByU/192-languages#comment-5e825ecf04330e0419889148 , nor do I know what the discussion resolution was in Europe. I think we should block on this.

I still haven't done a deep dive here; I'll do that after this step.

> 
> > ```
> > 2. Thanks for adding tests!
> > ```
> 
> Note that I setup linting and running tests, but it is currently running 0 tests.

Let's add some tests. at least enough to cover basic functionality.

> > ```
> > 3. I'm not sure we want a top-level module named `script`, as that could be confused with scripts that we run for maintenance or python. Not sure if we should rename, move everything into `python/` and `rust/` subdirectories, or implement some other solution.
> > ```
> 
> I don't have strong feelings on this.

PR #215, opened 2020-05-15
Bug 1635488 - add initial schema adjustments.

MihaiTabara:


treescript/setup.py
```diff
@@ -7,7 +7,7 @@
     version = f.read().rstrip()
 
 with open(os.path.join(os.path.abspath(os.path.dirname(__file__)), "requirements", "base.in")) as f:
-    install_requires = ["scriptworker_client"] + f.readlines()
+    install_requires = f.readlines()
```
escapewindow, line 10: I think this will break all the `from scriptworker_client.* import` lines.

MihaiTabara: It will, I was playing locally with some things and accidentally pushed it to PR. Will clean-it up before finalizing this PR. Thanks for raising! 👍

PR #219, opened 2020-05-22
Fix matching signing formats

catlee:


signingscript/src/signingscript/task.py
```diff
@@ -154,14 +153,16 @@ async def sign(context, path, signing_formats, **kwargs):
...
+def _get_signing_function_from_format(fmt):
     try:
-        _, signing_function = get_single_item_from_sequence(FORMAT_TO_SIGNING_FUNCTION.items(), condition=lambda item: re.match(item[0], format) is not None)
+        _, signing_function = get_single_item_from_sequence(FORMAT_TO_SIGNING_FUNCTION.items(), condition=lambda item: item[0] == fmt.split(":")[0])
         return signing_function
     except ValueError:
         # Regex may catch several candidate. If so, we fall back to the exact match.
         # If nothing matches, then we fall back to default
-        return FORMAT_TO_SIGNING_FUNCTION.get(format, FORMAT_TO_SIGNING_FUNCTION["default"])
+        func = FORMAT_TO_SIGNING_FUNCTION.get(fmt, FORMAT_TO_SIGNING_FUNCTION["default"])
```
escapewindow: Nit: We can probably replace this whole function with

`signing_function = FORMAT_TO_SIGNING_FUNCTION.get(fmt.split(':')[0], FORMAT_TO_SIGNING_FUNCTION["default"])`

I'm ok landing as-is if we're trying to deal with bustage sooner. We may want to follow up, if so.

escapewindow (approved):

signingscript/src/signingscript/task.py
escapewindow: (Also open to the function staying if we want to keep this warning. We probably don't need `get_single_item_from_sequence` anymore, though.)

PR #226, opened 2020-06-05
bug 1597598 - use taskgraph

### LINKED ISSUE #140: bug 1597598 - [scriptworker-scripts] use taskgraph
We may want to first port the existing logic to taskgraph, and then convert to cloudops downloading taskcluster artifacts rather than relying on dockerhub.

* [ ] .taskcluster.yml change to be taskgraph-ified
* [ ] decision task
* [ ] docker image task(s) - python37, python38 we can likely use directly, but we could add pushapk/pushsnap deps, f/e.
* [ ] cached tasks to avoid rebuilding/retesting all scripts if only one script changed
* [ ] relpro to promote a docker image to cloudops?
* [ ] optional leaf tasks that don't block promotion
* [ ] make sure existing checks still work. coverage, ben's standards check, `tox -e py3{7,8}`, `tox -e check`
* [ ] continue using secrets to publish docker images to dockerhub (relpro?) until we switch away. but also publish as a taskcluster artifact

escapewindow:
Fixes #140 .

This ports the existing tox and dockerhub tasks running on scriptworker-scripts to taskgraph, with the following changes:

- we create docker-image tasks to avoid running setup every time we run a tox or dockerhub task;
- we use `cached_tasks` so we only rerun tox and dockerhub tasks if something in their `digest-directories` change;
  - however, on `dev-*` and `production-*` pushes, we force-rerun the appropriate dockerhub task
- we enable `chain-of-trust` for future verification

We're still using secrets to publish docker images to dockerhub, using the old logic. After this lands, we can follow up by having cloudops download an artifact from the task, with CoT verification. Also, we can modernize our k8s-image task with either gecko-type docker-image logic, or Tom's new non-dind hotness.

We can also follow up with:
- relpro to promote a docker image to cloudops
- other tasks -- build wheels?


escapewindow:
Wip is in https://github.com/escapewindow/staging-scriptworker-scripts/commits/master . I figured rebasing all into one commit was cleaner. I can split this up more if preferred.

escapewindow:
@tomprince , @Callek are you up for review?

escapewindow:
Ah, I may need to add a `pr` kind.

escapewindow:
If you look at https://firefoxci.taskcluster-artifacts.net/UmJBeMDlQuGOk1q9f35kEQ/0/public/label-to-taskid.json , you see all the tasks that *would* have been scheduled, but are using `cached_task`.

escapewindow:
Hm. Not sure why the PR status isn't being reported. This may be due to github repo settings that assume we're using non-taskgraph checks?

escapewindow:
The error above seems to be regarding the merge to my fork; my fork doesn't have the right scopes. Most likely I need to either add scopes to my fork, or turn off the tc integration there.

Edit: did the latter.

.taskcluster.yml
```diff
@@ -1,210 +1,254 @@
-# This Source Code Form is subject to the terms of the Mozilla Public
-# License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
 version: 1
+reporting: checks-v1
 policy:
-  pullRequests: public
+    pullRequests: collaborators
```
MihaiTabara, line 5: 👍

taskcluster/ci/config.yml
```diff
@@ -0,0 +1,48 @@
...
+
+taskgraph:
+    register: scriptworker_taskgraph:register
+    repositories:
+        scriptworker:
+            name: "Scriptworker-scripts"
+            project-regex: scriptworker-scripts$
+            default-repository: https://github.com/mozilla-releng/scriptworker-scripts
+            default-ref: master
+            type: git
```
MihaiTabara, line 15: Huh, TIL, I should backport this in rest of repos.

taskcluster/ci/config.yml
```diff
@@ -0,0 +1,48 @@
...
+
+workers:
+    aliases:
+        b-linux:
+            provisioner: scriptworker-{level}
```
MihaiTabara, line 20: Workers are up and running looking at https://hg.mozilla.org/ci/ci-configuration/file/tip/worker-pools.yml#l2379 so this looks good 👍

taskcluster/ci/config.yml
```diff
@@ -0,0 +1,48 @@
...
+            worker-type: 'images'
+
+docker:
+    email: "release+dockerhub+services@mozilla.com"
+    user: mozillarelengservices
+
+notify:
+    email:
+        # TODO by-project/level
+        - aki+scriptworker-scripts@mozilla.com
```
MihaiTabara: Nit: do we want to update this now or in a follow-up, once we stabilize the CI via taskgraph?

taskcluster/ci/docker-image/kind.yml
```diff
@@ -0,0 +1,41 @@
...
+        definition: k8s-image
+    python38:
+        definition: python
+        args:
+            PYTHON_VERSION: "3.8"
+    python37:
+        definition: python
+        args:
+            PYTHON_VERSION: "3.7"
+    pushapkscript-py38:
```
MihaiTabara: ~Hm, question: how come there's only three scriptworkers here?~
Later edit: don't mind me, just realized by looking at the docker images later on that they all three have externals deps that are different than the base image that we build for the others ones.

taskcluster/scriptworker_taskgraph/transforms/cached.py
```diff
@@ -0,0 +1,95 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
MihaiTabara, line 1: Hm, I didn't stumble upon this type of transform in neither of the mobile projects nor in-tree gecko. Maybe worth adding a short comment on it?

tox.ini
```diff
@@ -12,6 +8,10 @@ envlist =
     beetmoverscript-py38
     bouncerscript-py37
     bouncerscript-py38
```
MihaiTabara, line 10: Thanks for sorting them 👍

MihaiTabara:
> The error above seems to be regarding the merge to my fork; my fork doesn't have the right scopes. Most likely I need to either add scopes to my fork, or turn off the tc integration there.
> 
> Edit: did the latter.

@escapewindow I think so. If I remember correctly, I've hit this too. That's because FirefoxCI TC-github listens to push events on the repo. We either grant both your fork and the main repo the same level of scopes or turn off the integration on the fork to not confuse the listener.

taskcluster/ci/pr/kind.yml
```diff
@@ -0,0 +1,18 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+loader: taskgraph.loader.transform:loader
+
+kind-dependencies:
+    - k8s-image
```
tomprince, line 9: I think this should depend on `tox` as well?

taskcluster/scriptworker_taskgraph/transforms/cached.py
```diff
@@ -0,0 +1,95 @@
...
+            # files.update(list_files(os.path.join(BASE_DIR, "taskcluster")))
+            for path in ADDITIONAL_FILES:
+                if os.path.exists(path):
+                    files.update({path})
+            for path in sorted(list(files)):
+                h.update(
+                    "{} {}\n".format(
+                        hash_path(os.path.realpath(os.path.join(BASE_DIR, path))), path
+                    )
+                )
```
tomprince: You can use `taskgraph.util.hash.hash_paths` for this.

taskcluster/scriptworker_taskgraph/transforms/cached.py
```diff
@@ -0,0 +1,95 @@
...
+            task.setdefault("attributes", {}).setdefault("cached_task", {})
```
tomprince: I'm not sure why this is necessary?

taskcluster/scriptworker_taskgraph/transforms/cached.py
```diff
@@ -0,0 +1,95 @@
...
+    repo_name = repo_name.replace(".git", "").rstrip("/")
+    repo_name = repo_name.split("/")[-1]
+
+    for task in tasks:
+        if task.get("cache", True) and not taskgraph.fast:
+            digest_data = []
+            h = hashlib.sha256()
+            h.update(
+                json.dumps(task.get("attributes", {}).get("digest-extra", {}), indent=2, sort_keys=True)
+            )
```
tomprince: You don't need to pre-hash the data passed to `digest_data`; it takes a list of strings and generates a digest from them.

taskcluster/scriptworker_taskgraph/transforms/python_version.py
```diff
@@ -0,0 +1,60 @@
...
+
+from taskgraph.transforms.base import TransformSequence
+
+
+transforms = TransformSequence()
+
+def _replace_string(obj, repl_dict):
+    if isinstance(obj, dict):
+        for k in obj.keys():
+            obj[k] = obj[k].format(**repl_dict)
```
tomprince:
```suggestion
        return {k: v.format(**subs) for k, v in obj.items()}
```
and similar?

taskcluster/scriptworker_taskgraph/transforms/python_version.py
```diff
@@ -0,0 +1,60 @@
...
+def skip_on_project_specific_branches(config, jobs):
+    """Skip if the branch is project-specific for a different project."""
+    project_specific_prefixes = ("refs/heads/dev-", "refs/heads/production-")
+    for job in jobs:
+        script_name = job["attributes"]["script-name"]
+        project_specific_branches = ["{}{}".format(prefix, script_name) for prefix in project_specific_prefixes]
+        if config.params['head_ref'].startswith(project_specific_prefixes) and \
+                config.params['head_ref'] not in project_specific_branches:
+            continue
+        yield job
```
tomprince: My inclination would be to handle this via setting `target_tasks_method` and script-name parameter based on the branch, I think.

taskcluster/scriptworker_taskgraph/transforms/python_version.py
```diff
@@ -0,0 +1,60 @@
...
+            task['worker']['docker-image'] = _replace_string(task['worker']['docker-image'], repl_dict)
+            task['description'] = _replace_string(task['description'], repl_dict)
+            if task.get('run', {}).get('command'):
+                task['run']['command'] = _replace_string(task['run']['command'], repl_dict)
+            if task['worker'].get('command'):
+                task['worker']['command'] = _replace_string(task['worker']['command'], repl_dict)
+            if task.get('docker-repo'):
+                task['docker-repo'] = _replace_string(task['docker-repo'], repl_dict)
+            task.setdefault("attributes", {}).update({
+                "script-name": job["name"],
```
tomprince: I am *very* slightly inclined to move this to its own transform function before this one.

taskcluster/ci/tox/kind.yml
```diff
@@ -0,0 +1,102 @@
...
+        cwd: '{checkout}'
+        command:
+            - sh
+            - -lxce
+            - >-
+              tox -e {name}-py{python_version}
+
+jobs:
+    addonscript:
+        digest-directories:
```
tomprince: I would be inclined to use `resources` for this, to match what toolchains use.

taskcluster/ci/tox/kind.yml
```diff
@@ -0,0 +1,102 @@
...
+    attributes:
+        code-review: true
+    worker-type: b-linux
+    worker:
+        docker-image: {in-tree: 'python{python_version}'}
+        max-run-time: 1800
+        artifacts:
+            - type: directory
+              name: public/
+              path: /builds/worker/artifacts/
```
tomprince: Do we generate anything here?

taskcluster/docker/pushapkscript/Dockerfile
```diff
@@ -0,0 +1,7 @@
+# %ARG DOCKER_IMAGE_PARENT
+FROM $DOCKER_IMAGE_PARENT
+
+VOLUME /builds/worker/checkouts
+VOLUME /builds/worker/.cache
+
+RUN apt-get update && apt-get install -y default-jdk
```
tomprince: It might be worth considering pulling out these commands into a script that can be shared with images we publish, so they don't get out-of-sync. Not a blocker.

taskcluster/docker/pushapkscript/Dockerfile
```diff
@@ -0,0 +1,7 @@
+# %ARG DOCKER_IMAGE_PARENT
+FROM $DOCKER_IMAGE_PARENT
+
+VOLUME /builds/worker/checkouts
+VOLUME /builds/worker/.cache
```
tomprince, line 5: It'd be good to get a bug on file so that we don't need to duplicate these in dependent images.

taskcluster/docker/pushsnapscript/Dockerfile
```diff
@@ -0,0 +1,9 @@
+# %ARG DOCKER_IMAGE_PARENT
+FROM $DOCKER_IMAGE_PARENT
+
+VOLUME /builds/worker/checkouts
+VOLUME /builds/worker/.cache
+
+RUN apt-get update \
+    && apt-get install -y libsodium-dev
+RUN truncate -s 0 /etc/os-release
```
tomprince: I bet this is an existing thing, but a comment on this line would be useful. I suspect I know why this is here, but it would  be better if wasn't folk knowledge.

taskcluster/scriptworker_taskgraph/transforms/cached.py
escapewindow: Sure, I'll add a comment. It's similar to https://searchfox.org/mozilla-central/source/taskcluster/taskgraph/transforms/cached_tasks.py . I use it in xpi and adhoc; every taskgraph repo uses cached tasks for docker images and fetches... this is what keeps tasks from running if the underlying code hasn't changed.

taskcluster/scriptworker_taskgraph/transforms/cached.py
```diff
@@ -0,0 +1,95 @@
...
+            for path in sorted(list(files)):
+                h.update(
+                    "{} {}\n".format(
+                        hash_path(os.path.realpath(os.path.join(BASE_DIR, path))), path
+                    )
+                )
+            task.setdefault("attributes", {}).setdefault("cached_task", {})
+            cache_name = task["label"].replace(":", "-")
+            task["cache"] = {
+                "type": "{}.v2".format(repo_name),
```
tomprince: I *think* this should probably be `<something>.v1.<kind>` where `<something>` is hard-coded.

taskcluster/ci/config.yml
escapewindow: Let's do it now :)

taskcluster/ci/tox/kind.yml
```diff
@@ -0,0 +1,102 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+---
+loader: taskgraph.loader.transform:loader
+
+kind-dependencies:
+    - docker-image
```
tomprince, line 8: If the intent is to have the digests here depend on the docker-image digests, I think you need to explicitly add the `docker-image` dependency; The `in-tree` specification gets turned into a dependency too late to impact digests (and this is at least partly deliberate, we don't want toolchains to depend on the digest of the image that built them).

taskcluster/scriptworker_taskgraph/transforms/cached.py
```diff
@@ -0,0 +1,98 @@
...
+    repo_name = repo_name.split("/")[-1]
+
+    for task in tasks:
+        if task.get("cache", True) and not taskgraph.fast:
+            digest_data = []
+            h = hashlib.sha256()
+            h.update(
+                json.dumps(task.get("attributes", {}).get("digest-extra", {}), indent=2, sort_keys=True)
+            )
+            directories = task.get("attributes", {}).get("digest-directories", [])
```
tomprince: I don't see this getting set as an attribute anywhere, but rather as a top-level key. I think that is correct, but this needs to be adjusted to match.

taskcluster/scriptworker_taskgraph/transforms/k8s_image.py
```diff
@@ -0,0 +1,92 @@
...
+        tasks_for = config.params['tasks_for']
+        scopes = job.setdefault("scopes", [])
+        attributes = job["attributes"]
+        env = job["worker"].setdefault("env", {})
+        env["HEAD_REV"] = config.params['head_rev']
+        env["REPO_URL"] = config.params['head_repository']
+        env["PROJECT_NAME"] = project_name
+        env["TASKCLUSTER_ROOT_URL"] = "$TASKCLUSTER_ROOT_URL"
+        env["DOCKER_TAG"] = "unknown"
+        env["DOCKER_REPO"] = job.pop("docker-repo")
```
tomprince: nit: this feels a bit visually noisy, I wonder if using `env.update({ ... })` would make this easier to parse.

taskcluster/scriptworker_taskgraph/transforms/k8s_image.py
```diff
@@ -0,0 +1,92 @@
...
+            env["DOCKER_TAG"] = "pull-request"
+        elif tasks_for == 'github-push':
+            for ref_name in ("dev", "production"):
+                if config.params['head_ref'] == "refs/heads/{}-{}".format(ref_name, project_name):
+                    env["DOCKER_TAG"] = ref_name
+                    force_push_docker_image = True
+                    break
+            else:
+                if config.params['head_ref'].startswith('refs/heads/'):
+                    env["DOCKER_TAG"] = config.params['head_ref'].replace('refs/heads/', '')
```
tomprince: Is the intention that this catch `refs/heads/{dev,production}`? If so, I'd inclined to be explicit about that. 

Maybe `re.match(r'refs/heads/(dev|production)(?:-[a-z]*)$')` might be closer to what we want. (Thinking about this, maybe it might make sense to make this a graph parameter).

taskcluster/scriptworker_taskgraph/transforms/python_version.py
```diff
@@ -0,0 +1,60 @@
...
+            repl_dict = {"name": job["name"], "python_version": python_version}
+            task["label"] = _replace_string(task["label"], repl_dict)
+            task['worker']['docker-image'] = _replace_string(task['worker']['docker-image'], repl_dict)
+            task['description'] = _replace_string(task['description'], repl_dict)
+            if task.get('run', {}).get('command'):
+                task['run']['command'] = _replace_string(task['run']['command'], repl_dict)
+            if task['worker'].get('command'):
+                task['worker']['command'] = _replace_string(task['worker']['command'], repl_dict)
+            if task.get('docker-repo'):
+                task['docker-repo'] = _replace_string(task['docker-repo'], repl_dict)
```
tomprince: I think this would be a bit clearer with a loop like [here](https://searchfox.org/mozilla-central/rev/027893497316897b8f292bde48dbb6da2391a331/taskcluster/taskgraph/transforms/l10n.py#233-258) over the fields being modified.

taskcluster/ci/pr/kind.yml
escapewindow: k8s-image depends on tox. Do you want me to explicitly list both kinds?

taskcluster/scriptworker_taskgraph/transforms/k8s_image.py
escapewindow: Sure. I'm hoping we can follow up with relpro soon, which will remove the need for dev and prod branches. Given that timeframe, I didn't prioritize making these any nicer.

taskcluster/ci/pr/kind.yml
tomprince: https://hg.mozilla.org/ci/taskgraph/file/tip/src/taskgraph/transforms/code_review.py#l21 only includes tasks from kinds listed here, and not every `tox` tasks had a dependent task in `k8s-image`, so if you want those tasks to be depended on here, that kind needs to be listed.

taskcluster/ci/tox/kind.yml
escapewindow: Not yet. Should we skip artifacts if we don't have an immediate need?

tomprince: I think I would say, it depends? I couldn't think of any artifacts we obviously would want here, and if we have some, it isn't entirely clear that having a directory to dump them in is how we would want to collect them.

I definitely don't feel strongly about this.

taskcluster/ci/pr/kind.yml
escapewindow: Commited to my `taskgraph2` branch. I'll address your review comments there, make sure it works, and then pull them into this PR.

taskcluster/scriptworker_taskgraph/transforms/cached.py
escapewindow: Looks like we lose the ability to ignore `.pyc` and `.swp`, but I think that can work.

taskcluster/docker/pushapkscript/Dockerfile
escapewindow: https://bugzilla.mozilla.org/show_bug.cgi?id=1433364 ?

taskcluster/scriptworker_taskgraph/transforms/k8s_image.py
escapewindow: I think we do have to set DOCKER_TAG no matter what, or the task will fail. The below check actually checks for whether we want to push to dockerhub.

taskcluster/scriptworker_taskgraph/transforms/cached.py
escapewindow: The above transform sets the attribute.

tomprince: I don't see it set for the `tox` kind.

escapewindow: `tox` uses the cached transform. I do see the (now-renamed to `resources`) attribute in the full task graph for tox.

taskcluster/scriptworker_taskgraph/transforms/k8s_image.py
escapewindow: I'm setting `parameters["docker_tag"]` now.

taskcluster/ci/config.yml
escapewindow: Thanks for the doublecheck!

escapewindow:
@MihaiTabara @JohanLorenzo I've tweaked the `push{apk,flatpak,snap}script` docker images. If you have cycles to sanity check those, please do.

taskcluster/scriptworker_taskgraph/transforms/cached.py
MihaiTabara: I was under the impression that it's based on the routes. Like using taskgraph in a repo generically sets specific routes, which are downstream used in taskgraph to find previously-run tasks. E.g. https://hg.mozilla.org/ci/taskgraph/file/tip/src/taskgraph/util/cached_tasks.py#l20 - but looking at that code again, maybe we use that only for toolchain tasks or so? 

Ah, looking at the URl you pointed, seems like that's used under the hood as well - https://searchfox.org/mozilla-central/source/taskcluster/taskgraph/transforms/cached_tasks.py#75 - I wonder why we never added this transform in any of the mobile projects. I'll dig further.

MihaiTabara:
> @MihaiTabara @JohanLorenzo I've tweaked the `push{apk,flatpak,snap}script` docker images. If you have cycles to sanity check those, please do.

LGTM

pushsnapscript/Dockerfile
```diff
@@ -3,20 +3,15 @@ FROM python:3.8
...
- && truncate -s 0 /etc/os-release \
- && ln -s /app/docker.d/healthcheck /bin/healthcheck
+RUN ln -s /app/docker.d/healthcheck /bin/healthcheck
 
-USER app
 WORKDIR /app
-
 COPY . /app
 
+RUN /app/docker.d/image_setup.sh libsodium-dev squashfs-tools xdelta3
```
JohanLorenzo, line 12: Note to self: we pass these arguments here because `Dockerfile.test` only install `libsodium-dev`.

taskcluster/ci/config.yml
```diff
@@ -0,0 +1,47 @@
...
+
+task-priority: high
+
+taskgraph:
+    register: scriptworker_taskgraph:register
+    decision-parameters: 'scriptworker_taskgraph.parameters:get_decision_parameters'
+    repositories:
+        scriptworker:
+            name: "Scriptworker-scripts"
+            project-regex: scriptworker-scripts$
```
JohanLorenzo, line 12: Question: what does `project-regex` do?

taskcluster/ci/k8s-image/kind.yml
```diff
@@ -0,0 +1,55 @@
...
+    - scriptworker_taskgraph.transforms.python_version:transforms
+    - scriptworker_taskgraph.transforms.k8s_image:transforms
+    - scriptworker_taskgraph.transforms.cached:transforms
+    - taskgraph.transforms.cached_tasks:transforms
+    - taskgraph.transforms.task:transforms
+
+job-defaults:
+    description: "{name} k8s image builder py{python_version}"
+    name: "{name}-python{python_version}"
+    run-on-tasks-for: ["action", "github-pull-request", "github-push"]
```
JohanLorenzo, line 20: Question: I've never used `action` in `run-on-tasks-for` before. What's the expected behavior, in this case?

taskcluster/ci/k8s-image/kind.yml
```diff
@@ -0,0 +1,55 @@
...
+            - >-
+              sh /usr/local/bin/build_and_push.sh
+    python-versions:
+        - 38
+    deploy-secret-url: http://taskcluster/secrets/v1/secret/project/releng/scriptworker-scripts/deploy
+    docker-repo: mozilla/releng-{name}
+    # TODO copy image to artifacts
+
+jobs:
+    addonscript: {}
```
JohanLorenzo, line 46: Followup: I wonder if we should create a dedicated loader to avoid repeating each subproject here. For instance, there are more than a hundred subprojects all defined [in this file](https://github.com/mozilla-mobile/android-components/blob/6744e62723898db1c5beb24f926303d9279b6030/.buildconfig.yml). [This loader](https://github.com/mozilla-mobile/android-components/blob/6744e62723898db1c5beb24f926303d9279b6030/taskcluster/ac_taskgraph/loader/build_config.py) reads that file, generates the right jobs and applies [the default configuration](https://github.com/mozilla-mobile/android-components/blob/6744e62723898db1c5beb24f926303d9279b6030/taskcluster/ci/build/kind.yml#L20) before passing jobs to transforms.

I see we repeat the list of project here and in another kind so maybe we want a dedicated loader too.

taskcluster/docker/k8s-image/Dockerfile
```diff
@@ -0,0 +1,27 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# We need to run a really old docker version because taskcluster is using a
+# really old version in their setup
+FROM          docker:1.6.2
```
JohanLorenzo: Nit: 

```suggestion
FROM docker:1.6.2
```

taskcluster/docker/pushsnapscript/Dockerfile
```diff
@@ -0,0 +1,11 @@
+# %ARG DOCKER_IMAGE_PARENT
+FROM $DOCKER_IMAGE_PARENT
+
+VOLUME /builds/worker/checkouts
+VOLUME /builds/worker/.cache
+
+# %include pushsnapscript/docker.d/image_setup.sh
+COPY topsrcdir/pushsnapscript/docker.d/image_setup.sh /usr/local/bin/
+RUN chmod a+x /usr/local/bin/image_setup.sh
```
JohanLorenzo: Nit: Why does this script need to have new permissions?

taskcluster/scriptworker_taskgraph/parameters.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def get_decision_parameters(graph_config, parameters):
+    """Add repo-specific decision parameters.
+
+    If we're on a production- or dev- branch, detect and set the `script_name`.
+
+    """
+    if parameters["tasks_for"] == "github-pull-request":
+        parameters["docker_tag" ] = "github-pull-request"
```
JohanLorenzo: Nit: 

```suggestion
        parameters["docker_tag"] = "github-pull-request"
```

taskcluster/scriptworker_taskgraph/parameters.py
```diff
@@ -0,0 +1,55 @@
...
+    elif parameters["head_ref"].startswith("refs/heads/"):
+        parameters["docker_tag"] = parameters["head_ref"].replace("refs/heads/", "")
+        force_push = False
+        for prefix, tag in PROJECT_SPECIFIC_PREFIXES.items():
+            if parameters["head_ref"].startswith(prefix):
+                parameters["script_name"] = parameters["head_ref"].replace(prefix, "")
+                parameters["docker_tag" ] = tag
+                force_push = True
+                break
```
JohanLorenzo, line 50: Improvement for later: Should we bail out if 2 prefixes happen to match? Otherwise, the decision task may silently ignore the error.

taskcluster/scriptworker_taskgraph/parameters.py
```diff
@@ -0,0 +1,55 @@
...
+PROJECT_SPECIFIC_PREFIXES = {
+    "refs/heads/dev-": "dev",
+    "refs/heads/production-": "production",
+}
+
+PUSH_TAGS = ("dev", "production")
+
+scriptworker_schema = {
+    Optional('docker_tag'): Any(basestring, None),
+    Optional('push_docker_image'): Any(True, "force", False, None),
```
JohanLorenzo, line 25: Nit: `False` is never used. I wonder if we should default to it, instead of `None`. That might help making allowed values more explicit.

taskcluster/scriptworker_taskgraph/parameters.py
```diff
@@ -0,0 +1,55 @@
...
+
+    """
+    if parameters["tasks_for"] == "github-pull-request":
+        parameters["docker_tag" ] = "github-pull-request"
+    elif parameters["head_ref"].startswith("refs/heads/"):
+        parameters["docker_tag"] = parameters["head_ref"].replace("refs/heads/", "")
+        force_push = False
+        for prefix, tag in PROJECT_SPECIFIC_PREFIXES.items():
+            if parameters["head_ref"].startswith(prefix):
+                parameters["script_name"] = parameters["head_ref"].replace(prefix, "")
```
JohanLorenzo, line 47: Followup: I wonder if we should bail out here when `script_name` doesn't match any known script. That will prevent the decision task to be green while it scheduled nothing.

taskcluster/scriptworker_taskgraph/transforms/cached.py
```diff
@@ -0,0 +1,63 @@
...
+        if resources:
+            task.setdefault("attributes", {})["resources"] = resources
+        yield task
+
+
+@transforms.add
+def build_cache(config, tasks):
+    repo_name = subprocess.check_output(["git", "remote", "get-url", "origin"]).rstrip()
+    repo_name = repo_name.replace(".git", "").rstrip("/")
+    repo_name = repo_name.split("/")[-1]
```
JohanLorenzo: Nit: `repo_name` seems to be unused past this point. I'm not sure what's it for.

taskcluster/scriptworker_taskgraph/transforms/cached.py
```diff
@@ -0,0 +1,63 @@
...
+transforms = TransformSequence()
+
+BASE_DIR = os.getcwd()
+
+@transforms.add
+def add_resources(config, tasks):
+    for task in tasks:
+        resources = task.pop("resources", None)
+        if resources:
+            task.setdefault("attributes", {})["resources"] = resources
```
JohanLorenzo: Nit: should we bail out if `task["attributes"]["resources"]` has already been defined before this point? If not, should we avoid overwriting it?

taskcluster/scriptworker_taskgraph/transforms/cached.py
```diff
@@ -0,0 +1,63 @@
...
+    repo_name = repo_name.replace(".git", "").rstrip("/")
+    repo_name = repo_name.split("/")[-1]
+
+    for task in tasks:
+        if task.get("cache", True) and not taskgraph.fast:
+            digest_data = []
+            digest_data.append(
+                json.dumps(task.get("attributes", {}).get("digest-extra", {}), indent=2, sort_keys=True)
+            )
+            resources = task.get("attributes", {}).get("resources", [])
```
JohanLorenzo: Nit: Should `task["attributes"]["resources"]` be set in `add_resources()` so we don't have to be cautious here?

taskcluster/scriptworker_taskgraph/transforms/python_version.py
```diff
@@ -0,0 +1,91 @@
...
+        "worker.docker-image",
+    ]
+    for job in jobs:
+        for python_version in job.pop("python-versions"):
+            task = deepcopy(job)
+            subs = {"name": job["name"], "python_version": python_version}
+            for field in fields:
+                _resolve_replace_string(task, field, subs)
+            task["attributes"]["python-version"] = python_version
+            yield task
```
JohanLorenzo, line 71: Followup: I wonder if the python version logic shouldn't be part of the dedicated loader. In android-components we also have 2 dimensions: 

1. Subprojects
2. Build types (regular, nightly, release)

The specific loader handles [the latter there](https://github.com/mozilla-mobile/android-components/blob/6744e62723898db1c5beb24f926303d9279b6030/taskcluster/ac_taskgraph/loader/build_config.py#L29).

JohanLorenzo (approved):
Looks great to me overall, nice job! 👍 I'm glad taskgraph has some use for projects like scriptworker-scripts! I think it's quite suited because of the repetition between projects. 

I found a few things that can be improved in followed up as well as minor nits. I don't see anything that immediately blocks this PR.

pushsnapscript/Dockerfile
escapewindow: Yeah. I thought we could use something like this, rather than install extra packages on the test images.

taskcluster/ci/config.yml
escapewindow: It helps us determine which repo is which when we have multiple repositories: https://hg.mozilla.org/ci/taskgraph/file/tip/src/taskgraph/transforms/base.py#l69 .

taskcluster/ci/k8s-image/kind.yml
escapewindow: We're probably going to use a different filter for actions, so it's possible this is extraneous. I originally used it as a way to filter for relpro, but we can also specifically filter for relpro elsewhere.

taskcluster/ci/k8s-image/kind.yml
escapewindow: Sure, possibly. We could add configuration as to whether we want tox, which versions of python, and whether we want a k8s image (and which versions of python). I think the current breakdown works too.

taskcluster/docker/pushsnapscript/Dockerfile
escapewindow: It doesn't. I can remove this.

taskcluster/scriptworker_taskgraph/parameters.py
escapewindow: Sure. I'm hoping we follow up with relpro and don't rely on this dev-/production-scriptname behavior for much longer, but that's a good sanity check if we keep this behavior.

taskcluster/scriptworker_taskgraph/transforms/python_version.py
escapewindow: Could do. [If we add a non-python script](https://github.com/mozilla-releng/scriptworker-scripts/pull/211), we either have to adjust the transforms or the loader, depending on which solution we go with. Which would make more sense?

Edit: thinking in this case, the transforms would make more sense. The `tox` kind is python-specific, so no changes there. `k8s-image` would have to support both python and non-python, so the `python_version` transforms could pass any non-python tasks through unchanged, and the `rust_version` (or other) transforms could pass any non-rust tasks through unchanged. Putting all of this in the loader might be a bit much, but it might be doable.

taskcluster/scriptworker_taskgraph/transforms/cached.py
tomprince: `taskgraph.transfrorms.cached_task` is a transform that takes a `cache` value on a task (with cache type, name and digest data), and turns it into the appropriate routes and optimizations so that the task will be re-used as long as the digest data is not changed.

That transform is used by the docker-image, toolchain and fetch transforms to implement their caching behavior. This transform is similar to those, in that it generates a `cache` value with appropriate digest to be consumed by `taskgraph.transforms.cached_task`; but unlike them, does not expect the task to have any particular artifacts, and doesn't have any other bits of taskgraph that are designed to interact with it.

> I wonder why we never added this transform in any of the mobile projects. I'll dig further.

It is not obvious to me where it would be useful for mobile projects. The most likely would be android-components or application-services. Though, given that android-components has a unified build system, I suspect it would be somewhat less useful for that project.

To a certain extent, this transform is being used to work-around the fact that we don't have a push-log for this repo, which would let us see if any relevant files changed since the last push. On the other hand, this precise tracking, as backing something out will not rerun tasks on the old code.

taskcluster/scriptworker_taskgraph/transforms/cached.py
escapewindow: Hm. We may inherit the k8s-image resources attribute from the tox kind, and intentionally overwrite.

Edit: no, I adjusted the checks. Now we inherit and don't overwrite with `[]` if it's set.

taskcluster/scriptworker_taskgraph/parameters.py
escapewindow: Not sure I understand. Are you saying `head_ref` may start with both `refs/heads/dev-` and `refs/heads/production-` simultaneously?

taskcluster/scriptworker_taskgraph/parameters.py
escapewindow: Sure. I couldn't find how to set default values on extended schemas; do you have an example?

tomprince: We don't support defaults in schemas. This comes from m-c, where `--fast` disables schema validation, which implements implements defaults by returning the new structure. Due to that, we ignore the return value, to make sure `--fast` doesn't change behavior.

taskcluster/scriptworker_taskgraph/transforms/python_version.py
tomprince: It is not clear to me that the python-version transform is providing value for the `k8s-image` kind in any case.

escapewindow: I could see us creating k8s images for multiple python versions at some point, especially after we drop support for docker hub. I would like to land this PR at some point sooner rather than later, as well.

tomprince: Yeah, I definitely wasn't suggesting changing anything here, just commenting on the future consideration with non-python things

PR #229, opened 2020-06-10
Bug 1444406 - use https for update downloads on DevEdition and Firefox beta

nthomas-mozilla:


nthomas-mozilla:
I did consider adding the protocol in-tree but would still have need as much change here, and gained all sorts of fun uplifting to multiple branches and products. So while it's not great that we still hard code the protocol inside balrogscript it's not really any worse either, at least once we have swapped everything to https.

nthomas-mozilla:
Testing on dev worker using `mach try scriptworker`, redoing toplevel tasks for recent releases in staging:
* 77.0.1 - [PPMw8VHrTp2OSARRXslSVg](https://firefox-ci-tc.services.mozilla.com/tasks/PPMw8VHrTp2OSARRXslSVg/runs/1) - still http://
* 78.0b4 - [ah2_8t2VQ9udiDIFM7F0fw](https://firefox-ci-tc.services.mozilla.com/tasks/ah2_8t2VQ9udiDIFM7F0fw/runs/1) - now https://

bhearsum (approved):

balrogscript/src/balrogscript/submitter/cli.py
```diff
@@ -30,6 +30,12 @@ def _getFileUrls(self, productName, version, buildNumber, updateChannels, ftpSer
         file_prefix = productName.lower()
         if file_prefix == "devedition":
             file_prefix = "firefox"
+        # bug 1444406 - eventually we're going to default to https
+        protocol = "http"
+        if productName.lower() == "devedition":
+            protocol = "https"
+        if productName.lower() == "firefox" and all([c.startswith(("beta",)) for c in updateChannels]):
```
catlee: Is this to prevent RC builds from also using https for now?

nthomas-mozilla: We spoke on video, so to record and correct what I said ... the idea is to restrict to beta for now, and later extend the tuple in startswith to add release, and maybe esr later still (or maybe do it all in one go, TBC). 

For the RC case, the main thing is to preserve http for release channels, and if we lose https we've had for the rest of the beta cycle then OK, that should only be temporary. But in fact we have two toplevel submit tasks for RCs, and the `release-secondary-balrog-submit-toplevel` kind (for RC-on-beta) will end up traversing the changes here too. I've checked in staging and we would get https for the beta channels, but there's a problem. The ordering of the two tasks matters because the last task overwrites the `"*"` match for completes - we actually want the secondary to run first, then the 'primary' to come along an set http for `"*"`. Turns out the task order is not stable because there is no dependency relationship between them. I'll see if I can come up with a solution.

nthomas-mozilla: 5750195 resolves this by only writing the `"*"` in the main toplevel task, and secondary adds beta/beta-cdntest/beta-localtest. @mozbhearsum could you double-check the logic and test case ?

bhearsum (approved):

PR #237, opened 2020-06-17
Bug 1644973 - mutate Nazgul jobs to become new vanilla

MihaiTabara:
* Tuxedo is dead so we no longer need to talk it
* Nazgul mutates to become the new vanilla
* remove all `nazgul` mentions to avoid confusion in the future

First I need to push my SOPS changes, then https://github.com/mozilla-services/cloudops-infra/pull/2240/ and then I merge this.


rail (approved):
Woot!

PR #241, opened 2020-06-24
always run k8s-image tasks on dev and prod

escapewindow:
- Fix `target_tasks`
- always run the k8s-image tasks on dev and production branch pushes.

### DIFF
--- a/taskcluster/scriptworker_taskgraph/target_tasks.py
+++ b/taskcluster/scriptworker_taskgraph/target_tasks.py
@@ -16,4 +16,18 @@ def target_tasks_default(full_task_graph, parameters, graph_config):
             return False
         if parameters.get("script_name"):
             return task.attributes.get("script-name") == parameters["script_name"]
+        return True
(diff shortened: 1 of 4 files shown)


taskcluster/scriptworker_taskgraph/target_tasks.py
```diff
@@ -16,4 +16,18 @@ def filter(task, parameters):
...
+
+@_target_task('docker-hub-push')
+def target_tasks_default(full_task_graph, parameters, graph_config):
+    """Filter by `run_on_tasks_for` and `script-name`."""
+
+    def filter(task, parameters):
+        if task.kind != "k8s-image":
+            return False
+        if parameters.get("script_name"):
+            return task.attributes.get("script-name") == parameters["script_name"]
```
tomprince: (optional) Could turn this into a `filter_for_script` function (that returns `True` if the parameter is unset) and then use a style like [this](https://searchfox.org/mozilla-central/source/taskcluster/taskgraph/target_tasks.py#292-295).

tomprince (approved):

Issue #73, opened 2019-11-16
impl treescript method that determines most recent good rev


lundjordan:

lundjordan:
this is so we can determine which rev is the most recent rev that is sane to release from.

e.g. will have builds, not a no-op, etc

lundjordan:
this is going to be done in shipitscript now

Issue #17, opened 2019-08-14
\*scripts should log their version on start.


escapewindow:
and/or revision.

We don't currently bump versions, and we may not have the git repo around to find the revision. Any solution here likely needs to work for both the docker-based scripts, and the mac-based scripts.

We may want to have a convention of creating a flatfile in a known location, with the scriptworker version and script revision, after creating the dockerfile and/or puppetizing the mac. Then it's a matter of reading this `motd` (message of the task, `mott`?) at the beginning if that file exists, possibly in a simple `scriptworker_client` function.

PR #247, opened 2020-07-29
Bug 1608874 - part 2: Remove Fennec Nightly/Beta/Release production w…

JohanLorenzo:
…orkers

We shouldn't sign/beetmover/ship any Fennec anymore. This PR removes the ability for gecko folks to use the real keys and the real accounts. I left the staging ones around because I foresee people willing to test something against Fennec. What do you all think?

### DIFF
--- a/beetmoverscript/docker.d/worker.yml
+++ b/beetmoverscript/docker.d/worker.yml
@@ -44,7 +44,6 @@ bucket_config:
           buckets:
             devedition: 'net-mozaws-prod-delivery-archive'
             firefox:    'net-mozaws-prod-delivery-firefox'
-            fennec:     'net-mozaws-prod-delivery-archive'
             mobile:     'net-mozaws-prod-delivery-archive'
           url_prefix: 'https://archive.mozilla.org'
         release:
@@ -54,7 +53,6 @@ bucket_config:
           buckets:
             devedition: 'net-mozaws-prod-delivery-archive'
             firefox:    'net-mozaws-prod-delivery-firefox'
-            fennec:     'net-mozaws-prod-delivery-archive'
             mobile:      'net-mozaws-prod-delivery-archive'
           url_prefix: 'https://archive.mozilla.org'
         partner:
@@ -70,7 +68,6 @@ bucket_config:
           buckets:
             devedition: 'net-mozaws-stage-delivery-archive'
             firefox:    'net-mozaws-stage-delivery-firefox'
-            fennec:     'net-mozaws-stage-delivery-archive'
             mobile:     'net-mozaws-stage-delivery-archive'
           url_prefix: 'https://ftp.stage.mozaws.net'
         dep-partner:
--- a/pushapkscript/docker.d/worker.yml
+++ b/pushapkscript/docker.d/worker.yml
@@ -14,33 +14,6 @@ taskcluster_scope_prefixes:
 products:
   $flatten:
     $match:
-      'COT_PRODUCT == "firefox" && ENV == "prod"':
-        - product_names: [ "beta", "aurora", "release" ]
-          digest_algorithm: SHA1
-          override_channel_model: 'choose_google_app_with_scope'
-          apps:
-            release:
-              default_track: production
-              service_account: { "$eval": "GOOGLE_PLAY_SERVICE_ACCOUNT_FIREFOX_RELEASE" }
-              package_names:
-                - "org.mozilla.firefox"
-              credentials_file: { "$eval": "GOOGLE_CREDENTIALS_FIREFOX_RELEASE_PATH" }
-              certificate_alias: release
-            beta:
-              default_track: production
-              service_account: { "$eval": "GOOGLE_PLAY_SERVICE_ACCOUNT_FIREFOX_BETA" }
-              package_names:
-                - "org.mozilla.firefox_beta"
-              credentials_file: { "$eval": "GOOGLE_CREDENTIALS_FIREFOX_BETA_PATH" }
-              certificate_alias: release
-            aurora:
-              default_track: beta
-              service_account: { "$eval": "GOOGLE_PLAY_SERVICE_ACCOUNT_FIREFOX_AURORA" }
-              package_names:
-                - "org.mozilla.fennec_aurora"
-              credentials_file: { "$eval": "GOOGLE_CREDENTIALS_FIREFOX_AURORA_PATH" }
-              certificate_alias: nightly
-
       'COT_PRODUCT == "firefox" && ENV != "prod"':
         - product_names: [ "dep" ]
           digest_algorithm: SHA1
--- a/signingscript/docker.d/passwords.yml
+++ b/signingscript/docker.d/passwords.yml
@@ -194,11 +194,6 @@ in:
                ["autograph_mar384", "autograph_hash_only_mar384"]
               ]
             - ["https://autograph-external.prod.autograph.services.mozaws.net",
-               {"$eval": "AUTOGRAPH_FENNEC_RELEASE_USERNAME"},
-               {"$eval": "AUTOGRAPH_FENNEC_RELEASE_PASSWORD"},
-               ["autograph_apk_fennec_sha1"]
-               ]
-            - ["https://autograph-external.prod.autograph.services.mozaws.net",
                {"$eval": "AUTOGRAPH_GPG_USERNAME"},
                {"$eval": "AUTOGRAPH_GPG_PASSWORD"},
                ["autograph_gpg"]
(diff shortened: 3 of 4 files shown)


beetmoverscript/docker.d/worker.yml
```diff
@@ -70,7 +68,6 @@ bucket_config:
           buckets:
             devedition: 'net-mozaws-stage-delivery-archive'
             firefox:    'net-mozaws-stage-delivery-firefox'
-            fennec:     'net-mozaws-stage-delivery-archive'
```
MihaiTabara, line 73: Hm, I see you've removed the `dep` fennec behavior here in favor of https://github.com/mozilla-releng/scriptworker-scripts/pull/247/files#diff-a6a8f5e85d3a21ae2c10bb948323a917R23. Just to confirm, we want to keep the DEV workers around for people to still test or do we want the staging workers (`prod` environment but `dep` keys, in this case). Or better said, which ones of the following to we want to keep:
a) https://firefox-ci-tc.services.mozilla.com/provisioners/scriptworker-k8s/worker-types/gecko-1-beetmover
b) https://firefox-ci-tc.services.mozilla.com/provisioners/scriptworker-k8s/worker-types/gecko-1-beetmover-dev

Based on the history, I presume we'd want to keep the staging ones (which is a) since that's used in staging releases triggered from gecko). b) type of workers are mostly used by RelEng only, with few exceptions. 

Depending on which we want to keep, we might need to adjust this file. If we want to keep a) we need to revert line #73 and remove line #23. If we wanted to keep b) instead, than we're good.

pushapkscript/docker.d/worker.yml
```diff
@@ -14,33 +14,6 @@ taskcluster_scope_prefixes:
 products:
   $flatten:
     $match:
-      'COT_PRODUCT == "firefox" && ENV == "prod"':
-        - product_names: [ "beta", "aurora", "release" ]
-          digest_algorithm: SHA1
```
MihaiTabara, line 19: 👋 goodbye SHA1!

beetmoverscript/docker.d/worker.yml
```diff
@@ -44,7 +44,6 @@ bucket_config:
           buckets:
             devedition: 'net-mozaws-prod-delivery-archive'
             firefox:    'net-mozaws-prod-delivery-firefox'
-            fennec:     'net-mozaws-prod-delivery-archive'
```
MihaiTabara, line 47: This makes sense. For beetmover we can't remove the workers themselves altogether from CloudOps as they are shared for other products as well. So yeah, trimming these buckets is the best we can do for now in beetmoverscript.

pushapkscript/docker.d/worker.yml
```diff
@@ -14,33 +14,6 @@ taskcluster_scope_prefixes:
...
-          override_channel_model: 'choose_google_app_with_scope'
-          apps:
-            release:
-              default_track: production
-              service_account: { "$eval": "GOOGLE_PLAY_SERVICE_ACCOUNT_FIREFOX_RELEASE" }
```
MihaiTabara, line 24: @JohanLorenzo I reckon we can remove these env vars from cloudops-infra repo too, and also the dedicated workers themselves along with their credentials from SOPS, right?

pushapkscript/docker.d/worker.yml
```diff
@@ -14,33 +14,6 @@ taskcluster_scope_prefixes:
...
-              certificate_alias: release
-            aurora:
-              default_track: beta
-              service_account: { "$eval": "GOOGLE_PLAY_SERVICE_ACCOUNT_FIREFOX_AURORA" }
-              package_names:
-                - "org.mozilla.fennec_aurora"
-              credentials_file: { "$eval": "GOOGLE_CREDENTIALS_FIREFOX_AURORA_PATH" }
-              certificate_alias: nightly
-
       'COT_PRODUCT == "firefox" && ENV != "prod"':
```
MihaiTabara, line 17: Hm, this is tricky. Looking at our workers, we have three of them in total:
1. the `prod` one from lines #17-43 which you correctly removed in this file - which is basically https://firefox-ci-tc.services.mozilla.com/provisioners/scriptworker-k8s/worker-types/gecko-3-pushapk
2. the `fake-prod` which is staging used in try staging releases - https://firefox-ci-tc.services.mozilla.com/provisioners/scriptworker-k8s/worker-types/gecko-1-pushapk
3. the `dev` which is used in dev workers - https://firefox-ci-tc.services.mozilla.com/provisioners/scriptworker-k8s/worker-types/gecko-1-pushapk-dev

the ENV for 2) is `fake-prod` while for 3) is `dev` so both of them theoretically check this logic. I wonder which one of them we want to get rid of, if any. Or keep them both for now to avoid adding narrowing logic here since we'll likely get rid of Fennec altogether by EOQ3.

signingscript/docker.d/passwords.yml
```diff
@@ -193,11 +193,6 @@ in:
                {"$eval": "AUTOGRAPH_MAR_RELEASE_PASSWORD"},
                ["autograph_mar384", "autograph_hash_only_mar384"]
               ]
-            - ["https://autograph-external.prod.autograph.services.mozaws.net",
-               {"$eval": "AUTOGRAPH_FENNEC_RELEASE_USERNAME"},
```
MihaiTabara, line 197: @JohanLorenzo same as for pushapk, do you want me to remove these env vars from the associated workers altogether?

signingscript/docker.d/passwords.yml
```diff
@@ -230,11 +225,6 @@ in:
                {"$eval": "AUTOGRAPH_MAR_NIGHTLY_PASSWORD"},
                ["autograph_mar384", "autograph_hash_only_mar384"]
               ]
-            - ["https://autograph-external.prod.autograph.services.mozaws.net",
```
MihaiTabara, line 233: @JohanLorenzo same as for pushapk, do you want me to remove these env vars from the associated workers altogether?

MihaiTabara (approved):
This overall LGTM, but I have a few clarifying questions.
Also, I can think of another few places where we can cleanup (ESR68 Fennec related taskgraph, declarative artifacts for Fennec, etc) but are unrelated to this PR. I presume you wanted to remove the ability to sign/beetmover/ship for now at the end of the pipeline while in the meantime we can continue the cleanup. Correct?

beetmoverscript/docker.d/worker.yml
JohanLorenzo: Nice! I wasn't sure whether we could anything else!

pushapkscript/docker.d/worker.yml
JohanLorenzo: The fennec nightly account can be deleted!

The Google Play accounts for `beta` and `release` are still used by the `mobile` workers. I think we should keep them around.

signingscript/docker.d/passwords.yml
JohanLorenzo: Let's keep this one around. It's used by `mobile`.

signingscript/docker.d/passwords.yml
JohanLorenzo: This one can go!

pushapkscript/docker.d/worker.yml
JohanLorenzo: I'd go for the latter. I don't know how `dev` and `fake-prod` are used. I'd prefer to not change anything and get everything removed at once.

beetmoverscript/docker.d/worker.yml
JohanLorenzo: Thanks for this explanation! I don't know how a) and b) are used in the real world. I'd tend to keep them around and delete all of them once esr68 is officially decommissioned. What do you think?

MihaiTabara: Keeping them both until we can fully decommision is a good idea, I agree. But just to confirm, in its current form, this patch will only keep Fennec behavior available in beetmover in the DEV workers, and not in the staging workers. If that's what we want, that's okay.

pushapkscript/docker.d/worker.yml
MihaiTabara: Yeah, I meant the nightly ones, sorry for confusion. Since we don't have dedicated workers for just nightly, we can only remove SOPS credentials and the exporting of these env vars for the nightly ones, in this cleanup.

pushapkscript/docker.d/worker.yml
MihaiTabara: Okay, then let's keep both and not (additionally) touch this file until we do the final cleanup once we EOL the ESR68 👍

signingscript/docker.d/passwords.yml
MihaiTabara: Right, good point.

PR #248, opened 2020-07-29
Bug 1652979 - Stop signing/shipping fennecNightly, fenixNightly, and …

JohanLorenzo:
…fenixBeta

Depends on https://github.com/mozilla-mobile/fenix/pull/12225. Do not land yet, we may want to revert that Fenix PR.

PR #249, opened 2020-07-31
Bug 1608874 - part 3:cleanup for Fennec nightly, Fenix nightly and Fenix beta

MihaiTabara:
Follow-up for https://github.com/mozilla-releng/scriptworker-scripts/pull/248/ and https://github.com/mozilla-releng/scriptworker-scripts/pull/247/ with more cleanup.

@JohanLorenzo 
I'm not touching the neither the workers nor SOPS so that we can self-serve in case we need to backout any of this(not that I expect us to, given the rollout but still, as a precautionary in case we're side effecting mobile workers). We can fully cleanup the workers once the migration is completed. The `min_replicas` is set to 0 anyway so for now, should there's no release promotion on ESR68 and the secrets are no longer on the workers, it's impossible to ship so I think it's good enough for now. 

What do you think?


JohanLorenzo (approved):
Thanks for taking care of these. I had forgotten them indeed 😅

PR #250, opened 2020-08-04
Allow taskcluster to run on pull-requests from non-collaboarators.

tomprince:


escapewindow:
<s>We may want to hold off on this until we're no longer using a docker hub secret.</s>

tomprince:
That secret is only available to the dev* and production* branches (see [here]). There are some additional scopes that we should remove first though: https://phabricator.services.mozilla.com/D85929

escapewindow (approved):

PR #252, opened 2020-08-06
Use isort <5 because it has different output whether it is on py3.7 o…

JohanLorenzo:
…r py3.8


escapewindow (approved):
We could also start looking at a) what's still using py37 (probably some of the unpuppetized macs) b) when should we remove support for py37 and start testing against py39beta ?

If py38/py39 both work well with `isort>=5`, we could add a comment here to remove the limitation once we EOL py37 support.

JohanLorenzo:
These are good points. I confess I made this hotfix because I'd like to get https://github.com/mozilla-releng/scriptworker-scripts/pull/242 landed. I'm happy to take on followups :)

PR #253, opened 2020-08-07
Update to isort 5

JohanLorenzo:
Follows https://github.com/mozilla-releng/scriptworker-scripts/pull/252 up. 

@Callek found out isort doesn't run on py3.7. So fixing the failures I saw should just be a matter of applying the changes required by isort 5. https://timothycrosley.github.io/isort/docs/upgrade_guides/5.0.0/


pushsnapscript/src/pushsnapscript/snap_store.py
```diff
@@ -5,11 +5,7 @@
 from mozilla_version.version import VersionType
 from scriptworker.exceptions import TaskVerificationError
 from scriptworker.utils import get_hash, get_single_item_from_sequence
-
-# XXX Hack to only import a subset of snapcraft. Otherwise snapcraft can't be built on any other
```
JohanLorenzo, line 9: This comment was making `black` and `isort` to output different results. It turns out it's not a valid comment anymore. That's why this change is here.

escapewindow (approved):

PR #254, opened 2020-08-18
l10n-bumper - add capability to poll hg pushlog for latest revision info

escapewindow:
Elmo is slated to be retired. Instead of getting signoffs in Elmo and
taking the latest signed off revisions as the revisions in
l10n-changesets, let's poll hg pushlog for the latest revision unless
the `pin` boolean is set in the locale.

### DIFF
--- a/treescript/src/treescript/l10n.py
+++ b/treescript/src/treescript/l10n.py
@@ -4,13 +4,15 @@
 Largely from https://hg.mozilla.org/mozilla-central/file/63ef0618ec9a07c438701e0357ef0d37abea0dd8/testing/mozharness/scripts/l10n_bumper.py
 
 """
+import asyncio
 import json
 import logging
 import os
 import pprint
 import tempfile
+from copy import deepcopy
 
-from scriptworker_client.aio import download_file, retry_async
+from scriptworker_client.aio import download_file, retry_async, semaphore_wrapper
 from scriptworker_client.exceptions import DownloadError
 from scriptworker_client.utils import load_json_or_yaml
 from treescript.mercurial import run_hg_command
@@ -68,6 +70,8 @@ def build_platform_dict(bump_config, repo_path):
             # locale is 1st word in line in shipped-locales
             if platform_config.get("format") == "shipped-locales":
                 locale = locale.split(" ")[0]
+            if locale in ("en-US",):
+                continue
             existing_platforms = set(platform_dict.get(locale, {}).get("platforms", []))
             platforms = set(platform_config["platforms"])
             ignore_platforms = set(ignore_config.get(locale, []))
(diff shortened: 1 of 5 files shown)


escapewindow:
I'm still running tests, but I think this is ready for a review pass.

treescript/src/treescript/l10n.py
```diff
@@ -77,26 +81,82 @@ def build_platform_dict(bump_config, repo_path):
...
+
+    Args:
+        locale (str): the locale to query
+        url (str): the [templatized] pushlog url
+
+    Returns:
+        tuple (locale, revision)
+
+    """
+    url = url % {"locale": locale}
```
bhearsum, line 96: nit: this should probably use `.format`

treescript/src/treescript/l10n.py
```diff
@@ -77,26 +81,82 @@ def build_platform_dict(bump_config, repo_path):
...
+    with tempfile.NamedTemporaryFile() as fp:
```
bhearsum, line 97: nit: is there a reason this has to go to disk instead of just keeping it in memory? size?

treescript/src/treescript/l10n.py
```diff
@@ -77,26 +81,82 @@ def build_platform_dict(bump_config, repo_path):
...
+        path = fp.name
+        await retry_async(download_file, args=(url, path), retry_exceptions=(DownloadError,))
+        revision_info = load_json_or_yaml(path, is_path=True)
+    last_push_id = revision_info["lastpushid"]
+    revision = revision_info["pushes"][str(last_push_id)]["changesets"][0]
```
bhearsum, line 102: Are these pushes guaranteed to only have one changeset? It might be safer to take `[-1]` instead of `[0]`.

treescript/src/treescript/l10n.py
```diff
@@ -77,26 +81,82 @@ def build_platform_dict(bump_config, repo_path):
...
+                value["pin"] = old_contents[locale]["pin"]
+            else:
+                tasks.append(asyncio.create_task(semaphore_wrapper(semaphore, get_latest_revision(locale, bump_config["l10n_repo_url"]))))
+                value["pin"] = False
+            revision_dict[locale] = value
+        await asyncio.gather(*tasks)
+        for task in tasks:
+            (locale, revision) = task.result()
+            revision_dict[locale]["revision"] = revision
+    elif dashboard_revision_info:
```
bhearsum, line 157: can we kill this branch of the code?

bhearsum (approved):
This looks fine. Do we need any patches to central, beta, or release to pin some locales before this lands?

treescript/src/treescript/l10n.py
escapewindow: I think beta and esr68 will use this branch of code til a) central-> beta merge day, and b) we turn off the bumper in esr68. Once those happen, then yes, I'd love to clean this up.

treescript/src/treescript/l10n.py
escapewindow: As long as we have `tipsonly=1` in the query args, that appears to be true.

treescript/src/treescript/l10n.py
escapewindow: > This looks fine. Do we need any patches to central, beta, or release to pin some locales before this lands?

Thanks! No, pinning will likely be manual once we determine the default revision for a locale is bad, and we're unable to fix the locale sooner.

**[Edit]** Er, this comment shouldn't have gone into this thread.

escapewindow: No, no reason.

escapewindow: Hm, I was going to port over `scriptworker.util.load_json_or_yaml_from_url` but that downloads the file and loads it from a tempfile as well. I'll probably leave this for another patch if we want to optimize.

treescript/src/treescript/l10n.py
escapewindow: Hm. I tried this, and it went back to using `default` for every revision. I haven't dug into why. I also have some vague memory there may be a `format` elsewhere that the `%()s` strings avoid.

Tl;dr, I'm probably going to go back to %s because that was working =\

PR #257, opened 2020-08-26
Bump all dependencies

bhearsum:


escapewindow (approved):
Thanks! This was via `pin.sh`, yes?

bhearsum:
> Thanks! This was via `pin.sh`, yes?

uhhh, nope! I forgot that was a thing -- I ran `pip-compile-multi` in each directory. Looks like pin.sh is coming up with slightly different results (mostly in comments, but also it doesn't include `setuptools` in the .txt files). I'll rerun fully with that and update.

bhearsum:
This worked fine in dev. I'm going to merge this with the intent to deploy to production on Monday.

PR #259, opened 2020-08-31
Treescript: Use config["upstream_repo"] instead of hardcoded URL in d…

jfx2006:
…o_merge

This is needed for Thunderbird support. config["upstream_repo"] is already set appropriately and is used in checkout_repo.

### DIFF
--- a/treescript/tests/test_merges.py
+++ b/treescript/tests/test_merges.py
@@ -297,3 +297,16 @@ async def test_bump_central(mocker, config, task, repo_context, merge_bump_info)
     for expected in expected_calls:
         assert expected in called_args
     assert result == [("https://hg.mozilla.org/mozilla-central", "some_revision")]
+
+
+@pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


treescript/tests/test_merges.py
```diff
@@ -296,16 +297,3 @@ async def noop_apply_rebranding(*arguments, **kwargs):
     for expected in expected_calls:
         assert expected in called_args
     assert result == [("https://hg.mozilla.org/mozilla-central", "some_revision")]
-
-
-@pytest.mark.parametrize(
```
escapewindow: Were these removals intentional?

jfx2006: Nope.

escapewindow (approved):

escapewindow:
What timeframe do you want this to be live?

jfx2006:
> What timeframe do you want this to be live?
Early this week would be good.

PR #260, opened 2020-09-01
bug 1444406: always use https links for firefox mars

bhearsum:
This may need an update if Thunderbird wants to use https.


escapewindow (approved):
Looks like the test bustage is a black format change.

PR #265, opened 2020-09-17
Bug 1663453 - create Firefox versions on AMO before uploading langpacks

MihaiTabara:


addonscript/src/addonscript/xpi.py
```diff
@@ -1,10 +1,29 @@
...
 
 import json
+import re
 from zipfile import ZipFile
 
+from addonscript.exceptions import BadVersionError
 
-def get_langpack_info(context, path):
+
+def get_stripped_version(version):
```
escapewindow, line 10: @JohanLorenzo Does MozillaVersion have a better solution here? (Even if so, this could be a followup).

addonscript/src/addonscript/script.py
```diff
@@ -64,6 +75,11 @@ async def async_main(context):
     async with aiohttp.ClientSession(connector=connector) as session:
         context.session = session
         build_locales_context(context)
+
+        # ensure the versions exists on AMO before proceeding with uploading langpacks
+        for locale in context.locales:
```
escapewindow: We may want to build a set of versions, and only call the API once per version. I'm guessing that we'll have the exact same version for each xpi in a normal beta, so we might create the same version 5 times for l10n.

escapewindow (approved):
Nice! Assuming this works, 🚢 

Not sure how we want to test. We could look into whether AMO stage is a good test on try, or if we should wait for a quiet day for betas and releases, and land before the nightly.

Thank you!

addonscript/src/addonscript/api.py
```diff
@@ -14,6 +14,34 @@
...
+
+    Raises:
+        BadVersionError: If the XPI's version fails to sanity check against AMO
+
+
+    """
+    url = get_api_url(context, ADD_VERSION, version=version)
+
+    try:
+        await amo_put(context, url, data=None)
```
MihaiTabara: @escapewindow I wasn't sure whether this is fine or not. I would've loved to log better in case of 201/202 to better debug later on if we have to. But `amo_put` strips the status code and returns the json directly. I can either duplicate that function and process the status code or keep as is for now.

addonscript/src/addonscript/script.py
MihaiTabara: Yeah, I was going to ask in the PR about this but you were faster :) I checked a couple of logs for a couple of releases and the version is actually the same for all. So yeah, I think we can still build the dict since that's an easy cheap operation and only perform the API call once.

addonscript/tests/manifest.json
```diff
@@ -0,0 +1,19 @@
+{
```
MihaiTabara: Note to self: need to kill this file, I accidentally uploaded it.

MihaiTabara:
> Nice! Assuming this works, 🚢
> 
> Not sure how we want to test. We could look into whether AMO stage is a good test on try, or if we should wait for a quiet day for betas and releases, and land before the nightly.
> 
> Thank you!

Yeah, was playing with `./mach try scriptworker amo --release-type beta --worker-override amo=gecko-1-addon-dev` last night, adjusting the try selector. Not sure it'll be picked up in staging release though. Will give this a try again this morning. https://treeherder.mozilla.org/#/jobs?repo=try&revision=a7e7f3fae5bb4b5ec9b2e514ab0627b92982ed3e

addonscript/src/addonscript/utils.py
```diff
@@ -75,7 +75,8 @@ async def amo_put(context, url, data):
         async with resp as r:
             log.debug('amo_put() for URL "{}" returned HTTP status code: {}'.format(url, r.status))
             r.raise_for_status()
-            returned_value = await r.json()
+            # we silence aiohttp in case we have Null returns from AMO API
+            returned_value = await r.json(content_type=None)
```
MihaiTabara, line 79: To silence errors like for None returns from AMO API. Not sure if there's a better workaround this.

From [here](https://firefox-ci-tc.services.mozilla.com/tasks/Z7idDlScRbGQEpp_zK1S9w/runs/1/logs/https%3A%2F%2Ffirefox-ci-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FZ7idDlScRbGQEpp_zK1S9w%2Fruns%2F1%2Fartifacts%2Fpublic%2Flogs%2Flive_backing.log)
```
    raise ContentTypeError(
aiohttp.client_exceptions.ContentTypeError: 0, message='Attempt to decode JSON with unexpected mimetype: ', url=URL('https://addons.allizom.org/api/v4/applications/firefox/82.0/')
exit code: 1
```

MihaiTabara:
Green [langpack task](https://firefox-ci-tc.services.mozilla.com/tasks/Z7idDlScRbGQEpp_zK1S9w/runs/2#artifacts) to AMO!
This is now ready for review. I'm still waiting for :mat to answer in the bug to see the best time to land this but I'm assuming early 82 beta cycle.

addonscript/README.md
```diff
@@ -5,9 +5,10 @@ Addonscript
 
 This is designed to be run from scriptworker, but runs perfectly fine as a standalone script.
 
-
 Docs
 ----
+More details on what this script does can be found in [RelEng docs](http://docs.mozilla-releng.net/en/latest/addons/addons.html).
```
escapewindow: I moved this to http://docs.mozilla-releng.net/en/latest/addons/langpacks.html aka https://moz-releng-docs.readthedocs.io/en/latest/addons/langpacks.html in https://github.com/mozilla/build-relengdocs/pull/51  :)

addonscript/src/addonscript/api.py
```diff
@@ -14,6 +14,36 @@
...
+    url = get_api_url(context, ADD_VERSION, version=version)
+
+    try:
+        result = await amo_put(context, url, data=None)
+    except ClientResponseError as exc:
+        if exc.status == 401:
+            raise AuthFailedError("Addonscript credentials are misconfigured")
+        elif exc.status == 403:
+            raise AuthInsufficientPermissionsError("Addonscript creds are missing permissions")
+        raise exc
```
escapewindow, line 43: In the future, we could log if the version was created or already existed by the 20x response. As you mentioned, this would require different behavior from `amo_put` or a separate function.

addonscript/src/addonscript/api.py
```diff
@@ -14,6 +14,36 @@
...
+ADD_VERSION = "api/v4/applications/firefox/{version}/"
+
+
+async def add_version(context, version):
+    """Add a new version to AMO.
+
+    Use the `min_version` here, rather than the string with the buildid etc.
+
+    Raises:
+        BadVersionError: If the XPI's version fails to sanity check against AMO
```
escapewindow: Hm, this isn't accurate. This should be `AuthFailedError` and `AuthInsufficientPermissionsError`.

escapewindow (approved):
Thanks!

addonscript/src/addonscript/api.py
MihaiTabara: Good catch, fixed!

addonscript/src/addonscript/api.py
MihaiTabara: 👍

addonscript/README.md
MihaiTabara: Good catch, I've updated it! :)

PR #270, opened 2020-09-24
Bug 1667231 - cleanup application-services scopes. r=#releng

MihaiTabara:


escapewindow (approved):
Looks like we have some test bustage? Guessing we can switch those tests to `app-services`.

Issue #236, opened 2020-06-16
fix docker tox


escapewindow:
The problem is we use `requirements/local.txt` for both py37 and py38. We need to use the python-specific local requirements file.

This will probably be easiest if we put the PYTHON_VERSION in both filenames, so we can `RUN pip install -r requirements/local.$PYTHON_VERSION.txt` or similar in the Dockerfile.

MihaiTabara:
@alopezz fixed this in https://github.com/mozilla-releng/scriptworker-scripts/pull/271, I think we can close this. Thank you for your help, Alex!

PR #271, opened 2020-09-28
Fix tox + Dockerfile.test for py37

alopezz:
I found out that running `tox` as per the instruction in the Readme of beetmoverscript failed because of unpinned dependencies. Turns out the problem was that the wrong requirements file was being used for running the tests in Docker for py37.

@JohanLorenzo I applied the fix to beetmoverscript in this first commit. If it looks good, I'll apply the same to all the others.

Edit: I just found out that this had already been reported as issue #236.


beetmoverscript/Dockerfile.test
```diff
@@ -7,7 +7,8 @@ WORKDIR /app
 COPY MANIFEST.in setup.py tox.ini /app/
 COPY requirements/ /app/requirements/
 
-RUN pip install -r requirements/local.txt
+ARG PYTHON_REQ_SUFFIX
```
JohanLorenzo, line 10: Nit: We can define a default value so that we don't have to pass `--build-arg PYTHON_REQ_SUFFIX=      ` on python 3.8. 

```suggestion
ARG PYTHON_REQ_SUFFIX=''
```

beetmoverscript/tox.ini
```diff
@@ -8,10 +8,10 @@ usedevelop = false
 depends =
 skip_install = true
 commands =
-    docker build --build-arg PYTHON_VERSION=3.7 -t beetmoverscript-{envname}-py37-test -f Dockerfile.test .
-    docker run --rm -v {toxinidir}:/app -v beetmoverscript-{envname}-py37-tox:/app/.tox beetmoverscript-{envname}-py37-test check,py37
-    docker build --build-arg PYTHON_VERSION=3.8 -t beetmoverscript-{envname}-py38-test -f Dockerfile.test .
-    docker run --rm -v {toxinidir}:/app -v beetmoverscript-{envname}-py38-tox:/app/.tox beetmoverscript-{envname}-py38-test py38
+    docker build --build-arg PYTHON_VERSION=3.7 --build-arg PYTHON_REQ_SUFFIX=.py37 -t beetmoverscript-{envname}-py37-test -f Dockerfile.test .
```
JohanLorenzo, line 11: 👍

JohanLorenzo (commented):
Looks great to me! I tested it on my machine and it works like a charm[1]. Thank you for spotting this issue and for suggesting to fix it on all subprojects. I'm happy to r+ this patch once they're handle. There's a small nit which I think can improve the py38 command line.

[1] I hit https://github.com/docker/for-mac/issues/4919 along the way, but that's unrelated to this PR nor to this project.

JohanLorenzo:
Hmmm, the `treescript` error looks unrelated to this PR. Investigating. 

Edit: Last time it passed was 19 days ago on https://github.com/mozilla-releng/scriptworker-scripts/commit/7fe73de6846fb6acc6e23376dc58fa9d8834c29f, https://firefox-ci-tc.services.mozilla.com/tasks/OxCzr8vbR6mjTgPwKzjrkQ

Later edit: This is going to be fixed by https://github.com/mozilla-releng/scriptworker-scripts/pull/273

beetmoverscript/Dockerfile.test
JohanLorenzo: Oh looks like `ARG PYTHON_REQ_SUFFIX` already sets it to an empty string. TIL!

JohanLorenzo (approved):
Looks great to me! Thank you again for discovering this issue and more importantly, for fixing it! 👍

PR #276, opened 2020-09-30
Githubscript should not alter target_commitish on a GitHub release

JohanLorenzo:
Otherwise it breaks Chain of Trust.

Fixes

```
2020-09-29T17:57:51    ERROR - Hit ScriptWorkerException: CoTError('scriptworker:parent el3i3MuyR_CCWVw71AwNdA: the runtime task doesn\'t match any rebuilt definition!\n["[(\'change\',\\n"\n "  \'payload.env.MOBILE_HEAD_REF\',\\n"\n "  (\'88fc5f8a7e1b51e14d1b68f692a629d9144e8e5c\', \'releases/v82.0.0\'))]"]')
```

https://firefox-ci-tc.services.mozilla.com/tasks/Y3nWYC5WRsW0Zf0DYz-E7A/runs/0/logs/https%3A%2F%2Ffirefox-ci-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FY3nWYC5WRsW0Zf0DYz-E7A%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Fchain_of_trust.log#L186

I looked into making Chain of Trust more resilient but we're limited by the Github API. Let me walk you through 2 examples:

Fenix v81.1.1 was shipped via Github releases and never touched by githubscript. See the field `target_commitish`: https://api.github.com/repos/mozilla-mobile/fenix/releases/31503131. On the other hand, Fenix v82.0.0b1 was modified by githubscript: https://api.github.com/repos/mozilla-mobile/fenix/releases/31683165. On the first example, the `target_commitish` is a git branch whereas it's a commit has in the second example.

I could change CoT to get the branch from a hash, but that's not doable. Moreover, a single hash can be on multiple branches, so we have no reliable way of getting the branch of v82b1. The same happens with the event used by `.taskcluster.yml`, you either get a hash or a branch but not both https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#release.

Long story short: the simplest solution is to stop changing the `target_commitish`. This way, the CoT uses the same data all over the graph. That's why I made this very PR.

The `target_commitish` will be passed to Github whenever we fully switch to shipit. Reason is: no github release is created until githubscript starts, so we somehow need to give something (here a hash) to Github in order to display that release.

We will be able to revert https://github.com/mozilla-mobile/fenix/pull/15546, once that change is live.

Let me know if you have any questions!

### DIFF
--- a/githubscript/src/githubscript/github.py
+++ b/githubscript/src/githubscript/github.py
@@ -1,8 +1,9 @@
 import logging
 from asyncio import ensure_future
 
-import github3
 from aiohttp_retry import RetryClient
+from github3 import GitHub
+from github3.exceptions import NotFoundError, ServerError
 from scriptworker_client.exceptions import TaskError
 from scriptworker_client.utils import async_wrap, get_single_item_from_sequence, raise_future_exceptions, retry_async_decorator
 
@@ -25,7 +26,7 @@ async def release(release_config):
         existing_release = await _get_release_from_tag(github_repository, git_tag)
         log.info(f"Release {release_name} already exists. Making sure it has the latest data...")
         await _update_release_if_needed(existing_release, release_config)
-    except github3.exceptions.NotFoundError:
+    except NotFoundError:
         log.info(f"Release {release_name} does not exist. Creating it...")
         await _create_release(github_repository, release_config)
 
@@ -43,12 +44,12 @@ async def release(release_config):
 # A delay factor of 7.5s means the second round of waiting will occur ~15s after the first one,
 # the third one ~30s and so on.
 _GITHUB_LIBRARY_SLEEP_TIME_KWARGS = {"delay_factor": 7.5}
-github_retry = retry_async_decorator(retry_exceptions=github3.exceptions.ServerError, sleeptime_kwargs=_GITHUB_LIBRARY_SLEEP_TIME_KWARGS)
+github_retry = retry_async_decorator(retry_exceptions=ServerError, sleeptime_kwargs=_GITHUB_LIBRARY_SLEEP_TIME_KWARGS)
 
 
 @github_retry
 async def _init_github_client(token):
-    async_github_constructor = async_wrap(github3.GitHub)
+    async_github_constructor = async_wrap(GitHub)
     return await async_github_constructor(token=token)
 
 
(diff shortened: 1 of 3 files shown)


escapewindow:
> I could change CoT to get the branch from a hash, but that's not doable. Moreover, a single hash can be on multiple branches, so we have no reliable way of getting the branch of v82b1. The same happens with the event used by `.taskcluster.yml`, you either get a hash or a branch but not both https://docs.github.com/en/free-pro-team@latest/developers/webhooks-and-events/webhook-events-and-payloads#release.

I was actually thinking about this from the other direction. Rather than try to get the branch from a hash, can we get the hash from a branch or tag? Basically, the revision can never change, barring a hash collision. A tag can move, and a branch can have additional revisions pushed to it. I'm wondering if we should expand to the revision on both sides so they match?

(Then again, branch/tag info may help in determining which scopes we should check for, if we allow for releases off master/main and a relbranch but not PR branches. Not sure.)

> Long story short: the simplest solution is to stop changing the `target_commitish`. This way, the CoT uses the same data all over the graph. That's why I made this very PR.
> 
> The `target_commitish` will be passed to Github whenever we fully switch to shipit. Reason is: no github release is created until githubscript starts, so we somehow need to give something (here a hash) to Github in order to display that release.

Sounds like this class of problem will go away when we move to shipit?
If so, this sounds like a good stopgap.

escapewindow (approved):
Thanks! Approved, depending on the answers to the questions in https://github.com/mozilla-releng/scriptworker-scripts/pull/276#issuecomment-701591445 .

JohanLorenzo:
> I was actually thinking about this from the other direction. Rather than try to get the branch from a hash, can we get the hash from a branch or tag?

@escapewindow and I quickly chatted on chat.m.o. If I remember correctly, expanding branches and tags is already done in taskgraph. That's why we have `MOBILE_HEAD_REV` and `MOBILE_HEAD_REF`. So expanding them in CoT wouldn't solve the issue we hit this week. I'm going to keep this solution and merge it. I'll deploy it on Monday.

PR #277, opened 2020-09-30
Bug 1600916 - Check maven artifact version

alopezz:
Addressing [Bug 1600916](https://bugzilla.mozilla.org/show_bug.cgi?id=1600916).

### DIFF
--- a/beetmoverscript/src/beetmoverscript/script.py
+++ b/beetmoverscript/src/beetmoverscript/script.py
@@ -13,6 +13,8 @@ import aiohttp
 import boto3
 from botocore.exceptions import ClientError
 from redo import retry
+from mozilla_version.maven import MavenVersion
+from mozilla_version.errors import PatternNotMatchedError
 from scriptworker import client
 from scriptworker.exceptions import ScriptWorkerRetryException, ScriptWorkerTaskException
 from scriptworker.utils import raise_future_exceptions, retry_async
@@ -198,9 +200,27 @@ async def push_to_maven(context):
     context.checksums = dict()  # Needed by downstream calls
     context.raw_balrog_manifest = dict()  # Needed by downstream calls
 
+    check_maven_artifact_map(context)
+
     # overwrite artifacts_to_beetmove with the declarative artifacts ones
     context.artifacts_to_beetmove = task.get_upstream_artifacts(context, preserve_full_paths=True)
     await move_beets(context, context.artifacts_to_beetmove, artifact_map=context.task["payload"]["artifactMap"])
+
+
+def check_maven_artifact_map(context):
+    """Check that versions in artifact map to be valid Maven versions"""
+    version = context.task["payload"]["version"]
+    try:
+        MavenVersion.parse(version)
+    except PatternNotMatchedError:
+        raise ScriptWorkerTaskException(f"bad version '{version}'")
+
+    for artifact_dict in context.task["payload"]["artifactMap"]:
+        for dest_dict in artifact_dict["paths"].values():
+            for dest in dest_dict["destinations"]:
+                dest_file = os.path.basename(dest)
+                if version not in dest_file:
+                    raise ScriptWorkerTaskException(f"version in artifact doesn't match expected version '{version}'")
 
 
 # copy_beets {{{1
--- a/beetmoverscript/src/beetmoverscript/script.py
+++ b/beetmoverscript/src/beetmoverscript/script.py
@@ -12,8 +12,6 @@ from multiprocessing.pool import ThreadPool
 import aiohttp
 import boto3
 from botocore.exceptions import ClientError
-from mozilla_version.errors import PatternNotMatchedError
-from mozilla_version.maven import MavenVersion
 from redo import retry
 from scriptworker import client
 from scriptworker.exceptions import ScriptWorkerRetryException, ScriptWorkerTaskException
@@ -200,27 +198,11 @@ async def push_to_maven(context):
     context.checksums = dict()  # Needed by downstream calls
     context.raw_balrog_manifest = dict()  # Needed by downstream calls
 
-    check_maven_artifact_map(context)
+    task.check_maven_artifact_map(context)
 
     # overwrite artifacts_to_beetmove with the declarative artifacts ones
     context.artifacts_to_beetmove = task.get_upstream_artifacts(context, preserve_full_paths=True)
     await move_beets(context, context.artifacts_to_beetmove, artifact_map=context.task["payload"]["artifactMap"])
-
-
-def check_maven_artifact_map(context):
-    """Check that versions in artifact map to be valid Maven versions"""
-    version = context.task["payload"]["version"]
-    try:
-        MavenVersion.parse(version)
-    except PatternNotMatchedError:
-        raise ScriptWorkerTaskException(f"bad version '{version}'")
-
-    for artifact_dict in context.task["payload"]["artifactMap"]:
-        for dest_dict in artifact_dict["paths"].values():
-            for dest in dest_dict["destinations"]:
-                dest_file = os.path.basename(dest)
-                if version not in dest_file:
-                    raise ScriptWorkerTaskException(f"version in artifact doesn't match expected version '{version}'")
 
 
 # copy_beets {{{1
--- a/beetmoverscript/src/beetmoverscript/script.py
+++ b/beetmoverscript/src/beetmoverscript/script.py
@@ -12,9 +12,9 @@ from multiprocessing.pool import ThreadPool
 import aiohttp
 import boto3
 from botocore.exceptions import ClientError
-from redo import retry
-from mozilla_version.maven import MavenVersion
 from mozilla_version.errors import PatternNotMatchedError
+from mozilla_version.maven import MavenVersion
+from redo import retry
 from scriptworker import client
 from scriptworker.exceptions import ScriptWorkerRetryException, ScriptWorkerTaskException
 from scriptworker.utils import raise_future_exceptions, retry_async
--- a/beetmoverscript/tests/test_script.py
+++ b/beetmoverscript/tests/test_script.py
@@ -13,7 +13,6 @@ import beetmoverscript.script
 from beetmoverscript.constants import PARTNER_REPACK_PRIVATE_REGEXES, PARTNER_REPACK_PUBLIC_REGEXES
 from beetmoverscript.script import (
     async_main,
-    check_maven_artifact_map,
     copy_beets,
     enrich_balrog_manifest,
     get_destination_for_partner_repack_path,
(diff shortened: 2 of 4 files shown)


beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -198,11 +200,29 @@ async def push_to_maven(context):
...
     context.raw_balrog_manifest = dict()  # Needed by downstream calls
 
+    check_maven_artifact_map(context)
+
     # overwrite artifacts_to_beetmove with the declarative artifacts ones
     context.artifacts_to_beetmove = task.get_upstream_artifacts(context, preserve_full_paths=True)
     await move_beets(context, context.artifacts_to_beetmove, artifact_map=context.task["payload"]["artifactMap"])
 
 
+def check_maven_artifact_map(context):
```
JohanLorenzo: Nit: Let's move this function in `beetmoverscript/src/beetmoverscript/task.py`. That's where most of the task validation happens.

beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -198,11 +200,29 @@ async def push_to_maven(context):
...
+    """Check that versions in artifact map to be valid Maven versions"""
+    version = context.task["payload"]["version"]
+    try:
+        MavenVersion.parse(version)
+    except PatternNotMatchedError:
```
JohanLorenzo: Good idea on matching a specific error. One thing I realize: `mozilla_version` raises another type of error: https://github.com/mozilla-releng/mozilla-version/blob/b541228cf29bd8286e0fbe9692a82bd39bb94804/mozilla_version/version.py#L142. It's not ideal because people unfamiliar with `mozilla-version` may except just a single exception - just like you did.

For now, let's use 

```suggestion
    except (ValueError, PatternNotMatchedError) as e:
```

but I believe your solution is what we want to have, in the end.

beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -198,11 +200,29 @@ async def push_to_maven(context):
...
+        raise ScriptWorkerTaskException(f"bad version '{version}'")
```
JohanLorenzo: Nit: I usually like when error messages are super explicit. It helps to know what happened whenever the exception is raised in prod and you have no easy way to reproduce the problem. Here, I think we can extend the message to something like: 

```suggestion
        raise ScriptWorkerTaskException(f"Version defined in the payload does not match the pattern of a MavenVersion. Got: {version}")
```

We can even go one step further by re-raising the original error. This way, we know what didn't MavenVersion didn't like.

```suggestion
        raise ScriptWorkerTaskException(f"Version defined in the payload does not match the pattern of a MavenVersion. Got: {version}") from e
```

What do you think?

beetmoverscript/tests/test_script.py
```diff
@@ -69,6 +70,45 @@ def fake_list(*args):
...
+    }
+
+    with pytest.raises(ScriptWorkerTaskException):
+        check_maven_artifact_map(context)
+
+    context.task["payload"]["version"] = fake_correct_version
+    context.task["payload"]["artifactMap"] = [artifact_map_entry_with_version(fake_correct_version)]
+    check_maven_artifact_map(context)
+
+    context.task["payload"]["artifactMap"] = [artifact_map_entry_with_version("a.bad.version")]
```
JohanLorenzo: Sweet! I love the fact you tested a failure case! 👍 This test looks good the way it is now. If you'd like to enhance it, you can leverage `@pytest.mark.parametrize` https://docs.pytest.org/en/latest/example/parametrize.html#parametrizing-conditional-raising. So this test could be split into 3 different cases. This will make the logic smaller and this will ensure that each case runs even though one fails. For instance: if we break the first `check_maven_artifact_map()` call, then the next ones won't be tested.

That's up to you. I don't mind merging this test as-is.

beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -198,11 +200,29 @@ async def push_to_maven(context):
...
+
+    for artifact_dict in context.task["payload"]["artifactMap"]:
+        for dest_dict in artifact_dict["paths"].values():
+            for dest in dest_dict["destinations"]:
+                dest_file = os.path.basename(dest)
```
JohanLorenzo: Good call on taking the base name out! I didn't of it, and that's smart! 

We can add one more check: The last folder is always the version, so we can take it out of `dest` and make sure it's strictly matching the version.

beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -198,11 +200,29 @@ async def push_to_maven(context):
...
+                if version not in dest_file:
+                    raise ScriptWorkerTaskException(f"version in artifact doesn't match expected version '{version}'")
```
JohanLorenzo: Nit: Similarly to the previous exception, let's expand the error message here: 

```suggestion
                    raise ScriptWorkerTaskException(f"Cannot find version '{version}' in file name '{dest_file}'. Path under test: {dest}")
```

JohanLorenzo (commented):
Thanks a lot for this PR. You provided a great patch! There a few nits here and there, plus a few improvements we can make. Let's get the nits fixed before merging. The improvements are up to you. Let me know what you think!

beetmoverscript/src/beetmoverscript/script.py
alopezz: OK. That was also my first thought, but since I wasn't sure of the criteria I ended up putting it in the same file. We'll move it there.

beetmoverscript/src/beetmoverscript/script.py
alopezz: Great! I had never used the `raise from` statement even though I knew it existed.

beetmoverscript/tests/test_script.py
alopezz: OK. I didn't know you could parametrize when exceptions were involved. I think I'll do what you suggest.

beetmoverscript/src/beetmoverscript/script.py
alopezz: I didn't want to make too many assumptions, but it makes sense so I'll add it as well.

beetmoverscript/src/beetmoverscript/script.py
alopezz: After I applied your suggested change I realized that the `parse` method we are calling (https://github.com/mozilla-releng/mozilla-version/blob/b541228cf29bd8286e0fbe9692a82bd39bb94804/mozilla_version/version.py#L36) seems to only raise `PatternNotMatchedError` exception, or `MissingFieldError` due to `get_value_matched_by_regex`. I see that all the errors raised by `mozilla-version` are subclasses of `ValueError` anyway, so catching `ValueError` works too (and makes catching `PatternNotMatchedError` redundant, I think). 

Maybe it would be interesting to add something like a `ParsingError` exception class to `mozilla-version` that all related errors would inherit from? Such an exception is probably what I would expect as a consumer of the library, and it's a backwards-compatible change.

beetmoverscript/tests/test_task.py
```diff
@@ -150,6 +152,49 @@ def test_validate_bucket_paths(bucket, path, raises):
...
+        (None, None, "a.bad.version", pytest.raises(ScriptWorkerTaskException)),
+    ),
+)
+def test_check_maven_artifact_map(context, payload_version, filename_version, folder_version, expectation):
+    context.action = "push-to-maven"
+
+    fake_correct_version = "12.3.20200920201111"
+    source = "fake_path/fake-artifact-{version}.jar"
+
+    payload_version = payload_version or fake_correct_version
```
JohanLorenzo: Nit: Let's replace the `None`s in parametrize with `"12.3.20200920201111"`. It will make the test even smaller and it will be easier to understand the test case, because the next reader won't have to mentally map `None` to the real value when this person reads lines 158-161.

JohanLorenzo (commented):
That pull request looks wonderful! Thanks a lot for making the code look nicer. Let's fix the last small nit and we'll be done with reviews.

The next step would be to add task.payload in https://github.com/mozilla-mobile/android-components/blob/a8e6bc071d1ffcfdc94db6884df0441c5ae7f1b7/taskcluster/ac_taskgraph/transforms/beetmover.py#L43, just like we have it there https://github.com/mozilla/application-services/blob/3fc42b212b12493e4113ed787cfa985b0b269a24/taskcluster/app_services_taskgraph/transforms/beetmover.py#L70

Would you be interested in doing so @alopezz? If so, I can give you more details!

beetmoverscript/tests/test_task.py
alopezz: OK. I was trying to make it obvious that those values have to be the same and avoiding repeating the string literal by using the None's as a replacement for using a variable, since I didn't want to create a top-level variable for this single test neither.

alopezz:
> The next step would be to add task.payload in https://github.com/mozilla-mobile/android-components/blob/a8e6bc071d1ffcfdc94db6884df0441c5ae7f1b7/taskcluster/ac_taskgraph/transforms/beetmover.py#L43, just like we have it there https://github.com/mozilla/application-services/blob/3fc42b212b12493e4113ed787cfa985b0b269a24/taskcluster/app_services_taskgraph/transforms/beetmover.py#L70
> 
> Would you be interested in doing so @alopezz? If so, I can give you more details!

Sure!

alopezz:
> The next step would be to add task.payload in https://github.com/mozilla-mobile/android-components/blob/a8e6bc071d1ffcfdc94db6884df0441c5ae7f1b7/taskcluster/ac_taskgraph/transforms/beetmover.py#L43, just like we have it there https://github.com/mozilla/application-services/blob/3fc42b212b12493e4113ed787cfa985b0b269a24/taskcluster/app_services_taskgraph/transforms/beetmover.py#L70

Working on this: https://github.com/mozilla-mobile/android-components/pull/8654

JohanLorenzo (approved):
Excellent, @alopezz! This PR has tests and is nicely laid out! We tested it end-to-end. For more information see https://github.com/mozilla-mobile/android-components/pull/8654#pullrequestreview-508139522.

MihaiTabara:
Mozilla-central's geckoview nightlies are breaking [here](https://treeherder.mozilla.org/#/jobs?repo=mozilla-central&revision=f6615f173552511e63b87cd8db1ab1434e187630&selectedTaskRun=Do2XjXgxRha3QG8bkqZDlg.0) with
```
scriptworker.exceptions.ScriptWorkerTaskException: Version defined in the payload does not match the pattern of a MavenVersion. Got: 83.0a1
```

At first glance, I think this works nicely for AC but it might break for GV which are using GeckoVersion. Looking at https://maven.mozilla.org/?prefix=maven2/org/mozilla/ I realized we also publish https://github.com/mozilla/application-services/ and https://github.com/mozilla/glean/ releases via the same `push_maven` function. So we should check that `version` is present there too, otherwise we might get a ` KeyError` [here](https://github.com/mozilla-releng/scriptworker-scripts/pull/277/files#diff-6e73a05ce2b9813ab96827278b61d9e8debf8350179a9f270b1475009463868aR88).

MihaiTabara:
> Looking at https://maven.mozilla.org/?prefix=maven2/org/mozilla/ I realized we also publish https://github.com/mozilla/application-services/ and https://github.com/mozilla/glean/ releases via the same `push_maven` function. So we should check that `version` is present there too, otherwise we might get a ` KeyError` [here](https://github.com/mozilla-releng/scriptworker-scripts/pull/277/files#diff-6e73a05ce2b9813ab96827278b61d9e8debf8350179a9f270b1475009463868aR88).

Don't mind me on these, they work as expected, just tested locally and noticed 

```
>>> from mozilla_version.maven import MavenVersion
>>> MavenVersion.parse('33.0.4')
MavenVersion(major_number=33, minor_number=0, patch_number=4, is_snapshot=False)
>>> MavenVersion.parse('63.0.0')
MavenVersion(major_number=63, minor_number=0, patch_number=0, is_snapshot=False)
```

On the other hand, the GeckoView version fails though:
```
>>> MavenVersion.parse('83.0a1')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/mtabara/envs/scriptworker/lib/python3.7/site-packages/mozilla_version/maven.py", line 27, in parse
    return super(MavenVersion, cls).parse(version_string, regex_groups=('is_snapshot', ))
  File "/Users/mtabara/envs/scriptworker/lib/python3.7/site-packages/mozilla_version/version.py", line 41, in parse
    raise PatternNotMatchedError(version_string, (cls._VALID_ENOUGH_VERSION_PATTERN,))
mozilla_version.errors.PatternNotMatchedError: "83.0a1" does not match the pattern: re.compile('\n        ^(?P<major_number>\\d+)\n        \\.(?P<minor_number>\\d+)\n        (\\.(?P<patch_number>\\d+))?\n        (?P<is_snapshot>-SNAPSHOT)?$', re.VERBOSE)

```

MihaiTabara:
CC: @JohanLorenzo @alopezz 
To unblock GV nightlies, as per Slack's #releaseduty-mobile, I've reverted beetmover workers to `7fe73de6` (HEAD of `production`) until we address the fix here. I think we're fine not reverting further https://github.com/mozilla-mobile/android-components/pull/8654 since that's just enriching payload version 👍

MihaiTabara:
Mea culpa, I should've seen this coming as I burned myself the same way a while ago. I've added https://github.com/mozilla-releng/scriptworker-scripts/pull/279 to try to make the docstrings clearer to prevent us from footguns in the future.

PR #279, opened 2020-10-14
No bug - update docstrings to prevent footguns.

MihaiTabara:
We burned ourselves a couple of times in the last few months in handling maven relateed issues. Not sure what's the best way to prevent those footguns for now, but at the very least we can add some better comments for now.


JohanLorenzo (approved):
👍

PR #280, opened 2020-10-15
Bug 1600916 - Check versions in payload and maven artifact maps

alopezz:
This tries to fix the problems that #277 introduced when dealing with geckoview versions.

The plan is to also modify how the version is set in the payload for geckoview beetmover jobs. However, we still have to support the current geckoview version format here as a special case until the changes we make in mozilla-central get to beta and release. 

@JohanLorenzo I split the logic into two functions so that it's easier to unit-test, but I may have gone overboard with switching on the specific major version, so I'd understand if you find this is not a satisfactory solution.

### DIFF
--- a/beetmoverscript/src/beetmoverscript/script.py
+++ b/beetmoverscript/src/beetmoverscript/script.py
@@ -13,6 +13,8 @@ import aiohttp
 import boto3
 from botocore.exceptions import ClientError
 from redo import retry
+from mozilla_version.maven import MavenVersion
+from mozilla_version.errors import PatternNotMatchedError
 from scriptworker import client
 from scriptworker.exceptions import ScriptWorkerRetryException, ScriptWorkerTaskException
 from scriptworker.utils import raise_future_exceptions, retry_async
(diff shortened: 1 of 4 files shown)


beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -208,7 +208,9 @@ async def push_to_maven(context):
     context.checksums = dict()  # Needed by downstream calls
     context.raw_balrog_manifest = dict()  # Needed by downstream calls
 
-    task.check_maven_artifact_map(context)
+    # Version validation
+    version = task.check_maven_version(context)
```
JohanLorenzo: Nit: `task.get_maven_version`

beetmoverscript/src/beetmoverscript/task.py
```diff
@@ -83,14 +84,28 @@ def validate_bucket_paths(bucket, s3_bucket_path):
         raise ScriptWorkerTaskException("Forbidden S3 {} destination".format(s3_bucket_path))
 
 
-def check_maven_artifact_map(context):
-    """Check that versions in artifact map are valid Maven versions"""
+def get_maven_version(context):
+    """Extract and validate a valid Maven version"""
     version = context.task["payload"]["version"]
+    release_props = context.release_props
+    # Version follows the FirefoxVersion pattern for Geckoview < 84.0
```
JohanLorenzo: Nit: Add a TODO section to remind us to remove that code path once Geckoview 84 is in mozilla-release

JohanLorenzo (approved):
Looks good to me overall. I managed to get a green task with the nits fixed. My patch is https://github.com/mozilla-releng/scriptworker-scripts/commit/5dc6646de1d213d5650e70e12718e673312b8710, if you want a closer look. Could you import it and I'll be please to merge this PR :)

beetmoverscript/src/beetmoverscript/script.py
alopezz: Whooops!

JohanLorenzo:
w00t! This patch looks great! I'm going to deploy it to production to see how it behaves with today's geckoview nightly. Looking forward to it!

PR #281, opened 2020-10-22
Bug 1672691 - Add integration tests to beetmover

alopezz:
https://bugzilla.mozilla.org/show_bug.cgi?id=1672691

It probably needs refinement regarding what assertions better reflect the expected behavior, as well as reducing the size of the example.

### DIFF
--- a/beetmoverscript/tests/test_integration.py
+++ b/beetmoverscript/tests/test_integration.py
@@ -1,0 +1,228 @@
+import json
+from contextlib import asynccontextmanager
+from unittest.mock import Mock
+
+import pytest
+
+import beetmoverscript.script
+from beetmoverscript.script import main
+
+from . import get_fake_valid_config
+
+
+@pytest.fixture
+def config():
+    config = get_fake_valid_config()
+    config["taskcluster_scope_prefix"] = "project:mobile:android-components:releng:beetmover:"
+    config["bucket_config"]["maven-nightly-staging"] = {"buckets": {"nightly_components": "dummy"}, "credentials": {"id": "dummy", "key": "dummy"}}
+    return config
+
+
+@pytest.fixture
+def task():
+    return {
+        "dependencies": ["dependency-task-id"],
+        "scopes": [
+            "project:mobile:android-components:releng:beetmover:action:push-to-maven",
+            "project:mobile:android-components:releng:beetmover:bucket:maven-nightly-staging",
+        ],
+        "payload": {
+            "version": "63.0.20201013153553",
+            "artifactMap": [
+                {"paths": {}, "locale": "en-US", "taskId": "AKjtHrqYQ32EDZ3GNzwXGg"},
+                {
+                    "paths": {
+                        "public/build/browser-awesomebar-63.0.20201013153553.aar.asc": {
+                            "destinations": [
+                                "maven2/org/mozilla/components/browser-awesomebar/63.0.20201013153553/browser-awesomebar-63.0.20201013153553.aar.asc"
+                            ],
+                            "checksums_path": "",
+                        },
+                        "public/build/browser-awesomebar-63.0.20201013153553.pom.asc": {
+                            "destinations": [
+                                "maven2/org/mozilla/components/browser-awesomebar/63.0.20201013153553/browser-awesomebar-63.0.20201013153553.pom.asc"
+                            ],
+                            "checksums_path": "",
+                        },
+                        "public/build/browser-awesomebar-63.0.20201013153553-sources.jar.asc": {
+                            "destinations": [
+                                "maven2/org/mozilla/components/browser-awesomebar/63.0.20201013153553/browser-awesomebar-63.0.20201013153553-sources.jar.asc"
+                            ],
+                            "checksums_path": "",
+                        },
+                    },
+                    "locale": "en-US",
+                    "taskId": "ZYfA8OtXRZuDeF5Xq3ubxg",
+                },
+                {
+                    "paths": {
+                        "public/build/browser-awesomebar-63.0.20201013153553.aar": {
+                            "destinations": ["maven2/org/mozilla/components/browser-awesomebar/63.0.20201013153553/browser-awesomebar-63.0.20201013153553.aar"],
+                            "checksums_path": "",
+                        },
+                        "public/build/browser-awesomebar-63.0.20201013153553.pom": {
+                            "destinations": ["maven2/org/mozilla/components/browser-awesomebar/63.0.20201013153553/browser-awesomebar-63.0.20201013153553.pom"],
+                            "checksums_path": "",
+                        },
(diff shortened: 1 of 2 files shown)


beetmoverscript/tests/test_integration.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+def create_dummy_artifacts_for_task(task, root_path):
+    """Create artifacts to upload as specified in the given task."""
+    for artifact_dict in task["payload"]["artifactMap"]:
+        task_id = artifact_dict["taskId"]
+        for path in artifact_dict["paths"]:
+            full_path = root_path / task_id / path
+            full_path.parent.mkdir(parents=True, exist_ok=True)
+            create_dummy_artifact(full_path)
```
JohanLorenzo, line 152: 👍

beetmoverscript/tests/test_integration.py
```diff
@@ -0,0 +1,228 @@
...
+        f.write(b"some data")
+
+
+@pytest.fixture
+def boto3_client_mock(monkeypatch):
+    client_mock = Mock(spec=beetmoverscript.script.boto3.client("s3"))
+    client_mock.generate_presigned_url.return_value = "presigned_url"
+
+    def fake_boto3_client_only_s3(*args, **kwargs):
+        assert args[0] == "s3", "A client other than 's3' was requested"
```
JohanLorenzo, line 182: Out of curiosity, what this assertion for? 🤔

beetmoverscript/tests/test_integration.py
```diff
@@ -0,0 +1,228 @@
...
+        return client_mock
+
+    monkeypatch.setattr(beetmoverscript.script.boto3, "client", fake_boto3_client_only_s3)
+
+    return client_mock
```
JohanLorenzo, line 187: TIL `@pytest.fixture` work with return statements too! I've always used `yield`.

beetmoverscript/tests/test_integration.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+def test_main_android_components_nightly(scriptworker_config, boto3_client_mock, aiohttp_session_mock):
+    main(config_path=scriptworker_config)
+
+    # Number of artifacts put matches expected
+    assert len(aiohttp_session_mock["put"]) == 12
+
+    # url to where artifacts are put is the presigned url
+    assert all(put_call["url"] == "presigned_url" for put_call in aiohttp_session_mock["put"])
```
JohanLorenzo: Good call on making sure we're using pre-signed URLs!

beetmoverscript/tests/test_integration.py
```diff
@@ -0,0 +1,228 @@
...
+    # The data put matches the data in the files
+    assert all(put_call["data"] == b"some data" for put_call in aiohttp_session_mock["put"])
```
JohanLorenzo: Nit: Let's add another assertion about the 12 paths. They should match the paths requested in the task payload. This way, if we end up breaking these paths (beetmover used to generate on the fly), then this test will let us know.

JohanLorenzo (approved):
This test is fantastic! It's small, well split between the setup, the run, and the assertions. The latter makes sense (I'd add another assertion, but I don't see many more to add). 

I love the work you've done! Could you address the few nits, and I'll be glad to get this patch merged! 👍

JohanLorenzo:
I'm asking Mihai's opinion. He has a better knowledge about what beetmoverscript. Maybe he'll see some things to improve. That said, the patch is really good the way it is now!

beetmoverscript/tests/test_integration.py
alopezz: Because of how the boto3 api works, where a generic [`boto3.client`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#client) function is called to get a client class for any service, a call to `boto3.client` returns a `S3.Client` object only when the first argument is `"s3"`. This fake function returns a S3 client mock based on the assumption that this is the only way the `boto3.client` function is called, and this assertion confirms that assumption. This is also to avoid having to mock every other service client that the function could potentially return.

beetmoverscript/tests/test_integration.py
alopezz: I learned the opposite when I read your code :smile:! Turns out that `yield` can be used to [allow for teardown code in a fixture](https://docs.pytest.org/en/stable/fixture.html#fixture-finalization-executing-teardown-code), so using `return` or `yield` is equivalent when it's the last statement of the fixture.

beetmoverscript/tests/test_integration.py
alopezz: :+1: I already had in mind that adding some assertion about the paths would be a good idea. However, should we check source paths, destination paths, or even both?

beetmoverscript/tests/test_script.py
```diff
@@ -125,6 +125,8 @@ def fake_filter(**kwargs):
 def test_setup_mimetypes():
     non_default_types = ["https://foo.com/fake_artifact.bundle", "http://www.bar.com/fake_checksum.beet"]
 
+    # ensure we start with mimetypes in its initial state
+    mimetypes.init()
```
MihaiTabara, line 129: This is useful since we're extending the list of mimetypes beyond the defualt one. Thanks for adding this!

beetmoverscript/tests/test_integration.py
```diff
@@ -0,0 +1,228 @@
...
+        },
+    }
+
+
+def create_dummy_artifacts_for_task(task, root_path):
+    """Create artifacts to upload as specified in the given task."""
+    for artifact_dict in task["payload"]["artifactMap"]:
+        task_id = artifact_dict["taskId"]
+        for path in artifact_dict["paths"]:
+            full_path = root_path / task_id / path
```
MihaiTabara, line 150: 👍 for using path manipulations

beetmoverscript/tests/test_integration.py
```diff
@@ -0,0 +1,228 @@
...
+    create_dummy_artifacts_for_task(task, work_dir / "cot")
+
+    return config_path
+
+
+def test_main_android_components_nightly(scriptworker_config, boto3_client_mock, aiohttp_session_mock):
+    main(config_path=scriptworker_config)
+
+    # Number of artifacts put matches expected
+    assert len(aiohttp_session_mock["put"]) == 12
```
MihaiTabara, line 239: +1 on adding this test!

MihaiTabara (approved):
As Johan said, this is fantastic work, thanks a lot for adding this test! I don't see anything that we might've missed so +1 to merging this. Thanks again for the great work!

JohanLorenzo (approved):
Amazing! This test is great! I'm really excited to get it merged 😃 

Thanks a lot for the marvelous work you've done again 👍

PR #285, opened 2020-10-28
Bug 1635488 - part 1: [treescript] Expose commit() function in mercurial

JohanLorenzo:
This refactor is now useful. I plan to create a `git` module that uses the same API as the `mercurial` one. There are several places where treescript commits something. By centralizing the command in the `mercurial`, we enable simpler modifications to wire calls either to the (upcoming) `git` module or the `mercurial` one.

This call is one of them:
https://github.com/mozilla-releng/scriptworker-scripts/blob/022acabfa8af4e928f27ca6f388d33127f2a4284/treescript/src/treescript/versionmanip.py#L94 

This refactor looks a little bit risky. I'd love to land it without anything related to git, yet.

### DIFF
--- a/treescript/src/treescript/l10n.py
+++ b/treescript/src/treescript/l10n.py
@@ -16,7 +16,7 @@ from scriptworker_client.aio import download_file, retry_async, semaphore_wrappe
 from scriptworker_client.exceptions import DownloadError
 from scriptworker_client.utils import load_json_or_yaml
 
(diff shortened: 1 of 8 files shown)


MihaiTabara (approved):
++ on exposing this in a generic way to support both `mercurial` and `git` modules' operations.

JohanLorenzo:
This has been deployed to production.

PR #288, opened 2020-10-30
Bug 1635488 - part 2 & 3: Dynamically choose vcs module

JohanLorenzo:
Follows https://github.com/mozilla-releng/scriptworker-scripts/pull/285 up. Unlike this PR, I didn't update `treescript/src/treescript/merges.py`, for 2 reasons:

 * That file is really tied to the way hg works. It calls `run_hg_command()` for instance.
 * We won't support git merges in bug 1635488. 

So I preferred to focus on the parts that will be used in the current bug.


treescript/src/treescript/script.py
```diff
@@ -51,31 +55,33 @@ async def do_actions(config, task, actions, repo_path):
...
+        await perform_merge_actions(config, task, actions, repo_path, repo_type)
         return
 
     num_changes = 0
     if "tag" in actions:
-        num_changes += await do_tagging(config, task, repo_path)
+        num_changes += await vcs.do_tagging(config, task, repo_path)
     if "version_bump" in actions:
-        num_changes += await bump_version(config, task, repo_path)
+        num_changes += await bump_version(config, task, repo_path, repo_type)
```
JohanLorenzo, line 72: `bump_version` and `l10n_bump` are in their own modules. That's why I'm passing `repo_type` down.

treescript/src/treescript/merges.py
```diff
@@ -205,13 +205,7 @@ async def do_merge(config, task, repo_path):
...
     if merge_config.get("merge_old_head", False):
         await run_hg_command(config, "debugsetparents", tagged_from_rev, base_to_rev, repo_path=repo_path)
-        await run_hg_command(
-            config,
-            "commit",
-            "-m",
-            "Merge old head via |hg debugsetparents {} {}| CLOSED TREE DONTBUILD a=release".format(tagged_from_rev, base_to_rev),
-            repo_path=repo_path,
-        )
+        await commit(config, repo_path, "Merge old head via |hg debugsetparents {} {}| CLOSED TREE DONTBUILD a=release".format(tagged_from_rev, base_to_rev))
```
MihaiTabara, line 208: ++ on this being generic now

treescript/src/treescript/script.py
```diff
@@ -51,31 +55,33 @@ async def do_actions(config, task, actions, repo_path):
         actions (list): the actions to perform
         repo_path (str): the source directory to use.
     """
-    await checkout_repo(config, task, repo_path)
+    repo_type = "hg"  # TODO parametrize when git support is added
```
MihaiTabara, line 58: 👍

treescript/tests/test_l10n.py
```diff
@@ -5,6 +5,13 @@
 from scriptworker_client.utils import makedirs
 
 import treescript.l10n as l10n
+from treescript import mercurial
```
MihaiTabara: Hm, do we need this import in this module?

treescript/tests/test_script.py
```diff
@@ -211,33 +197,29 @@ async def mocked_outgoing(*args):
...
-        return 1
+    vcs_mock = AsyncMock()
 
-    mocker.patch.object(script, "checkout_repo", new=noop_async)
+    mocker.patch.object(script, "get_vcs_module", return_value=vcs_mock)
     mocker.patch.object(script, "do_merge", new=mocked_do_merge)
-    mocker.patch.object(script, "push", new=mocked_push)
     mocker.patch.object(script, "should_push", return_value=should_push)
-    await script.perform_merge_actions({}, {}, actions, "/some/folder/here")
+    await script.perform_merge_actions({}, {}, actions, "/some/folder/here", "hg")
```
MihaiTabara, line 211: Not-blocking, I'm wondering whether it's worth adding the `git` case too to catch the `NotImplementedError` exception.

MihaiTabara: Later edit: don't mind me, I just noticed the `test_task.py` testing below which covers this case for that particular function.

MihaiTabara (approved):
LGTM! Great refactoring to allow for `git` injection later on.

treescript/tests/test_l10n.py
JohanLorenzo: We don't. Thanks for catching it!

PR #292, opened 2020-11-10
Bug 1672691 - part 4: [beetmoverscript] Add integration tests to beetmover (partner & releases)

alopezz:
Final part of this Bug.

These paths have more differences with the ones that were involved in the previous integration tests and thus required new mocks. 

- For `push-to-partner`, I simply relaxed the assertions needed to pass the test, as destinations are computed and not available in the task json. I preferred to do this rather than adding more logic to the test.
- For `push-to-releases`, I had to mock a few more parts of the `boto3` library. I tried to keep things as simple as possible while at least making sure that things work reasonably as expected. It's still probably the flakiest test so far.

CC'ing @JohanLorenzo as mentor for the bug and @MihaiTabara for possibly being more familiar with the functionality being tested.


JohanLorenzo (approved):
Wonderful work, as always! 💯 👍 Thank you so much for your help, this incredibly valuable to our team!

PR #293, opened 2020-11-10
bug 1676493: balrog toplevel secondary submission uses v1 api still

bhearsum:


escapewindow (approved):

escapewindow:
Stamped, pending tests. Thanks!

balrogscript/src/balrogscript/submitter/cli.py
```diff
@@ -154,12 +154,17 @@ def run(
...
-            blob["name"] = name
-            data = {"product": productName, "blob": blob}
             url = self.api_root + "/v2/releases/" + name
+            data = {"product": productName, "blob": {}}
             session = get_balrog_session(auth0_secrets=self.auth0_secrets)
+            current_data = balrog_request(session, "get", url)
+            if current_data:
+                data["blob"] = current_data["blob"]
+                data["old_data_versions"] = current_data["data_versions"]
+            data["blob"].update(blob)
```
bhearsum: This causes race conditions if toplevel and secondary toplevel submission run at the same time - we end up losing some of the fileUrls. If we deep merge here (so we keep all entries) it should work fine.

bhearsum (commented):
This patch almost works - but it needs to deep merge instead of just updating before it can be merge (see below).

bhearsum:
If someone else needs to verify this, push this to dev balrogscript, run `./mach try release --migration beta-to-release --version 85.0 --worker-suffix balrog=-dev`, and then add the two `toplevel` jobs to the resulting push. Once it's complete, make sure you have beta+release entries in the resulting balrog blob.

balrogscript/src/balrogscript/submitter/cli.py
escapewindow: https://stackoverflow.com/questions/7204805/how-to-merge-dictionaries-of-dictionaries/7205107#7205107

https://www.electricmonk.nl/log/2017/05/07/merging-two-python-dictionaries-by-deep-updating/

https://y.tsutsumi.io/2017/04/24/deepmerge-deep-merge-dictionaries-lists-and-more-in-python/

The first seems to only deal with sub-dictionaries; the other two deal with lists and sets as well.

PR #296, opened 2020-11-18
Remove unreachable code in beetmoverscript

alopezz:
While working on #280 (Bug 1600916) I found that there is at least one code path that is never taken when running the script normally, since it would require a task to be both `push-to-nightly` and `push-to-maven` at the same time. I've removed the affected code in this commit.

I already talked to @JohanLorenzo about this and I think @MihaiTabara also confirmed that this is safe to remove. The integration tests added in [Bug 1672691](https://bugzilla.mozilla.org/show_bug.cgi?id=1672691) pass with these changes.

Now I wonder if this code path can also be removed:

https://github.com/mozilla-releng/scriptworker-scripts/blob/a196a7a660a09de392e891d1fe0ea1c24e2381a8/beetmoverscript/src/beetmoverscript/script.py#L104-L122

Removing that `else`, the integration tests keep passing. It's mainly a matter of knowing if there are still `push-to-nightly` or `push-to-candidates` tasks that may come without an `artifactMap`. The fact that templates only exist for `firefox_partner` and and `firefox_partner_repacks` seems to support this. And this would mean that `generate_beetmover_manifest` and the code that uses jinja templates would be reduced to only the `push-to-partner` tasks. So please let me know if you think it would be a good idea to trim these parts as well.


beetmoverscript/src/beetmoverscript/utils.py
```diff
@@ -123,9 +121,6 @@ def generate_beetmover_template_args(context):
     task = context.task
     release_props = context.release_props
 
-    if is_maven_action(context.action):
```
MihaiTabara, line 126: Nit: not sure we can remove this function as it's used here too - https://github.com/mozilla-releng/scriptworker-scripts/blob/master/beetmoverscript/src/beetmoverscript/task.py#L25 in order to check the schema validation.

beetmoverscript/src/beetmoverscript/utils.py
```diff
@@ -178,30 +173,6 @@ def generate_beetmover_template_args(context):
     return tmpl_args
 
 
-def _generate_beetmover_template_args_maven(task, release_props):
```
MihaiTabara, line 181: This LGTM! the only place where we don't have declarative artifacts is handling the partner repacks. Everywhere else, we no longer rely on manifests, but on payload definitions so removing this here makes total sense as it's dead code.

MihaiTabara (commented):
First things first, sorry for delays in reviewing this patch, we've had a few days off as a company and things piled up.

Secondly, thank you for continuing with this great work! The PR looks good and helps us clean the dead code that we have, following up the switch to declarative artifacts. The PR looks good, modulo one tiny nit with one of the functions. Let me know if that makes sense.

beetmoverscript/src/beetmoverscript/utils.py
MihaiTabara: Come to think about it, I'm surprised tests passed with this function being removed. I would've assumed https://github.com/mozilla-releng/scriptworker-scripts/blob/master/beetmoverscript/tests/test_task.py#L33 to fail.

MihaiTabara: Indeed https://firefox-ci-tc.services.mozilla.com/tasks/U8VagEHyQ1CG2pi88P5_eg/runs/0/logs/https%3A%2F%2Ffirefox-ci-tc.services.mozilla.com%2Fapi%2Fqueue%2Fv1%2Ftask%2FU8VagEHyQ1CG2pi88P5_eg%2Fruns%2F0%2Fartifacts%2Fpublic%2Flogs%2Flive.log#L142 seems to have passed; that's odd 🤔

alopezz: But is it? The function is still there (L87), it's just this one call that I removed, which would always return false in that context.

MihaiTabara:
Hm, something's very odd with the tests, I've tried to reproduce this locally and when I try to run the tests, I expectedly get this:

```================================================================================================================================================== short test summary info ===================================================================================================================================================
FAILED tests/test_integration.py::test_main_maven_nightly_candidates[android_components_nightly-project:mobile:android-components:releng:beetmover:-12] - AttributeError: module 'beetmoverscript.utils' has no attribute 'is_maven_action'
FAILED tests/test_integration.py::test_main_maven_nightly_candidates[application_services_release-project:mozilla:app-services:releng:beetmover:-12] - AttributeError: module 'beetmoverscript.utils' has no attribute 'is_maven_action'
FAILED tests/test_integration.py::test_main_maven_nightly_candidates[geckoview_nightly-project:releng:beetmover:-16] - AttributeError: module 'beetmoverscript.utils' has no attribute 'is_maven_action'
FAILED tests/test_integration.py::test_main_maven_nightly_candidates[glean_release-project:mozilla:glean:releng:beetmover:-20] - AttributeError: module 'beetmoverscript.utils' has no attribute 'is_maven_action'
FAILED tests/test_integration.py::test_main_maven_nightly_candidates[firefox_desktop_nightly-project:releng:beetmover:-52] - AttributeError: module 'beetmoverscript.utils' has no attribute 'is_maven_action'
FAILED tests/test_integration.py::test_main_maven_nightly_candidates[firefox_langpacks_beta-project:releng:beetmover:-1] - AttributeError: module 'beetmoverscript.utils' has no attribute 'is_maven_action'
FAILED tests/test_integration.py::test_main_maven_nightly_candidates[firefox_generated_screenshots_release-project:releng:beetmover:-7] - AttributeError: module 'beetmoverscript.utils' has no attribute 'is_maven_action'
FAILED tests/test_integration.py::test_main_push_to_partner[firefox_eme_free_release-project:releng:beetmover:-2] - AttributeError: module 'beetmoverscript.utils' has no attribute 'is_maven_action'
FAILED tests/test_script.py::test_async_main[push-to-nightly-False-task.json] - AttributeError: module 'beetmoverscript.utils' has no attribute 'is_maven_action'
FAILED tests/test_script.py::test_async_main[push-to-nightly-False-task_artifact_map.json] - AttributeError: module 'beetmoverscript.utils' has no attribute 'is_maven_action'
FAILED tests/test_script.py::test_async_main[push-to-unknown-True-task.json] - AttributeError: module 'beetmoverscript.utils' has no attribute 'is_maven_action'
FAILED tests/test_task.py::test_get_schema_key_by_action[scopes0-maven_schema_file] - AttributeError: module 'beetmoverscript.utils' has no attribute 'is_maven_action'
FAILED tests/test_task.py::test_get_schema_key_by_action[scopes2-schema_file] - AttributeError: module 'beetmoverscript.utils' has no attribute 'is_maven_action'
FAILED tests/test_task.py::test_get_schema_key_by_action[scopes3-schema_file] - AttributeError: module 'beetmoverscript.utils' has no attribute 'is_maven_action'
FAILED tests/test_task.py::test_get_schema_key_by_action[scopes4-schema_file] - AttributeError: module 'beetmoverscript.utils' has no attribute 'is_maven_action'
FAILED tests/test_task.py::test_validate_task - AttributeError: module 'beetmoverscript.utils' has no attribute 'is_maven_action'
========================== 16 failed, 165 passed, 5 warnings in 2.34s 
```

@alopezz Out of curiosity, does this PR pass the tests locally for you?

beetmoverscript/src/beetmoverscript/utils.py
MihaiTabara: Don't mind me, I misread the code initially and changed it locally without double-checking again in the PR. This looks good!

MihaiTabara:
> Now I wonder if this code path can also be removed:
> 
> https://github.com/mozilla-releng/scriptworker-scripts/blob/a196a7a660a09de392e891d1fe0ea1c24e2381a8/beetmoverscript/src/beetmoverscript/script.py#L104-L122
> 
> Removing that `else`, the integration tests keep passing. It's mainly a matter of knowing if there are still `push-to-nightly` or `push-to-candidates` tasks that may come without an `artifactMap`. The fact that templates only exist for `firefox_partner` and and `firefox_partner_repacks` seems to support this. And this would mean that `generate_beetmover_manifest` and the code that uses jinja templates would be reduced to only the `push-to-partner` tasks. So please let me know if you think it would be a good idea to trim these parts as well.

Your assessment is 100% correct. Other than push to partners tasks such as [this](https://firefox-ci-tc.services.mozilla.com/tasks/eW-JIchfRDOf-BMSsInRUg) we no longer have any other tasks that don't have declarative artifacts (payload definition via `artifactMap` of the desired destinations), neither in the mobile world,  nor in the Gecko.

+1 to remove the lines you pointed at, either in this PR or the next, up to you.

MihaiTabara (approved):
@alopezz This LGTM, ready to merge. If you wanted to follow-up with your suggested removal change in `push_to_nightly`, happy to wait for that before merging. Otherwise I can merge as is, either wfm.

Thank you again for this!

alopezz:
I chose to raise an explicit exception on the else (when there is no `artifactMap`), rather than removing the check and letting it raise a `KeyError`. Let me know if that looks right.

Now I don't want to get overly ambitious with removing code here, but the functions `task.validate_bucket_paths` and `utils.alter_unpretty_contents` are no longer used in the code after removing this bit. However, I don't know if any of those may come handy again sometime in the future or if you think it's fine to get rid of them as well.

MihaiTabara:
> I chose to raise an explicit exception on the else (when there is no `artifactMap`), rather than removing the check and letting it raise a `KeyError`. Let me know if that looks right.

+1 for this, good point!

> Now I don't want to get overly ambitious with removing code here, but the functions `task.validate_bucket_paths` and `utils.alter_unpretty_contents` are no longer used in the code after removing this bit. However, I don't know if any of those may come handy again sometime in the future or if you think it's fine to get rid of them as well.

Good catch. Short answer: we can remove both those functions.
Longer answer: While for the `validate_bucket_paths` it's an easy answer, since that validation moved in-tree, for the second one I had to clean the dust off my brain for this to remember the context. Before switching from beetmover templates to in-tree tasks payload, we were doing some pretty naming manually.

In the `script_config` which governs the scriptworker's configuration, we have [this](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/beetmoverscript/config_example.json#L7) variable defined which describes which artifacts we should tamper to pretty-name their contents, before uploading them to S3. AFAIK, it's always been the `test_packages.json` which is an artifact generated by the builds, to keep track of the tests suites. 

Now, that file comes like [this](https://archive.mozilla.org/pub/firefox/candidates/84.0b7-candidates/build1/linux-i686/en-US/firefox-84.0b7.test_packages.json) from the builds. Beetmover was originally reading the contents of that file and was replacing each of the `<something>.tar.gz` file to their corresponding `s3_key` content. I dug in the old implementation of beetmover and found [this](https://github.com/mozilla-releng/beetmoverscript/commit/a899676dfc4dabd61cbd0e4e311481d086c27184#diff-47f8bbc29fa6df3ee97ff4dbbe4dc0e82dea679edcc29e93dfd5bf29eebeaad4L53). So before uploading the file, it would update its contents on the fly and then submit it to S3.

Bottom line, when we migrated this code to declarative artifacts, we missed doing this step it seems. But since that was two years ago and nobody complained, we're safe to remove it. Worst case we can revert it if we ever need to.

Interestingly, a while ago we've run an audit on our artifacts and [data showed](https://bugzilla.mozilla.org/show_bug.cgi?id=1530728#c15) that `test_packages.json` had 1-3 accesses per major release and almost never used for betas.

So that said, the bottom line is:
* yes, it's safe to remove both functions.
* if we ever need the pretty naming we can always revert the function from this PR
* that file that needs pretty naming is barely used so likely nobody noticed the error. at some point we should probably look into removing that altogether from the build artifacts that are being generated
* we should also update the `script_config` to trim that un-needed config value since we're here:
```
> grep -R 'blobs_needing_prettynaming_contents' * 
README.md:    "blobs_needing_prettynaming_contents": [
config_example.json:    "blobs_needing_prettynaming_contents": [
docker.d/worker.yml:blobs_needing_prettynaming_contents:
src/script_config.json:    "blobs_needing_prettynaming_contents": [
Binary file src/beetmoverscript/__pycache__/script.cpython-37.pyc matches
src/beetmoverscript/script.py:        blobs = context.config.get("blobs_needing_prettynaming_contents", [])
src/beetmoverscript/data/config_schema.json:        "blobs_needing_prettynaming_contents",
src/beetmoverscript/data/config_schema.json:        "blobs_needing_prettynaming_contents": {
tests/fake_config.json:    "blobs_needing_prettynaming_contents": [
```

alopezz:
Thanks for the detailed explanation.

This should now be ready to merge; no more removal suggestions from me ;)

MihaiTabara (approved):

MihaiTabara:
Great stuff, thanks for making this lighter!

PR #302, opened 2020-12-01
[RELENG-117] - iscript vpn

escapewindow:
Taking changes from https://gist.github.com/bhearsum/e4ca33eab3aa119bbdea94f568b4fe01 in a way that should work for both firefox and mozillavpn.

Draft: Needs testing.

### DIFF
--- a/iscript/src/iscript/mac.py
+++ b/iscript/src/iscript/mac.py
@@ -184,6 +184,25 @@ async def sign_geckodriver(config, sign_config, all_paths):
 
 
 # sign_app {{{1
+async def _do_sign_file(top_dir, abs_file, file_, sign_command, app_path_len, app_executable):
+    """Avoid flake8 complaining about sign_app being too complex."""
+    # Deal with inner .app's in sign_app, not here.
+    if top_dir[app_path_len:].count(".app") > 0:
+        log.debug("Skipping %s because it's part of an inner app.", abs_file)
+        return
+    # app_executable gets signed with the outer package.
+    if file_ == app_executable:
+        log.debug("Skipping %s because it's the main executable.", abs_file)
+        return
+    dir_ = os.path.dirname(abs_file)
+    await retry_async(
+        run_command,
+        args=[sign_command + [file_]],
+        kwargs={"cwd": dir_, "exception": IScriptError, "output_log_on_exception": True},
+        retry_exceptions=(IScriptError,),
+    )
+
+
 async def sign_app(sign_config, app_path, entitlements_path):
     """Sign the .app.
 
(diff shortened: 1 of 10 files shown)


escapewindow:
@bhearsum if you have cycles, a quick "does this look sane" might help. I do need to make sure I didn't break anything Firefox or MozillaVPN in making these changes.

(Let me know if I should split the changesets up to be more review friendly.)

iscript/src/iscript/mac.py
```diff
@@ -515,17 +516,22 @@ async def create_one_notarization_zipfile(work_dir, all_paths, path_attrs=("app_
         app.check_required_attrs(required_attrs)
         for path_attr in path_attrs:
             app_paths.append(os.path.relpath(getattr(app, path_attr), work_dir))
-    await run_command(["zip", "-r", zip_path, *app_paths], cwd=work_dir, exception=IScriptError)
+    if sign_config["zipfile_cmd"] == "zip":
+        await run_command(["zip", "-r", zip_path, *app_paths], cwd=work_dir, exception=IScriptError)
+    elif sign_config["zipfile_cmd"] == "ditto":
+        await run_command(["ditto", "-c", "-k", "--sequesterRsrc", "--keepParent", "0", zip_path], cwd=work_dir, exception=IScriptError)
```
bhearsum: If we're going to move commands out to the config, what do you think about putting all the args in the config too?

I also wonder if we should just try using `ditto` for Firefox and avoid this branch altogether.

iscript/src/iscript/mac.py
```diff
@@ -920,11 +926,11 @@ async def create_pkg_files(config, key_config, all_paths):
     for app in all_paths:
         # call set_app_path_and_name because we may not have called sign_app() earlier
         set_app_path_and_name(app)
-        app.pkg_path = app.app_path.replace(".app", ".pkg")
+        app.pkg_path = app.app_path.replace(".appex", ".pkg").replace(".app", ".pkg")
```
bhearsum, line 924: Ah...thank you for improving this from the gist version!

bhearsum (commented):
This looks pretty sane to me, modulo testing. AFAICT this should leave things unchanged for Firefox, and includes everything needed for MozillaVPN. There's one minor improvement suggestion below, but not a blocker. I'm happy to help test this on mac-v3-signing1 when you're ready.

iscript/src/iscript/mac.py
escapewindow: > If we're going to move commands out to the config, what do you think about putting all the args in the config too?

Hm, could do.

> I also wonder if we should just try using `ditto` for Firefox and avoid this branch altogether.

The reason I branched was Nick and I tried using `ditto` for Firefox and ended up with a ton of extra cruft in the .app that ended up breaking notarization. I don't know if we used these exact args, though, so it might be worth trying again.

bhearsum: Maybe we'll give it a try once and see if it works? If it breaks notarization that's a pretty clear and obvious failure.

escapewindow: I was able to notarize Firefox successfully! It passed `codesign -vvvv` and `spctl --assess --verbose` checks. We need to test MozillaVPN signing, and then we're ready for review/landing/rollout.

bhearsum: I tested this for MozillaVPN this morning by:
* Replacing the `iscript` package in the scriptworker mac-v3-signing1 virtualenv with the version from your branch
* Tweaking `task.json` to set `product` to `mozillavpn`
* Signing a mozilla vpn dmg.

AFAICT it worked fine. I'm good with landing this at your convenience.

bhearsum (approved):

escapewindow:
Thanks!

PR #305, opened 2020-12-04
RELENG-167: Add support for signing system extensions in iscript

bhearsum:
This has been tested for MozillaVPN, but not for other products. It's hard to imagine how this could break Firefox though.

### DIFF
--- a/iscript/src/iscript/mac.py
+++ b/iscript/src/iscript/mac.py
@@ -926,7 +926,7 @@ async def create_pkg_files(config, sign_config, all_paths):
     for app in all_paths:
         # call set_app_path_and_name because we may not have called sign_app() earlier
         set_app_path_and_name(app)
-        app.pkg_path = app.app_path.replace(".appex", ".pkg").replace(".app", ".pkg").replace(".systemextension", ".pkg")
+        app.pkg_path = app.app_path.replace(".appex", ".pkg").replace(".app", ".pkg")
         app.pkg_name = os.path.basename(app.pkg_path)
         pkg_opts = []
         if sign_config.get("pkg_cert_id"):
@@ -1037,6 +1037,7 @@ async def notarize_behavior(config, task):
 
     sign_config = get_sign_config(config, task, base_key="mac_config")
     entitlements_path = await download_entitlements_file(config, sign_config, task)
+    path_attrs = ["app_path"]
 
     all_paths = get_app_paths(config, task)
     langpack_apps = filter_apps(all_paths, fmt="autograph_langpack")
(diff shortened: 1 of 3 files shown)


iscript/src/iscript/mac.py
```diff
@@ -377,7 +377,7 @@ def get_app_dir(parent_dir):
         UnknownAppDir: if there is no single app dir
 
     """
-    apps = glob("{}/*.app*".format(parent_dir)) + glob("{}/*/*.app*".format(parent_dir))
+    apps = glob("{}/*.app*".format(parent_dir)) + glob("{}/*/*.app*".format(parent_dir)) + glob("{}/*.systemextension*".format(parent_dir))
```
escapewindow, line 380: I'm ok with this. I could also see `iscript.constants.MAC_PRODUCT_CONFIG.PRODUCT.app_globs` or `app_extensions` if we need to customize this anymore :)

iscript/src/iscript/mac.py
```diff
@@ -926,7 +926,7 @@ async def create_pkg_files(config, sign_config, all_paths):
     for app in all_paths:
         # call set_app_path_and_name because we may not have called sign_app() earlier
         set_app_path_and_name(app)
-        app.pkg_path = app.app_path.replace(".appex", ".pkg").replace(".app", ".pkg")
+        app.pkg_path = app.app_path.replace(".appex", ".pkg").replace(".app", ".pkg").replace(".systemextension", ".pkg")
```
escapewindow: Interesting, we want to create .pkg files for the system extension(s)?
If not, we could also skip `.systemextension` apps during pkg creation.

escapewindow (approved):
I suppose we'll see if someone adds a `.systemextension` to Firefox?
I don't see one today. Otherwise, I might have asked you to do a manual run of an l10n `mac_notarize` task.

iscript/src/iscript/mac.py
bhearsum: We don't really want to (the very first thing we do with these is unpacking them to get the actual bits out), but this is what iscript spits out right now. Ideally (at least, IMO) we'd spit out the same format as the input.

bhearsum:
> I suppose we'll see if someone adds a .systemextension to Firefox?

If we do that I _think_ we'll need to start signing that separately from Firefox (like we do for the MozillaVPN components right now). I could be wrong though.

iscript/src/iscript/mac.py
escapewindow: iscript spits out a tarball and a pkg. if we only extract it from the tarball, we can probably skip the pkg ?

bhearsum: Oh god, I can't believe I missed that (pkgs are listed last in the log output, which is the only thing I saw). Yes, let's skip pkg generation for mozillavpn! I'll need to update the braindump scripts + this patch to do that.

escapewindow: Ok. Non-blocker. We waste a tiny bit of compute time and a tiny bit of disk space/network bandwidth for the pkg, but if we end up landing as-is I'm ok with it.

bhearsum: Too late, I already did this! I'll have a new version up for review shortly.

bhearsum:
I tested this with a recent notarize_1 and notarize_3 jobs from mozilla-beta, as well as with both distribution and developer id mozillavpn signing. They all worked fine.

escapewindow:
LGTM! Thank you!

PR #307, opened 2020-12-07
add dev env for githubscript

escapewindow:


githubscript/docker.d/worker.yml
```diff
@@ -20,4 +20,10 @@ github_projects:
         github_token: dummy
         github_owner: non-existing-owner
         github_repo_name: non-existing-repo
+    'COT_PRODUCT == "mobile" && ENV == "dev"':
+      mock:
+        allowed_actions: ["release"]
+        github_token: dummy
```
JohanLorenzo, line 31: We could use `{ "$eval": "GITHUB_TOKEN_WRITE_ACCESS_STAGING" }` here. This way, we actually exercise code that uploads something onto a github repo. What do you think?

githubscript/docker.d/worker.yml
```diff
@@ -20,4 +20,10 @@ github_projects:
...
+        github_owner: non-existing-owner
```
JohanLorenzo, line 32: Same here, we could wire this to your fork, so we sure this works end-to-end. The `mock` product is quite simple doesn't do much.

JohanLorenzo (approved):
This looks good to me. We can use the `mock` product against the dev workers. Although I think we could use this type of instance to actually write on a github thanks to https://github.com/mozilla-release-automation-bot-staging. 

I'm fine with both options, so feel free to choose the one you prefer.

escapewindow:
Added the `android-components` repo, pointing at my fork, for both staging and dev. Depends on https://github.com/mozilla-services/cloudops-infra/pull/2752 .

PR #308, opened 2020-12-08
Prevent beetmoverscript tests from modifying source files

### LINKED ISSUE #133: beetmoverscript tests modify source files
Beetmoverscript tests modify `beetmoverscript/tests/test_work_dir/cot/eSzfNqMZT_mSiQQXu8hyqg/public/build/buildhub.json`. Let's stop doing that, probably by making a copy. We could even make this a zip or tarball, and extract it before the test.

alopezz:
Fixes: #133

I've used a very simple approach of adding a fixture where I copy the affected file to a temporary location before the test, and restore it after the test. Only one test is affected by this, so this may be a good solution to keep things simple. However, I'm open to making it more sophisticaded if requested; or also I may move the fixture from immediately above the affected test somewhere else if it makes more sense.

### DIFF
--- a/beetmoverscript/tests/test_script.py
+++ b/beetmoverscript/tests/test_script.py
@@ -1,6 +1,7 @@
 import logging
 import mimetypes
 import os
+import shutil
 
 import boto3
 import mock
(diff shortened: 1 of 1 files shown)


MihaiTabara (approved):
Sorry for delays in reviewing this. I think the solution works for now, given the limited scope of the test. Thank you for this fix!

PR #309, opened 2020-12-10
Bug 1619767 - [treescript] Do not use upstream_repo on Thunderbird clones.

jfx2006:
As there is no comm-unified repository with all of the various branch heads,
separate pulls are needed for the to_branch and from_branch when doing merges
on those repositories.

The separate pulls are triggered by a false value for upstream_repo, in this case
an empty string. Usually, that's set to comm-central and hg robustcheckout does
what it needs to do, but it's not a required parameter.


jfx2006:
With this change, I'm able to complete a comm-central to comm-beta merge and a comm-central version bump running Treescript locally.

escapewindow (approved):
This looks good, and it looks like this won't affect firefox. Thank you!

escapewindow:
Pushed to `production-treescript`, so this should roll out in ~15 min.

PR #312, opened 2020-12-14
RELENG-177: Add support for inserting provisioning profiles in iscript

bhearsum:
Nothing revolutionary here - it pretty much just follows the entitlements url model, except I didn't see a need to be able to disable it at the `sign_config` level, so I made it key off of presence (or not) in the payload.

I've done one test of this on mac-v3-signing1, and it appeared to work fine.


escapewindow (approved):
This looks good, thank you!

PR #313, opened 2020-12-17
make contact_github per-product rather than top level config

### LINKED ISSUE #311: githubscript: make `contact_github` project specific
Let's make the `mock` `contact_github` false, but allow for `contact_github` to be true for `staging-android-components` and the like. That means we need to make `contact_github` dependent on the project, rather than the environment.

escapewindow:
Fixes #311

Tested with https://firefox-ci-tc.services.mozilla.com/tasks/R-Nj1SCbR862krJX5ksnQQ#artifacts , looks like it's working?


JohanLorenzo (approved):
Great fix! That's the way githubscript should evolve, in my opinion. Thanks a lot for taking care of it!

escapewindow:
Thanks!

PR #315, opened 2020-12-18
Don't sign some files with entitlements for MozillaVPN

bhearsum:
We discovered today that `wg` and `wireguard-go` break if they're signed with entitlements. So, we have to sign them as part of the Mozilla VPN.app signing, but the signing command has to be different.

### DIFF
--- a/iscript/src/iscript/mac.py
+++ b/iscript/src/iscript/mac.py
@@ -225,14 +225,14 @@ async def sign_app(sign_config, app_path, entitlements_path, provisioning_profil
     identity = sign_config["identity"]
     keychain = sign_config["signing_keychain"]
     sign_command = _get_sign_command(identity, keychain, sign_config)
+    sign_command.extend(["-o", "runtime"])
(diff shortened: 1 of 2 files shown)


iscript/src/iscript/mac.py
```diff
@@ -257,7 +257,10 @@ async def sign_app(sign_config, app_path, entitlements_path, provisioning_profil
 
         for file_ in files:
             abs_file = os.path.join(top_dir, file_)
-            await _do_sign_file(top_dir, abs_file, file_, sign_command, app_path_len, app_executable)
+            if sign_config.get("sign_with_entitlements", False) or file_ in sign_config.get("no_entitlements_files", []):
```
escapewindow: shouldn't this be `if not sign_config.get("sign_with_entitlements", False) or ...` ?

escapewindow (approved):
Looks good, other than the boolean bug. Thanks!

iscript/src/iscript/mac.py
bhearsum: ummm, oops!

PR #318, opened 2021-01-15
Update docstring for skip_test_only

jcristau:
The doc was out of sync with the code, since we now look at every changeset and not just the tip.


escapewindow (approved):

escapewindow:
Thanks!

PR #322, opened 2021-01-28
bug 1689376 - convert pkg to a distribution pkg

escapewindow:
Also back to 100% coverage.

### DIFF
--- a/iscript/src/iscript/mac.py
+++ b/iscript/src/iscript/mac.py
@@ -933,17 +933,18 @@ async def create_pkg_files(config, sign_config, all_paths):
 
     """
     log.info("Creating PKG files")
-    futures = []
+    steps = [[], []]
     semaphore = asyncio.Semaphore(config.get("concurrency_limit", 2))
     for app in all_paths:
         # call set_app_path_and_name because we may not have called sign_app() earlier
         set_app_path_and_name(app)
+        app.tmp_pkg_path = app.app_path.replace(".appex", ".tmp.pkg").replace(".app", ".tmp.pkg")
         app.pkg_path = app.app_path.replace(".appex", ".pkg").replace(".app", ".pkg")
         app.pkg_name = os.path.basename(app.pkg_path)
         pkg_opts = []
         if sign_config.get("pkg_cert_id"):
             pkg_opts = ["--sign", sign_config["pkg_cert_id"]]
-        futures.append(
+        steps[0].append(
             asyncio.ensure_future(
                 semaphore_wrapper(
                     semaphore,
@@ -956,6 +957,27 @@ async def create_pkg_files(config, sign_config, all_paths):
                             "/Applications",
                             "--component",
                             app.app_path,
+                            app.tmp_pkg_path,
+                        ]
+                        + pkg_opts,
+                        cwd=app.parent_dir,
+                        exception=IScriptError,
+                    ),
+                )
+            )
+        )
+        # Bug 1689376 - distribution pkg
+        steps[1].append(
+            asyncio.ensure_future(
+                semaphore_wrapper(
+                    semaphore,
+                    run_command(
+                        [
+                            "productbuild",
+                            "--keychain",
+                            sign_config["signing_keychain"],
+                            "--package",
+                            app.tmp_pkg_path,
                             app.pkg_path,
                         ]
                         + pkg_opts,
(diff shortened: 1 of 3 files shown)


escapewindow:
Broken: https://gist.github.com/escapewindow/1f443c851c69ac876052ce737cc67a9a

escapewindow:
e7c4b05 fixes the `productbuild` call, currently with `--synthesize`. However, notarization stapling appears broken. Are we supposed to notarize the `pkgbuild` pkg, but not the `productbuild` pkg? Are we not supposed to staple at all?

escapewindow:
https://developer.apple.com/documentation/xcode/notarizing_macos_software_before_distribution mentions we can notarize `Flat installer packages`. I'm going to try without `--synthesize`.

It's also possible we need to:
- sign with `pkgbuild`
- notarize
- poll Apple
- staple
- *THEN* `productbuild`.

I hope not, but that would be the next thing I can try after removing `--synthesize`.

escapewindow:
Removing `--synthesize` made notarization stapling work again :)
[Suspicious Package](https://www.mothersruin.com/software/SuspiciousPackage/) thinks the pkg looks good!
I was able to install it locally.
I sent it to @mkaply over Slack DM.

iscript/src/iscript/mac.py
```diff
@@ -954,11 +955,39 @@ async def create_pkg_files(config, sign_config, all_paths):
...
+                            "--keychain",
+                            sign_config["signing_keychain"],
                         ]
-                        + pkg_opts,
+                        + pkg_opts
+                        + [
+                            "--package",
+                            app.tmp_pkg_path,
+                            app.pkg_path,
+                        ],
```
bhearsum, line 1025: If this is the final pkg, I think we need to run `productsign`, too, otherwise the pkg is totally unsigned. (See https://hg.mozilla.org/build/braindump/file/tip/signing-related/sign-mozilla-vpn.sh#l106 for how we do this for MozillaVPN.)

bhearsum (commented):
Are there any other apps we need to test here? MozillaVPN can be ignored (it's packaged outside of iscript, and already uses pkgbuild+productbuild).

See also the comment below about potentially needing to sign it.

iscript/src/iscript/mac.py
escapewindow: Ah. Maybe that's why mkaply said the pkg installed fine but crashed on startup with a signing error.

escapewindow:
> Are there any other apps we need to test here? MozillaVPN can be ignored (it's packaged outside of iscript, and already uses pkgbuild+productbuild).

I'm not aware of anything else that uses the pkg (or the mac signing pool).

escapewindow:
We're still crashing with `productsign`. It's possible I'm doing something wrong, and it's possible something's different about productsigning Firefox than MozillaVPN.

escapewindow:
I'm also wondering if we need a different pkg signing cert for a distribution package.

escapewindow:
The distribution pkg starts up if we back out 176c04e .

bhearsum (approved):
> The distribution pkg starts up if we back out [176c04e](https://github.com/mozilla-releng/scriptworker-scripts/commit/176c04e46688841f5c844720bbf07912f800fd32) .

Interesting. Based on this code and the configs I see I'm a bit surprised we're signing with the correct cert, but if it works, great! I'd love to inspect one of the signed packages a sanity check still.

mkaply:
So that backout actually makes sense to me. I was working on another project outside Mozilla trying to get it signed and I kept getting that exact same crash on startup. I discovered it was because I add --entitlements to too many things.

If you look at:

https://searchfox.org/mozilla-central/source/security/mac/hardenedruntime/codesign.bash#115

We don't do entitlements for everything.

bhearsum:
> So that backout actually makes sense to me. I was working on another project outside Mozilla trying to get it signed and I kept getting that exact same crash on startup. I discovered it was because I add --entitlements to too many things.
> 
> If you look at:
> 
> https://searchfox.org/mozilla-central/source/security/mac/hardenedruntime/codesign.bash#115
> 
> We don't do entitlements for everything.

I am somewhat surprised by this. My experience with MozillaVPN is that we needed entitlements for some inner parts of the packages as well; but maybe that was special they use Network Extensions.

escapewindow:
Per group DM, the signatures look good, and we're able to install and launch without crashing. Merging...

PR #323, opened 2021-01-29
Openh264 mac

escapewindow:


iscript/src/iscript/data/i_task_schema.json
```diff
@@ -65,6 +65,9 @@
             "entitlements-url": {
               "type": "string"
             },
+            "single-file-name": {
```
bhearsum: Can we make this a list of files to sign instead of a single file? I can easily see a case in the future where we sign multiple, single files, from an upstream task rather than just one.

iscript/src/iscript/mac.py
```diff
@@ -179,7 +178,7 @@ async def sign_geckodriver(config, sign_config, all_paths):
         env["COPYFILE_DISABLE"] = "1"
         makedirs(os.path.dirname(app.target_tar_path))
         await run_command(
-            ["tar", _get_tar_create_options(app.target_tar_path), app.target_tar_path, file_], cwd=app.parent_dir, env=env, exception=IScriptError
+            ["tar", _get_tar_create_options(app.target_tar_path), app.target_tar_path, filename], cwd=app.parent_dir, env=env, exception=IScriptError
```
bhearsum: Not a blocker, but why do we even bother tar'ing this up whine it's just one file?

iscript/src/iscript/mac.py
```diff
@@ -1294,6 +1293,7 @@ async def geckodriver_behavior(config, task):
 
     """
     sign_config = get_sign_config(config, task, base_key="mac_config")
+    filename = get_single_file_name(task)
```
bhearsum: AFAICT this will break existing geckodriver tasks, since they don't define `single-file-name`.

bhearsum (commented):
I have a couple of suggestions below, but I also wonder if we need to support both the old and new geckodriver behaviour at once, to avoid busting everything when this lands, unless you think it's easy enough to update all of the geckodriver tasks on all branches before this lands.

iscript/src/iscript/data/i_task_schema.json
escapewindow: I thought about this. Now it's an optional `singleFileName` in each `upstreamArtifacts` dictionary, that defaults to `geckodriver` until we update all the geckodriver tasks.

iscript/src/iscript/mac.py
escapewindow: I don't know. I think it's because they wanted geckodriver in a tarball?

iscript/src/iscript/mac.py
escapewindow: I think this will work, since it defaults to `geckodriver` if it isn't specified. We'll see.

iscript/src/iscript/data/i_task_schema.json
bhearsum: Good idea!

iscript/src/iscript/mac.py
escapewindow: Now that we have a list of globs, in theory they're all going to go in the tarball.

iscript/src/iscript/data/i_task_schema.json
escapewindow: singleFileName -> singleFileGlobs

bhearsum (approved):
This looks good to me.

PR #324, opened 2021-02-02
[bug 1654250] surface -11 and -15 exit codes in run_command

escapewindow:
I think we're hitting -11 exit codes in iscript when we restart scriptworker. Scriptworker would turn a -11 exit code into an `intermittent-task`, which would automatically retry the task. However, `run_command` turns that -11 exit code into an exit code of 1, which is `failed`.

Let's surface the -11 and -15 exit codes if they show up in `run_command`.

If we're worried about how large of a change this is, we could set the default `copy_exit_codes` to `None` and only override the `copy_exit_codes` in the iscript `run_command` calls.

Besides the unit tests, I also had a test script where I did

```python
#!/usr/bin/env python

from scriptworker_client.exceptions import TaskError

t = TaskError("foo", exit_code=-11)
print(t.exit_code)
```

which gave me -11. Setting the exit code through bash seems to translate `-11` to `245`, though.

### DIFF
--- a/scriptworker_client/src/scriptworker_client/utils.py
+++ b/scriptworker_client/src/scriptworker_client/utils.py
@@ -29,7 +29,7 @@ from typing import (
     Union,
 )
 from urllib.parse import unquote, urlparse
-from scriptworker_client.exceptions import TaskError
+from scriptworker_client.exceptions import ClientError, TaskError
 
 log = logging.getLogger(__name__)
 
@@ -166,6 +166,13 @@ def get_log_filehandle(log_path=None):
 
 
 # run_command {{{1
+def _get_exception_kwargs(exception, exitcode, copy_exit_codes):
+    kwargs = {}
+    if issubclass(exception, ClientError) and exitcode in copy_exit_codes:
+        kwargs["exit_code"] = exitcode
+    return kwargs
+
+
 async def run_command(
     cmd,
     log_path=None,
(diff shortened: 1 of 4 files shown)


escapewindow:
Hm. I'm wondering if we should also add `245` and `241` to the scriptworker [`reversed_statuses`](https://github.com/mozilla-releng/scriptworker/blob/master/src/scriptworker/constants.py#L52)...

escapewindow:
Tested, along with https://github.com/mozilla-releng/scriptworker/pull/480 , in https://treeherder.mozilla.org/jobs?repo=try&revision=3a178a506ea21ac97a82dfb434fb1d0e5117ec61&selectedTaskRun=MQsIpnh9S5mIzih6fsWEJw.9

Took me a while to figure out what was going on. `reversed_statuses` hasn't ever worked, as far as I can tell. =\
Works now...

scriptworker_client/src/scriptworker_client/utils.py
```diff
@@ -175,6 +182,7 @@ async def run_command(
     env=None,
     exception=None,
     expected_exit_codes=(0,),
+    copy_exit_codes=(245, 241),  # 245 == -11, 241 == -15
```
bhearsum: This could do with a comment. I understand that we get to this by subtracting the real exit code from 256, but I have no idea why that happens.

bhearsum (approved):

escapewindow:
Thanks!

PR #325, opened 2021-02-05
bug 1689232 - add single files to the zipfile

escapewindow:
I'm not sure if we'll want this behavior for future zipfiles or not. Following YAGNI here.


iscript/src/iscript/mac.py
```diff
@@ -165,9 +165,7 @@ async def sign_single_files(config, sign_config, all_paths):
 
     for app in all_paths:
         app.check_required_attrs(["orig_path", "parent_dir", "artifact_prefix", "single_file_globs"])
-        app.target_tar_path = "{}/{}{}".format(config["artifact_dir"], app.artifact_prefix, app.orig_path.split(app.artifact_prefix)[1]).replace(
-            ".zip", ".tar.gz"
-        )
+        app.target_tar_path = "{}/{}{}".format(config["artifact_dir"], app.artifact_prefix, app.orig_path.split(app.artifact_prefix)[1])
```
bhearsum: Can we rename this variable, given that it's not always a tarball now?

bhearsum (approved):
Just to make sure I understand: this changes means that we'll output the same type of package as we got in, right? If so, I think that's fine, and we can adjust later if needed.

iscript/src/iscript/mac.py
escapewindow: I thought about that. `target_bundle_path` ?

bhearsum: wfm

PR #326, opened 2021-02-05
Sign with entitlements when required, including app bundles

bhearsum:
Previous version of this was #315, which didn't sign app bundles with entitlements ever (only inner files), which caused a Firefox start-up crash. I ended up doing a bit of refactoring to make this cleaner (at least, IMO).

Tested Mozilla VPN & Firefox signing. I can send signed binaries privately if you want to inspect them.


iscript/src/iscript/mac.py
```diff
@@ -143,8 +143,26 @@ def get_bundle_executable(appdir):
...
+        "-fv",
+        "--keychain",
+        keychain,
+        "--requirement",
+        sign_config["designated_requirements"] % {"subject_ou": identity},
+    ]
+
+    if file_ and file_ in sign_config.get("hardened_runtime_files", []):
+        sign_command.extend(["-o", "runtime"])
+    elif sign_config.get("sign_with_entitlements", False) and (file_ and file_ not in sign_config.get("no_entitlements_files", [])):
```
escapewindow: Is there a reason we have both `hardened_runtime_files` and `no_entitlements_files`? (Will we need some files with no `-o runtime` at some point?)

Similarly, since these blocks are mutually exclusive, is there a reason why `wg` and `wireguard-go` are in both `hardened_runtime_files` and `no_entitlements_files` ?

If we see needing more complex special configs per file, we may want a dictionary keyed by filename. We don't have to go down that route unless we see it coming shortly. I mainly want to make sure we want two config settings here, when it seems like one would solve the issue at hand.

iscript/src/iscript/mac.py
```diff
@@ -279,11 +293,13 @@ async def sign_app(sign_config, app_path, entitlements_path, provisioning_profil
 
         for file_ in files:
             abs_file = os.path.join(top_dir, file_)
+            sign_command = _get_sign_command(identity, keychain, sign_config, file_, entitlements_path)
```
escapewindow: Nit: I tend to prefer explicitly setting kwargs, e.g. `file_=file_`. Non-blocking if it passes the linter.

iscript/src/iscript/mac.py
```diff
@@ -279,11 +293,13 @@ async def sign_app(sign_config, app_path, entitlements_path, provisioning_profil
...
             await _do_sign_file(top_dir, abs_file, file_, sign_command, app_path_len, app_executable)
 
-    await sign_libclearkey(contents_dir, sign_command, app_path)
+    await sign_libclearkey(contents_dir, _get_sign_command(identity, keychain, sign_config), app_path)
 
     # sign bundle
+    sign_command = _get_sign_command(identity, keychain, sign_config, entitlements_path=entitlements_path)
```
escapewindow, line 300: Will we ever need to not use an entitlements file or hardened runtime on the top level app?

escapewindow (changes requested):
It's possible we can land this without changes, depending on the answers to my questions.

iscript/src/iscript/mac.py
bhearsum: I can try to simplify this, I agree it's not great, and it ended up this why through evolution, not design...

Truthfully though, I'm treating this is a stop-gap solution until we have a proper manifest for specifying these things for an entire app (ideally one that comes in as part of of the task payload).

escapewindow: LMK. If it turns out this is the best option for reasons, we can land as is. If there's a quick way to improve it, let's do that.

iscript/src/iscript/mac.py
bhearsum: I don't know! I suspect not. I didn't think a YAGNI philosophy would make sense here though, since it was trivial to support all paths.

iscript/src/iscript/mac.py
bhearsum: Okay, so after some review I believe the following is true:
* We have to keep `no_entitlements_files` (or something equivalent), because we cannot sign two of the vpn binaries with entitlements.
* If we want to preserve the behaviour of _not_ signing with a hardened runtime by default, we have to keep `hardened_runtime_files` (to enable it for these same files without enabling entitlements). The alternative is to always include `-o runtime` in the signing command, which may or may not break something.

Do you have a preference here?

escapewindow: > Okay, so after some review I believe the following is true:
> 
>     * We have to keep `no_entitlements_files` (or something equivalent), because we cannot sign two of the vpn binaries with entitlements.

The `no_entitlements_files` check is in an `elif` where we've already hit matched the `if` block (`hardened_runtime_files`). As far as I can see, nothing uses `no_entitlements_files` without `hardened_runtime_files`. Do you think we'll still need both in some future use case?

escapewindow (approved):
I'm going to mark this approved, to unblock you when you think you've got this right.

iscript/src/iscript/mac.py
escapewindow: (Marked approved. If the answer is `no`, we can remove `no_entitlements_files`. If the answer is `yes`, we can land as is.)

bhearsum: Oh, I see. Yes, we could get rid of `no_entitlements_files`, but I'd probably rename `hardened_runtime_files` to `hardened_runtime_only_files` for clarity (otherwise one might expect them to still get entitlements, and anything else _not_ to have hardened runtime).

bhearsum:
@escapewindow - do you want another look at this before it lands?

iscript/src/iscript/mac.py
```diff
@@ -279,11 +291,13 @@ async def sign_app(sign_config, app_path, entitlements_path, provisioning_profil
 
         for file_ in files:
             abs_file = os.path.join(top_dir, file_)
+            sign_command = _get_sign_command(identity, keychain, sign_config, file_=file_, entitlements_path=entitlements_path)
             await _do_sign_file(top_dir, abs_file, file_, sign_command, app_path_len, app_executable)
 
-    await sign_libclearkey(contents_dir, sign_command, app_path)
+    await sign_libclearkey(contents_dir, _get_sign_command(identity, keychain, sign_config), app_path)
```
escapewindow: Does this mean we no longer sign libclearkey with entitlements?

escapewindow (commented):
Other than the libclearkey question, this looks good! Thank you!

iscript/src/iscript/mac.py
escapewindow: (Did we check to make sure Firefox still starts up with this patch applied?)

bhearsum: I did! But I'm going to do some log comparison to make sure signing command invocations haven't changed vs. current deployed code.

bhearsum: Indeed! Thanks for catching this. Firefox still runs with it, but I imagine this would mean Widevine would break.

bhearsum:
With the latest patch, codesign commands are 100% the same as the current ones for Firefox.

PR #327, opened 2021-02-09
Fix dep pkg

escapewindow:
Bustage in #firefox-ci

### DIFF
--- a/iscript/src/iscript/mac.py
+++ b/iscript/src/iscript/mac.py
@@ -1018,27 +1018,28 @@ async def create_pkg_files(config, sign_config, all_paths):
         )
     await raise_future_exceptions(futures)
     futures = []
-    for app in all_paths:
-        # Bug 1689376 - sign distribution pkg
-        futures.append(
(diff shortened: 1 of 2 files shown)


iscript/src/iscript/mac.py
```diff
@@ -1019,26 +1019,29 @@ async def create_pkg_files(config, sign_config, all_paths):
...
+                            ],
+                            cwd=app.parent_dir,
+                            exception=IScriptError,
+                        ),
+                    )
                 )
             )
-        )
+        else:
+            copy2(app.tmp_pkg_path2, app.pkg_path)
```
jmaher, line 1044: is there a risk of people accidentally thinking this is signed when in fact it is not?

escapewindow: Currently we're only skipping the signing on dep signing, which is ok. It would be problematic if we skipped signing on release/nightly signing, yes.

jmaher (approved):

PR #330, opened 2021-02-11
[RELENG-42] bump scriptworker 36.0.5, other deps

escapewindow:
Largely to roll out the new CoT keys, but it's good to pick up the other dependency updates periodically as well.


addonscript/requirements/base.py37.txt
```diff
@@ -100,63 +109,65 @@ cffi==1.14.4 \
...
+    --hash=sha256:b97d804b1e9b523befed77c48dacec60e6dcb0b5391d57af6a65a312a90648c0
+    # via
+    #   requests
+    #   yarl
+immutabledict==1.2.0 \
+    --hash=sha256:1e6bb628eb158d32caf961bed5bb283f9e57201b50f48cd5063175612f0f3efd \
+    --hash=sha256:9ba015dc189af8aac5921248af18a06d76a1d026e362f917a8e9ee29e8de5133
     # via scriptworker
-importlib-metadata==1.7.0 ; python_version < "3.8" --hash=sha256:90bb658cdbbf6d1735b6341ce708fc7024a3e14e99ffdc5783edea9f9b077f83 --hash=sha256:dc15b2969b4ce36305c51eebe62d418ac7791e9a157911d58bfb1f9ccd8e2070 # via -r requirements/base.in, jsonschema
+importlib-metadata==1.7.0 ; python_version < "3.8"     --hash=sha256:90bb658cdbbf6d1735b6341ce708fc7024a3e14e99ffdc5783edea9f9b077f83     --hash=sha256:dc15b2969b4ce36305c51eebe62d418ac7791e9a157911d58bfb1f9ccd8e2070
```
escapewindow, line 153: Ooh, this may mean we no longer need two separate sets of requirements files??

addonscript/requirements/base.py37.txt
```diff
@@ -195,84 +206,108 @@ multidict==5.1.0 \
...
     --hash=sha256:5e25e7e6818e8877178b175ff43d2e6548afad72694aa125f404a7329ece0973 \
     --hash=sha256:b25e122ecec249c4299ac7b20b08db76e3e2025bdaeb699a9d444556de5fd367 \
-    --hash=sha256:f66dcbd6572a6216ab65949f0fa0b91f2df647918028436c384e6af5cd12ae2b \
+    --hash=sha256:f66dcbd6572a6216ab65949f0fa0b91f2df647918028436c384e6af5cd12ae2b
     # via taskcluster
-taskcluster==39.1.1 \
-    --hash=sha256:608631462b881118a279e1825aa5b10745d8acefb612e72c6fdb1af728bcfae3 \
-    --hash=sha256:66820075f6f4b6911e4df0679c460451e86e203b73a07f9ccd0780f671822628 \
-    --hash=sha256:fbf05197991134fc541a4cc5676bb9eebbf409e61fa889aa38e40e38a73ab476 \
+taskcluster==41.0.0 \
```
escapewindow: Hm. Our cluster is on 39.x.x and we're bumping the taskcluster client to 41.0.0. Asking to see if this will cause issues.

escapewindow:
> aki
> will we hit issues if we upgrade scriptworkers to use taskcluster-client.py to 41.0.0 while our cluster is on 39.x ? not recommended ?
> dustin (he/him)
> yes, don't do that
> that's general advice but I'll add something to the v41 release notes too
> because it very specifically breaks in this case (queue.claimWorker(taskQueueId, ..) is not compatible with old queue services)
> aki
> ty
> dustin (he/him)
> 
> https://github.com/taskcluster/taskcluster/releases/tag/v41.0.0
> 
> NOTE: client compatibility is only guaranteed in a forward direction -- an older client can talk to a newer service, and that remains the case with this release. However, this change will cause issues if a newer (41.x or higher) client is used to make any of the above API calls to an older (40.x or lower) service,

addonscript/requirements/base.py37.txt
escapewindow: It will, so I pinned tc to <40 in all relevant base.in files. We'll have to bump this pin manually when our cluster upgrades to 40. https://github.com/mozilla-releng/scriptworker/issues/485 will block us upgrading to 41.

bhearsum (approved):

escapewindow:
Thanks!

PR #338, opened 2021-03-08
bump dependencies (scriptworker-37.0.2)

escapewindow:
The main bump here is scriptworker 37.0.2, which will prevent scriptworker from dying due to arrow timestamp incompatibilities.


escapewindow:
Addonscript bustage appears to be https://github.com/pnuckowski/aioresponses/issues/183 .

escapewindow:
We can pin aiohttp to 3.7.4 for addonscript, but if we do so, we should pin chardet to <4.0.0 since [3.7.4.post0 was released to support chardet 4.0.0](https://docs.aiohttp.org/en/stable/changes.html).

bhearsum (approved):

PR #342, opened 2021-03-23
[treescript] Add minimal support for Suite version bumps.

jfx2006:
There are two cases where suite versions will be changed by automation.
- comm-central-to-beta
  This will be handled by a simple substitution in config.yml as it's "s/a1$/b1/"
- comm-bump-central
  This is what this commit covers. Just bump the minor number.


jfx2006:
I've worked around this another way, and suite code is being removed from comm-central soon so this is not needed.

PR #344, opened 2021-04-12
Bug 1703324 - Update windows certificate

ahal:


ahal:
TODO: actually generate the cert and key files

jmaher:
is this something that needs to be done in a certain time window?

escapewindow:
Looks like I appended it to the existing crt file in #354.

PR #350, opened 2021-04-20
Xpi + githubscript

jmaher:


githubscript/docker.d/worker.yml
```diff
@@ -42,26 +42,25 @@ github_projects:
...
         contact_github: true
     'COT_PRODUCT == "xpi" && ENV == "prod"':
-      xpi-manifest:
+      mozilla-extensions/xpi-manifest:
         allowed_actions: ["release"]
         github_token: { "$eval": "GITHUB_TOKEN_WRITE_ACCESS_PROD" }
         github_owner: mozilla-extensions
         github_repo_name: xpi-manifest
         contact_github: true
+        allow_github_repo_override: false
```
escapewindow: This will probably be `true` when we roll out to production, so we can push releases to the various addons repos.

githubscript/docker.d/worker.yml
```diff
@@ -42,26 +42,25 @@ github_projects:
         github_repo_name: staging-android-components
         contact_github: true
     'COT_PRODUCT == "xpi" && ENV == "prod"':
-      xpi-manifest:
+      mozilla-extensions/xpi-manifest:
```
escapewindow: This will probably be `mozilla-extensions/*` when we roll out to production, so we can push releases to the various addons repos. (We don't need to worry about this while you're testing.)

githubscript/src/githubscript/release_config.py
```diff
@@ -9,13 +9,22 @@
 
 
 def get_release_config(product_config, task_payload, config):
+    # TODO: add unittests for this, including fallback
+    # support repo override for xpi-manifest, but support original workflow for fenix
+    if "allow_github_repo_override" in product_config and product_config["allow_github_repo_override"] is True:
```
escapewindow: This could be `if product_config.get("allow_github_repo_override"):` which will be `True` or `False` if it's specified, or `None` if it isn't specified. We'd only enter the block if it were explicitly set to `True`.

githubscript/src/githubscript/script.py
```diff
@@ -16,7 +17,20 @@
...
-    project_config = config["github_projects"][project]
+
+    # TODO: write a unittest for this
+    projects = config["github_projects"].keys()
+    if project in projects:
+        project_config = config["github_projects"][project]
+    else:
+        project_config = {}
+        for p in projects:
+            if p.endswith("*"):
```
escapewindow: we probably want something like

```
if not p.endswith("*"):
    continue
cleaned = p.rstrip("*")
```

so we don't iterate over any non-`*` projects in this loop (including the `startswith` check).

githubscript/src/githubscript/release_config.py
```diff
@@ -9,13 +9,31 @@
 
 
 def get_release_config(product_config, task_payload, config):
+    # support repo override for xpi-manifest, but support original workflow for fenix
+    if product_config.get("allow_github_repo_override", False):
+        if not task_payload.get("githubOwner", ""):
+            raise TaskVerificationError("missing githubOwner from config")
```
escapewindow: `from task` ?

githubscript/src/githubscript/script.py
```diff
@@ -16,7 +17,20 @@
...
+        project_config = config["github_projects"][project]
+    else:
+        project_config = {}
+        for p in projects:
+            if not p.endswith("*"):
+                continue
+            cleaned = p.rstrip("*")
+            if project.startswith(cleaned):
+                if project_config != {}:
+                    raise TaskError(f'project "{project}" matches multiple configs in "{projects}"')
```
escapewindow, line 32: Nice!

githubscript/tests/test_release_config.py
```diff
@@ -43,6 +43,131 @@ def _dummy_get_artifacts(*args, **kwargs):
...
+        "git_tag": "v1.0.0",
+        "github_owner": "task_owner",
+        "github_repo_name": "task_repo",
+        "github_token": "some_secret_token",
+        "is_prerelease": False,
+        "release_name": "SomeProduct v1.0.0",
+    }
+
+
+def test_release_config_override_false(monkeypatch):
```
escapewindow, line 86: We might be able to avoid 3 different tests by using `@pytest.mark.parametrize`, but I don't think that needs to block.

escapewindow (approved):
Nice! The main question is, do we want to merge ~now and land another PR if we find issues while testing with shipit, or leave open until we have all the pieces ready? Either works for me.

PR #354, opened 2021-05-04
[RELENG-504] - New authenticode prod cert, same key

escapewindow:
The new cert expires `Jun 19 23:59:59 2024 GMT`. The keypair and keyid
are the same, so in theory we just need to update the cert?

I'm not sure if I should be adding the new cert to the .crt file, or replacing the old cert. I did the latter. I also kept the 2 other certs, which are both from Digicert and still valid. The 2nd is valid until 2028:

```
Not Before: Oct 22 12:00:00 2013 GMT
Not After : Oct 22 12:00:00 2028 GMT
```

The 3rd is valid until 2031:

```
Not Before: Nov 10 00:00:00 2006 GMT
Not After : Nov 10 00:00:00 2031 GMT
```

I'm also uncertain how to test this. One option is to create a signingscript virtualenv on a mac signer, create a config file with the appropriate autograph creds, and run a test task locally. Another option is to roll out to production and be ready to roll back. What do you two think?


bhearsum:
I think it would probably be best to keep both the old and the new cert in, at least temporarily, to make sure the transition can be made smoothly (unless we know we can cut over to the new signer at the same time as replacing this cert). Either way, I think adding a new .crt file with a new date is probably sensible for clarity alone.

If it's possible to associate a new format with a new autograph signer, that seems like the safest way to test. I'm OK with yolo'ing though, as long as a backout is quick.

escapewindow:
> I think it would probably be best to keep both the old and the new cert in, at least temporarily, to make sure the transition can be made smoothly (unless we know we can cut over to the new signer at the same time as replacing this cert). Either way, I think adding a new .crt file with a new date is probably sensible for clarity alone.
> 
> If it's possible to associate a new format with a new autograph signer, that seems like the safest way to test. I'm OK with yolo'ing though, as long as a backout is quick.

Aiui a new signer will require an off-cycle Autograph deploy. I can ask if that's doable.

escapewindow:
In Slack #autograph, we determined that this PR is probably doing the preferred thing, that the YOLO test-in-production approach is probably the best we can do given our current time and team bandwidth constraints, and that we will probably be in good shape after this PR rolls out, but we're not 100% sure about that.

escapewindow:
Ideally we roll out and test earlier. Tomorrow (Thurs) is probably the soonest we can; we need to be aware of the nightly schedule and Fx 89.0b9.

bhearsum (approved):

PR #355, opened 2021-05-08
allow for multiple beetmover taskcluster_scope_prefixes

escapewindow:
Currently untested.


beetmoverscript/src/beetmoverscript/task.py
```diff
@@ -34,9 +34,39 @@ def validate_task_schema(context):
...
+    return ["{}{}:".format(prefix, sub_namespace) for prefix in prefixes]
+
+
+def _extract_scopes_from_unique_prefix(scopes, prefixes):
+    scopes = [scope for scope in scopes for prefix in prefixes if scope.startswith(prefix)]
+    _check_scopes_exist_and_all_have_the_same_prefix(scopes, prefixes)
+    return scopes
+
+
+def _check_scopes_exist_and_all_have_the_same_prefix(scopes, prefixes):
```
ahal, line 59: nit: I think in-lining this to the function above is a bit more readable.

ahal (approved):
Thanks, this looks reasonable to me! I hope this doesn't remove Mihai's review.

ahal:
I'll let you decide whether you want to block on Mihai's review or not. It should still show up in his myqonly notifications.

beetmoverscript/src/beetmoverscript/task.py
ahal: Another slightly more efficient way of doing this check would be to assert the `len` of scopes is the same before and after the filtering.

beetmoverscript/src/beetmoverscript/task.py
```diff
@@ -34,9 +34,39 @@ def validate_task_schema(context):
...
+    for prefix in prefixes:
+        if all(scope.startswith(prefix) for scope in scopes):
+            break
+    else:
+        raise ScriptWorkerTaskException("Scopes must exist and all have the same prefix. " "Given scopes: {}. Allowed prefixes: {}".format(scopes, prefixes))
```
ahal, line 64: nit: Any reason not to use f-strings here? I also tend to like to format lists per-line in error messages, e.g:
```python
scope_str = " - ".join(scopes)
prefix_str = " - ".join(prefixes)
... f"Given scopes:\n  -{scope_str}\n\nAllowed prefixes:\n -{prefix_str}"...
```

though it seems like this probably won't ever happen so, meh. Up to you.

escapewindow:
Re: both review comments: agreed; I mainly copied over [signingscript's code](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/signingscript/src/signingscript/task.py#L105-L126) with minor changes. We might even want to move these helper functions to, say, `scriptworker_client` so we can share them. I'm open to whichever: land as is, address the review comments, move these to `scriptworker_client` and import in both places.

MihaiTabara:
Was chatting with Stefan on adjusting https://github.com/mozilla-mobile/android-components/blob/master/taskcluster/ci/config.yml#L190 and it seemed that we'd only need to land that on `master`, if need be. However, that changes scopes for everything, signing and pushapk included. Tackling the multiple prefixes might remove us from the weeds sooner. Looking at the PR now

beetmoverscript/docker.d/worker.yml
```diff
@@ -4,7 +4,20 @@ aiohttp_max_connections: 5
...
+taskcluster_scope_prefixes:
+  $flatten:
+    $match:
+      'COT_PRODUCT == "firefox"':
+        - 'project:releng:beetmover:'
+      'COT_PRODUCT == "thunderbird"':
+        - 'project:comm:thunderbird:releng:beetmover:'
+      'COT_PRODUCT == "mobile"':
+        - 'project:mobile:android-components:releng:beetmover:'
+        - 'project:mobile:focus:releng:beetmover:'
```
MihaiTabara, line 16: I presume we want `project:mobile:fenix:releng:beetmover` as well here? We'll want Focus APKs published too so ++ on adding Focus too from now.

MihaiTabara:
@escapewindow @ahal for the record, I've noticed we do have a `beetmover-dev-relengworker-firefoxci-mobile-1-1` within our workers so we could also YOLO this in the `dev-beetmoverscript`, provided that ahal's changing his patches in Fenix to point to the `-dex` workes instead of the staging ones.

beetmoverscript/docker.d/worker.yml
MihaiTabara: I assume we also need to add the Fenix corresponding S3 credentials, something like down below, but that's not blocking this PR - I suppose you want to see whether this PR first is backward compatible? :
```
diff --git a/beetmoverscript/docker.d/worker.yml b/beetmoverscript/docker.d/worker.yml
index f06f089..f4e4905 100644
--- a/beetmoverscript/docker.d/worker.yml
+++ b/beetmoverscript/docker.d/worker.yml
@@ -128,6 +128,27 @@ bucket_config:
             nightly_components: 'maven-nightly-stage-s3-upload-bucket'
         dummy: {}
       'COT_PRODUCT == "mobile" && ENV == "prod"':
+        nightly:
+          credentials:
+            id: { "$eval": "FENIX_NIGHTLY_ID" }
+            key: { "$eval": "FENIX_NIGHTLY_KEY" }
+          buckets:
+            fenix: 'net-mozaws-prod-delivery-archive'
+          url_prefix: 'https://archive.mozilla.org'
+        release:
+          credentials:
+            id: { "$eval": "FENIX_RELEASE_ID" }
+            key: { "$eval": "FENIX_RELEASE_KEY" }
+          buckets:
+            fenix: 'net-mozaws-prod-delivery-archive'
+          url_prefix: 'https://archive.mozilla.org'
+        dep:
+          credentials:
+            id: { "$eval": "FENIX_DEP_ID" }
+            key: { "$eval": "FENIX_DEP_KEY" }
+          buckets:
+            fenix: 'net-mozaws-stage-delivery-archive'
+          url_prefix: 'https://ftp.stage.mozaws.net'
         maven-production:
           credentials:
             id: { "$eval": "MAVEN_ID" }
```

escapewindow: > I presume we want `project:mobile:fenix:releng:beetmover` as well here? We'll want Focus APKs published too so ++ on adding Focus too from now.

Oops, good catch :) I mean, I *meant* to proactively add Focus.

ahal:
I thought you were surreptitiously leaving it out to see if I would figure
out to add it later ;)

escapewindow:
Added the fenix prefix, and pushed to dev-beetmoverscript to either give us an error message or create the mobile-1-beetmover-dev pool.

MihaiTabara:
> Added the fenix prefix, and pushed to dev-beetmoverscript to either give us an error message or create the mobile-1-beetmover-dev pool.

Makes sense. Though I think we'd need an actual task scheduled afterwards, given https://github.com/mozilla-releng/k8s-autoscale/blob/master/configs/relengworker-nonprod/config.yml#L28

beetmoverscript/docker.d/worker.yml
escapewindow: > I assume we also need to add the Fenix corresponding S3 credentials, something like down below, but that's not blocking this PR - I suppose you want to see whether this PR first is backward compatible?

Yeah, I wanted to solve the prefix problem here, and then we can add the bucket creds and such in a followup.

escapewindow: Ok, this is on dev-beetmoverscript now, and appears to work. We probably want to add the staging bucket secrets to SOPS and update the worker secrets, and then we can test from staging-fenix. @ahal how much of this do you want to handle? Do you want a chat / writeup / zoom call about this? (@MihaiTabara or I probably need to update SOPS.)

ahal:
I don't mind pushing on the secrets if you think that's useful for me to
know. If so then yes, I'd need some guidance.

 Otherwise I have some work I can do in the meantime.. basically get to the
point where I can see a failure pushing to the bucket due to missing
credentials.

MihaiTabara:
Pushing in a bit my stashed changes for this.

ahal:
Please cc me on the PR/bug (assuming I'm allowed to see it), just so I can
see how it's done.

MihaiTabara:
@escapewindow @ahal I've added a bit more support for this in beetmover. I'm not very happy with the code changes, but they should unblock ahal for now. I have thoughts on how to improve this further but may be good enough for now and tackle in a follow-up more? 

Note: https://gist.github.com/MihaiTabara/594bf50c4000b7909f01db7fa7f1760b is the `task.json` I've used to test these changes.

Note to self:
* fix existing tests
* I still need to add tests for this
* cleanup the schemas, they are a mess right now
* add the corresponding `DEP_ID` and `DEP_KEY` with dummy values for now in SOPS + re-deploy the dev workers

beetmoverscript/docker.d/init_worker.sh
```diff
@@ -59,24 +58,41 @@ case $COT_PRODUCT in
...
-    test_var_set 'MAVEN_NIGHTLY_ID'
-    test_var_set 'MAVEN_NIGHTLY_KEY'
-    export TASKCLUSTER_SCOPE_PREFIX="project:mobile:android-components:releng:${PROJECT_NAME}:"
+    case $ENV in
+      dev|fake-prod)
+        test_var_set 'MAVEN_ID'
+        test_var_set 'MAVEN_KEY'
+        test_var_set 'MAVEN_NIGHTLY_ID'
+        test_var_set 'MAVEN_NIGHTLY_KEY'
+        test_var_set 'DEP_ID'
```
MihaiTabara, line 69: Note to self: I need to add these in SOPS  + redeploy dev workers

beetmoverscript/docker.d/init_worker.sh
```diff
@@ -59,24 +58,41 @@ case $COT_PRODUCT in
...
+        test_var_set 'DEP_KEY'
+        ;;
+      prod)
+        test_var_set 'MAVEN_ID'
+        test_var_set 'MAVEN_KEY'
+        test_var_set 'MAVEN_NIGHTLY_ID'
+        test_var_set 'MAVEN_NIGHTLY_KEY'
+        test_var_set 'DEP_ID'
```
MihaiTabara, line 77: Note to self: I need to add these in SOPS + redeploy dev workers

beetmoverscript/docker.d/worker.yml
MihaiTabara: Added a follow-up commit with missing changes.

beetmoverscript/docker.d/worker.yml
```diff
@@ -127,6 +141,14 @@ bucket_config:
           buckets:
             nightly_components: 'maven-nightly-stage-s3-upload-bucket'
         dummy: {}
+        dep:
```
MihaiTabara, line 144: We're likely getting real credentials only next week but I think for the dev workers, we can totally fake with dummy values in SOPS until then.

beetmoverscript/src/beetmoverscript/data/artifactMap_beetmover_task_schema.json
```diff
@@ -57,12 +57,7 @@
                         }
```
MihaiTabara, line 57: This schema was unused since we switched to declarative artifacts so I cleared it and reused it. I don't like at all how schemas look like currently, it's a mess, but this should unblock for now at least.

beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -109,6 +109,26 @@ async def push_to_nightly(context):
     add_checksums_to_artifacts(context)
 
 
+# direct_push_to_releases {{{1
+async def direct_push_to_releases(context):
```
MihaiTabara, line 113: @escapewindow curious what your thoughts are here. For Fenix, we'd likely push directly into ~releases, so no `candidates` so it didn't make sense to reuse that logic. On the other hand, the `push-to-candidates` also assumes balrog manifest logic + checksums which we don't really need here for now. So I thought adding a new action scope to reflect that since the `push-to-releases` is an existing different snippet.

beetmoverscript/src/beetmoverscript/task.py
MihaiTabara: Note to self: @ahal and I spoke about this and we agreed to ignore for now, file an issue and move these all in `scriptworker_client` to benefit in the future.

docs/scriptworkers-FAQ.md
```diff
@@ -34,15 +34,20 @@ on your local machine:
 find the latest image pushed
 * pull that image locally by `docker pull mozilla/releng-$script:production--$timestamp-$hash`
 * use `pass` to define some dummy values replacing the one it'd expect in production (e.g. a file called `local-prod`)
```
MihaiTabara, line 36: Adjusting a bit the docs since I found it incomplete - https://scriptworker-scripts.readthedocs.io/en/latest/scriptworkers-FAQ.html

beetmoverscript/docker.d/init_worker.sh
escapewindow: We also need to update the tests to populate fake values here.

beetmoverscript/src/beetmoverscript/script.py
escapewindow: I'm ok with that. We may want to somehow list which trust domains support direct pushes, and which ones support candidates->releases, so we don't accidentally do the wrong thing in the wrong project.

beetmoverscript/docker.d/init_worker.sh
MihaiTabara: > Note to self: I need to add these in SOPS + redeploy dev workers

I've populated with fake values the `DEP_ID` and `DEP_KEY` values and redeployed (I foced redeploy to skip the tox blocking dep) `dev-mobile-beetmoverworkers` so that they contain the HEAD of this PR. Locally this works against https://gist.github.com/MihaiTabara/594bf50c4000b7909f01db7fa7f1760b, with the caveat of 401 against S3, of course. I have the production credentials from Jeremy, but he's going to follow-up with the staging ones in the next days too. 

I'll address the rest of the leftovers TODOs from this PR when I'm back from PTO, but for now, this should be enough to unblock @ahal's testing to see a beetmover red task. CC @escapewindow

MihaiTabara: I believe @hneiva has already overwritten the dep keys in SOPS so we should be good on this section.

beetmoverscript/docker.d/init_worker.sh
MihaiTabara: @CC @hneiva I think we'll have to make sure we have these setup properly in SOPS + cloudops-infra. I may have set them to dummy values last time, but if I didn't, we need to make sure there are valid values there.

beetmoverscript/docker.d/worker.yml
MihaiTabara: I think @hneiva deployed this successfully this week so we should be good here.

MihaiTabara: @hneiva It might be worth checking with @sarah-clements as well whether Focus team wants this uploaded to S3 as well in the future? We don't need to block this PR but if we do, it'd be good to double-check the structure so that when that request comes along, it'll be easier to fill in those credentials.

beetmoverscript/docker.d/worker.yml
MihaiTabara: We did, I think we're good here.

beetmoverscript/src/beetmoverscript/script.py
MihaiTabara: @escapewindow Resurrecting this thread - so it sounds like you're thinking of some additional check similar to https://github.com/mozilla-releng/scriptworker/blob/master/src/scriptworker/constants.py#L320 somewhere there but adjusted to cot product / actions?  CC @hneiva

escapewindow: Yes, and/or baking it into the script config, e.g. iscript's [supported behaviors](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/iscript/config_example.yaml#L15-L20) allowlist.

hneiva:
@escapewindow We should close this PR

PR #358, opened 2021-05-21
validate config has scopes

jmaher:


githubscript/src/githubscript/script.py
```diff
@@ -34,6 +34,17 @@ async def async_main(config, task):
 
     release_config = get_release_config(project_config, task["payload"], config)
 
+    # validate we have a matching scope for owner and repo name
+    project_owner = project_config["github_owner"]
+    project_repo = project_config["github_repo_name"]
```
escapewindow: Hm, this won't work when we're using globs (`allow_github_repo_override` is `True`). And that's the only case when we need to verify this.

Maybe
```diff
diff --git a/githubscript/src/githubscript/release_config.py b/githubscript/src/githubscript/release_config.py
index 51c31e54..470ecd2f 100644
--- a/githubscript/src/githubscript/release_config.py
+++ b/githubscript/src/githubscript/release_config.py
@@ -12,18 +12,19 @@ def get_release_config(product_config, task_payload, config):
     # support repo override for xpi-manifest, but support original workflow for fenix
     if product_config.get("allow_github_repo_override", False):
         if not task_payload.get("githubOwner", ""):
             raise TaskVerificationError("missing githubOwner from task")
         if not task_payload.get("githubRepoName", ""):
             raise TaskVerificationError("missing githubRepoName from task")
 
         owner = task_payload["githubOwner"]
         repo_name = task_payload["githubRepoName"]
+        # verify owner/reponame matches product_config["scope_project"] glob
     else:
         if not product_config.get("github_owner", ""):
             raise TaskVerificationError("missing github_owner from config")
         if not product_config.get("github_repo_name", ""):
             raise TaskVerificationError("missing github_repo_name from config")
 
         owner = product_config["github_owner"]
         repo_name = product_config["github_repo_name"]
 
diff --git a/githubscript/src/githubscript/script.py b/githubscript/src/githubscript/script.py
index bee61731..3c0c06f1 100755
--- a/githubscript/src/githubscript/script.py
+++ b/githubscript/src/githubscript/script.py
@@ -25,18 +25,19 @@ async def async_main(config, task):
         project_config = {}
         for p in projects:
             if not p.endswith("*"):
                 continue
             cleaned = p.rstrip("*")
             if project.startswith(cleaned):
                 if project_config != {}:
                     raise TaskError(f'project "{project}" matches multiple configs in "{projects}"')
                 project_config = config["github_projects"][p]
+                project_config.setdefault("scope_project", p)
 
     release_config = get_release_config(project_config, task["payload"], config)
 
     contact_github = bool(release_config.get("contact_github"))
     _warn_contact_github(contact_github)
 
     action = get_action(task, prefix)
     check_action_is_allowed(project_config, action)
     if action == "release":
```

githubscript/src/githubscript/script.py
```diff
@@ -21,6 +21,7 @@ async def async_main(config, task):
     projects = config["github_projects"].keys()
     if project in projects:
         project_config = config["github_projects"][project]
+        project_config.setdefault("scope_project", project)
```
escapewindow, line 24: I don't think we need this; we only need to verify the project matches the scope if we're using a glob.

githubscript/src/githubscript/release_config.py
```diff
@@ -18,6 +18,14 @@ def get_release_config(product_config, task_payload, config):
 
         owner = task_payload["githubOwner"]
         repo_name = task_payload["githubRepoName"]
+        config_owner = product_config.get("github_owner", "")
+        config_repo = product_config.get("github_repo_name", "")
```
escapewindow, line 22: I think these will be `""` since we don't set the owner and repo in `allow_github_repo_override` repos: https://github.com/mozilla-releng/scriptworker-scripts/blob/master/githubscript/docker.d/worker.yml#L45-L49.

This check should instead be something like

```
expected_project_prefix = product_config["scope_project"].rstrip("*")
if not f"{owner}/{repo_name}".startswith(expected_project_prefix):
    raise TaskVerificationError(...)
```

Essentially, `product_config["scope_project"]` will be something like `mozilla-extensions/*` or `mozilla-releng/staging-xpi-*`. `f"{owner}/{repo_name}"` will be something like `mozilla-releng/staging-xpi-public` or `mozilla-extensions/normandy-devtools` or `mozilla-mobile/fenix` if someone's trying to do something sneaky. We need to make sure that the repo we're pointing at from the task definition starts with the `scope_product` minus the final `*`.

If that makes sense to you and @bhearsum, and you can get this fixed up and landed when I'm out, great. If this needs to wait til I get back, that's ok.

escapewindow (changes requested):

escapewindow:
@jmaher is this on your radar to pick up again at some point?
Not sure if you got blocked here, or if it got lost in the noise.

PR #367, opened 2021-06-08
RELENG-587: Add a passwords.yml block for mozillaonline fenix builds

bhearsum:
I think this is actually really straightforward? I opened https://github.com/mozilla-services/cloudops-infra/pull/3181 and pushed a sops change for the needed cloudops parts, but AFAICT all we need to do is add the new certs to the mapping here.

There's not much testing to do against dev, because we don't have a special dep cert for mozillaonline.


bhearsum:
(looks like these failures are mypy related; we'll need some requirements updates to fix them)

escapewindow:
At least some of the failures are `KeyError: 'AUTOGRAPH_FENIX_MOZILLA_ONLINE_USERNAME'`

escapewindow:
And yes, the scriptworker_client and iscript failures are mypy related.
I'm not sure if we're getting much of a win from mypy, and I know we're getting a hit in productivity for it. I'm ok with disabling, but I'm also ok with a quick fix.

signingscript/docker.d/passwords.yml
```diff
@@ -81,7 +81,14 @@ in:
           - ["https://autograph-external.prod.autograph.services.mozaws.net",
              {"$eval": "AUTOGRAPH_FENIX_USERNAME"},
              {"$eval": "AUTOGRAPH_FENIX_PASSWORD"},
-             ["autograph_apk", "autograph_apk_fennec_sha1"]
+             ["autograph_apk"]
+            ]
+        project:mobile:fenix:releng:signing:cert:dep-mozillaonline-signing:
+          - ["https://autograph-external.prod.autograph.services.mozaws.net",
```
escapewindow: Hm. I think we want to keep this `dep-signing`, and change the format name to, say, `autograph_apk_mozillaonline` or similar.

It's up to us if we actually want to create another format for dep, or just change the taskgraph configs to use `dep-signing` + `autograph_apk` for mozillaonline dep signing?

signingscript/docker.d/passwords.yml
```diff
@@ -279,6 +286,12 @@ in:
              {"$eval": "AUTOGRAPH_FENIX_PASSWORD"},
              ["autograph_apk"]
             ]
+        project:mobile:fenix:releng:signing:cert:production-mozillaonline-signing:
+          - ["https://autograph-external.prod.autograph.services.mozaws.net",
+             {"$eval": "AUTOGRAPH_FENIX_MOZILLA_ONLINE_USERNAME"},
+             {"$eval": "AUTOGRAPH_FENIX_MOZILLA_ONLINE_PASSWORD"},
+             ["autograph_apk"]
```
escapewindow: Same here. We do need a separate set of autograph credentials for mozillaonline, so let's keep the `cert:production-signing`, and add a new `autograph_apk_mozillaonline` or similar format.

escapewindow (changes requested):
Let me know if you have questions?
And agreed, this should be pretty quick once we get the cloudops PR merged.

signingscript/docker.d/passwords.yml
escapewindow: (If we did change the cert, we would need to:
- write a ci-config patch to grant the new cert scopes to fenix, and
- update the [CoT restricted scopes](https://github.com/mozilla-releng/scriptworker/blob/master/src/scriptworker/constants.py#L366) to make sure we only use the new production-mozillaonline-signing cert type against the fenix repo)

signingscript/docker.d/passwords.yml
bhearsum: I don't see any reason to use a new cert for dep signing. Is there a reason we need to change the scope name, if we're using the same cert?

signingscript/docker.d/passwords.yml
bhearsum: It sounds like you're saying the reason we need to use the existing certs is to satisfy CoT? If so, that makes sense. (I couldn't find any reason within `signingscript` to require it.)

signingscript/docker.d/passwords.yml
escapewindow: Nope. Maybe just json-e in the signing kind: for depsigning, use `dep-signing` + `autograph_apk` format; if we're release signing, use `production-signing` + `autograph_apk_mozillaonline` or similar.

signingscript/docker.d/passwords.yml
escapewindow: It's mainly for separating out scopes and yes, restricting certain cert scopes to certain repos in CoT. Formats are, I think, just in the upstreamArtifacts and are no longer scopes at all.

signingscript/docker.d/passwords.yml
bhearsum: sgtm; thanks for the pointers.

escapewindow (approved):
LGTM, thank you!

I think we can roll these out at any point.

PR #368, opened 2021-06-10
RELENG-587: add mozillaonline signing format for dep signing

bhearsum:
Turns out we need this for staging releases (they use dep-signing workers), eg: https://firefox-ci-tc.services.mozilla.com/tasks/ceXzy5U9TAyuBaI4wR3cCw/runs/0/logs/public/logs/live_backing.log. The alternative is to change the format in `fenix_taskgraph`, but I think this is more straightforward?


escapewindow (approved):
WFM!

PR #369, opened 2021-06-23
Add support for maxRunTime in notarization_poller (bug 1716793)

jcristau:


escapewindow (approved):
Looks good! We're moving towards f-strings everywhere, e.g. `f" ({dependencies[0]})"` if you want to adjust that. The log `%s` still makes sense to make the string formatting lazy.

escapewindow:
Once this merges, the next step to roll this out will be to update the `scriptworker_scripts_revision` in https://github.com/mozilla-platform-ops/ronin_puppet/blob/master/data/common.yaml#L116-L142 .

Once this is rolled out to the notarization pool, we can land the in-tree patch that adds the `payload.maxRunTime` setting.

escapewindow:
> Once this merges, the next step to roll this out will be to update the `scriptworker_scripts_revision` in https://github.com/mozilla-platform-ops/ronin_puppet/blob/master/data/common.yaml#L116-L142 .
> 
> Once this is rolled out to the notarization pool, we can land the in-tree patch that adds the `payload.maxRunTime` setting.

@jcristau just double-checking - do you want to tackle these steps to cross-train? Or did you want this pushed out sooner?

PR #371, opened 2021-07-14
Run taskgraph generation with Python 3

ahal:


escapewindow:
As mentioned in channel, it looks like:

- we built the k8s-image docker image with dind everywhere previously;
- we switched gecko over to kaniko a while back;
- Tom switched standalone taskgraph to kaniko [almost a year ago](https://hg.mozilla.org/ci/taskgraph/rev/004bada4f170794a94ad43b84dd7562946305d22);
- but we're still trying to use dind in scriptworker-scripts' k8s-image tasks, so its docker-image task uses the old docker:1.6.2 as its FROM;
- and the old docker:1.6.2 image required for dind is so old that kaniko doesn't support pulling it.

We were playing with the idea of switching the k8s-image images over to kaniko previously. We may be forced to either:
- switch to kaniko now, or
- switch back to using dind, build k8s-image, and punt the problem down the road another year.

I prefer the former, but I don't know how much work it will be... anywhere from "fairly straightforward" to "OMG that sucked". The latter isn't a good solution by any means, but if we don't have the time or headspace now, buying time works.

escapewindow:
Filed #372.

ahal:
Without knowing anything about the time involved, blocking this on switching feels like the right call. With the caveat that we don't really need to switch taskgraph right away either (we can leave it pinned to an old revision). So we can buy ourselves a bit of time if we want.

escapewindow:
We may need to remove the k8s-image docker image task, though let's keep the Dockerfile around (just edit taskcluster/ci/docker-image/kind.yml, I think)

escapewindow (approved):

PR #373, opened 2021-07-15
bump to scriptworker 38.2.0

jcristau:
via maintenance/pin.sh.


escapewindow (approved):
Stamping, thank you!

It's good to check the changelogs of various projects before we roll out. Some historically problematic projects have been

aiohttp (possibly all aio* modules)
arrow
immutabledict

I also look at major version bumps. Looking at the first handful of files, the modules to double check appears to include

arrow
certifi
idna
immutabledict
python-jose

I also doublecheck that the thing I wanted to bump (scriptworker 38.2.0 in this case) is the correct version in each file. This is generally only a problem if we run `pin.sh` within a few minutes of uploading the release to pypi.

jcristau:
No aio* changes this time around.

As for the other you mentioned:

- arrow 1.1.0 → 1.1.1: https://github.com/arrow-py/arrow/commit/57332ed039a974ddeee566790d14c0e28cdb3888 sounds reasonably safe
- immutabledict 1.3.0 → 2.0.0: https://github.com/corenting/immutabledict/blob/master/CHANGELOG.md#version-20 "Remove *args, **kwargs from the fromkeys() method", as far as I can tell we don't use that method
- certifi 2020.12.5 → 2021.5.30: https://github.com/certifi/python-certifi/compare/2020.12.05...2021.05.30 updated CA bundle
- idna 2.10 → 3.2: https://github.com/kjd/idna/blob/master/HISTORY.rst#32-2021-05-29 removed support for python2 and python3.4, sounds fine
- python-jose 3.2.0 → 3.3.0: https://github.com/mpdavis/python-jose/blob/master/CHANGELOG.md#330----2021-06-04 removed support for python2.7 and 3.5, add 3.9, removed PyCrypto backend, also seems ok

escapewindow:
Sounds good, thank you for the doublecheck!

PR #375, opened 2021-07-19
Disable the k8s-image task

ahal:
We pinned the k8s-image to one in docker hub so the image build here is
no longer being used. We'll need to upgrade this image to kaniko and
figure out how to distribute it to cloudops-jenkins.

Commenting out the task rather than removing so we remember why we
still have all the associated Dockerfiles lying around.


escapewindow (approved):

escapewindow:
Thank you!

PR #377, opened 2021-07-23
WIP - add update-{rules,releases} balrog behaviors

escapewindow:
For System Addons, Widevine, and custom release workflows, we sometimes want more granular behaviors than is currently supported in the current balrog submitter cli.

Let's add support for `update-{rules,releases}` behaviors, make sure there are sufficient protections around these behaviors, and roll out to the appropriate pool(s).


escapewindow:
@bhearsum if you have cycles, let me know if I should change direction here?
Currently planning on testing the `update_rules` behavior against my local balrog, and then implementing `update-releases` behavior, and seeing how those two things might work for their intended products.

escapewindow:
I'm not sure if `balrogclient` supports the creation of new rules, except directly hitting the API?
Same with the deletion of rules.

balrogscript/src/balrogscript/data/balrog_update-releases_schema.json
```diff
@@ -0,0 +1,55 @@
...
+            "minItems": 1,
+            "uniqueItems": true,
+            "items": {
+                "type": "string"
+            }
+        },
+        "payload": {
+            "type": "object",
+            "properties": {
+                "behavior": {"type": "string", "enum": ["update-rules"], "default": "submit-locale"},
```
bhearsum: this looks wrong

bhearsum (commented):
Are you planning to do the heavy lifting of figuring out what actually needs to change here, or just take a list of changes, and apply them? I'm guessing the latter based on the payloads, which seems smart to me.

> I'm not sure if `balrogclient` supports the creation of new rules, except directly hitting the API?
> Same with the deletion of rules.

This is correct. We sortof moved away from mirroring the backend models in the client anyways - so it's totally fine to use the API directly. If you'd prefer `balrogclient.Rule` to support it, I'm fine with that too.

balrogscript/src/balrogscript/data/balrog_update-releases_schema.json
escapewindow: Oops :)

escapewindow:
> Are you planning to do the heavy lifting of figuring out what actually needs to change here, or just take a list of changes, and apply them? I'm guessing the latter based on the payloads, which seems smart to me.

Yes. My main concern is it may provide a bit too much flexibility, and we need to be careful that we don't allow for messing up Firefox rules and releases in System Addons automation, etc. I may put in a product and/or channel allowlist.

> > I'm not sure if `balrogclient` supports the creation of new rules, except directly hitting the API?
> > Same with the deletion of rules.
> 
> This is correct. We sortof moved away from mirroring the backend models in the client anyways - so it's totally fine to use the API directly. If you'd prefer `balrogclient.Rule` to support it, I'm fine with that too.

Sounds good, thank you!

PR #381, opened 2021-08-04
Update version-control-tools to current version (for firefoxtree).

jfx2006:
version-control-tools in the Docker image is from 2019, and the
firefoxtree extension is not aware of comm-esr91. Thunderbird does not
currently have a "unified" repository, so to perform merge-day
actions with comm-esr91 an update is needed.

An unintended side effect is that robustcheckout is also updated.


escapewindow (approved):

escapewindow:
Rolling out now.

PR #386, opened 2021-08-24
Add Android signers to README

bhearsum:


signingscript/README.md
```diff
@@ -5,6 +5,74 @@ Signingscript
 
 This is designed to be run from scriptworker, but runs perfectly fine as a standalone script.
 
+Available Signers & Keys
+------------------------
+
+This is a best effort list of known signers and their keyids. Ones in active use end up in [https://github.com/mozilla-releng/scriptworker-scripts/blob/master/signingscript/docker.d/passwords.yml](passwords.yml) and the sops repo used to deploy signingscript.
```
escapewindow, line 13: I know `git blame` shows this, but we may want a `last updated` date so we have some idea of how stale this info might be.

escapewindow (approved):

signingscript/README.md
bhearsum: Yeah, that's a good idea.

PR #389, opened 2021-08-30
Replace mpd001 workers with mozillavpn; drop EV certificate support

bhearsum:


signingscript/docker.d/init_worker.sh
```diff
@@ -56,11 +56,7 @@ case $COT_PRODUCT in
     ;;
   xpi)
     ;;
-  mpd001)
-    export AUTHENTICODE_CERT_PATH=/app/signingscript/src/signingscript/data/authenticode_mpd_dep.crt
```
bhearsum, line 60: These overrides were for the EV certificate.

escapewindow (approved):
Nice, thanks for the cleanup!

MihaiTabara:
Does this mean that our current pools for signing the Windows builds (in the current old way, non-taskgraph) are now renamed to `mozillavpn`? Asking as I know that infra was fragile and want to make sure we validated a VPN release since then to confirm the pools are still working fine.

bhearsum:
We ended up reviving the `mpd001` level 3 windows pool for the current release process, actually. But eventually, yes, this will fully replace it.

PR #391, opened 2021-09-02
Parallelize pin.sh

bhearsum:
4 processes seemed like a good default for most places this will run (laptops).

### DIFF
--- a/maintenance/pin.sh
+++ b/maintenance/pin.sh
@@ -25,5 +25,5 @@ else
     "
 fi
 
-docker run -t -v $PWD:/src -w /src python:3.8 maintenance/pin-helper.sh $DIRS
-docker run -t -v $PWD:/src -e SUFFIX=py37.txt -w /src python:3.7 maintenance/pin-helper.sh $DIRS
+echo $DIRS | xargs -n4 -P4 docker run -t -v $PWD:/src -w /src python:3.8 maintenance/pin-helper.sh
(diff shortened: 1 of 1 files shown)


escapewindow (approved):
WFM :)
Non-blocking thoughts:
- I wonder what the logs look like if something breaks? Would we be able to tell which script's dependencies are having issues?
- If there's some way to `| tee logs/signingscript.py38.log` or similar, we might be able to differentiate the issues. not sure if that's plausible without another helper script.
- I wonder if Joel will have issues on Windows?
- If we reduce the number of scripts to <4, does this still work properly? I think so, but haven't verified.

bhearsum:
> WFM :)
> Non-blocking thoughts:
> 
>     * I wonder what the logs look like if something breaks? Would we be able to tell which script's dependencies are having issues?
> 
>     * If there's some way to `| tee logs/signingscript.py38.log` or similar, we might be able to differentiate the issues. not sure if that's plausible without another helper script.
> 
>     * I wonder if Joel will have issues on Windows?
>     * If we reduce the number of scripts to <4, does this still work properly? I think so, but haven't verified.

I tested this - it works fine. We'll probably want to bump `-n` higher if we end up with more than 4*4 scriptworkers (otherwise we get an additional run _after_ all of the other ones), but that's pretty unlikely.

I'm going to punt on the other things if/until they become issues.

PR #392, opened 2021-09-02
bug 1728330 - stop pushing snap to the snap store

escapewindow:
We don't want to upload to the store at all.

This PR is to disable store uploads for 92, which is already in-progress. Patches coming in bug 1728330 to remove the tasks and delete the pools.

@mkaply


mkaply:
We shouldn't upload to the store at all.

bhearsum (approved):

escapewindow:
I imagine we'll delete these pools shortly.

PR #393, opened 2021-09-02
bug 1728330 - Delete pushsnap

escapewindow:


bhearsum (approved):
We should probably cleanup `cloudops-infra` and the associated sops repo, too.

escapewindow:
Blocked on esr78 EOL.

jmaher:
I think we can merge this now?

escapewindow:
Let's land https://github.com/mozilla-services/cloudops-infra/pull/3525 first, so we don't deploy any new pushsnap k8s pools; then we can clean up code and configs at will.

PR #395, opened 2021-09-07
Fix claimWork for notarization poller

bhearsum:
TIL that notarization poller doesn't use `scriptworker`


escapewindow (approved):
Yup, it's its own worker, because it needs to claim `n` tasks in parallel, and scriptworker doesn't/shouldn't support that.

MihaiTabara:
TIL too.

PR #400, opened 2021-09-13
Add 'focus-android' scope prefix to signingscript

ahal:


escapewindow (changes requested):
We probably also need to add the section in `docker.d/passwords.yml`.

signingscript/docker.d/passwords.yml
```diff
@@ -3,7 +3,6 @@ $let:
     $match:
       'COT_PRODUCT == "firefox"': 'project:releng:signing:'
       'COT_PRODUCT == "thunderbird"': 'project:comm:thunderbird:releng:signing:'
-      'COT_PRODUCT == "mobile"': 'project:mobile:focus:releng:signing:'
```
ahal, line 6: This caused me some confusion as to why we were matching the entire `mobile` trust domain to just `focus`. But on closer inspection I noticed that this `scope_prefix` is never being used where `COT_PRODUCT == "mobile"`. So I think it's safe to remove and should avoid some confusion.. though I can revert if you like.

escapewindow (approved):
lgtm!

PR #402, opened 2021-09-14
Fix incorrect signing format for focus-android in signingscript

ahal:


ahal:
The error I'm seeing might be a task misconfiguration. Marking draft for now and may close invalid later.

MihaiTabara:
Superseeded by https://github.com/mozilla-releng/scriptworker-scripts/pull/400/ ?

ahal:
Nah, this was a follow-up to #400 but I think it turns out that the configuration here was correct and it was the task that needed to change.

PR #404, opened 2021-09-23
balrogscript: don't use the prod instance for "dep"

jcristau:
When a task has the balrog:server:dep scope we shouldn't talk to the
production instance, even on a production worker.


escapewindow (approved):
Let's push this to `production-balrogscript` after we merge to `master`. I can do this and/or watch the deploy if it's your EOD.

escapewindow:
This will probably break `dep` behavior on prod, which is preferable to the status quo.
As a followup, we may want to customize the configs, via json-e or other, so we get just the dep block on dep pools, and we don't get the dep block on production pools.

PR #408, opened 2021-09-28
Bug 1732888 - add new mime types for beetmoverscript

jcristau:
https://bugzilla.mozilla.org/show_bug.cgi?id=1732727 added a new ".module" file
for geckoview, which appears to be json, as well as new checksums.


bhearsum (approved):
This looks fine. https://github.com/mozilla-releng/scriptworker-scripts/blob/master/beetmoverscript/tests/test_script.py#L127 should probably be updated (and maybe parametrized?) -- but that can be done as a follow-up.

agi:
Thank you so much @jcristau !

PR #416, opened 2021-09-30
[bouncerscript] add a set of products with and without msix, to handl…

bhearsum:
…e overlap period


escapewindow (approved):
Looks like we're adding another set to keep this check valid for existing nightlies. https://github.com/mozilla-releng/scriptworker-scripts/blob/a5e2e0ce389350026acdfbdfa04a8ba3ff7ebee0/bouncerscript/src/bouncerscript/task.py#L113-L116 r+

bhearsum:
> Looks like we're adding another set to keep this check valid for existing nightlies.
> 
> https://github.com/mozilla-releng/scriptworker-scripts/blob/a5e2e0ce389350026acdfbdfa04a8ba3ff7ebee0/bouncerscript/src/bouncerscript/task.py#L113-L116
> r+

Yes, exactly!

PR #418, opened 2021-10-01
Parameterize similar tests in test_release_config.py

### LINKED ISSUE #352: [githubscript] use pytest parameterize to reduce test configurations in test_release_config.py and test_script.py
I have added some new tests to githubscript for some changes made in https://github.com/mozilla-releng/scriptworker-scripts/pull/350, but these could be simplified via pytest parameters, spending a day cleaning these up would be beneficial- especially as we start looking at releng for mozilla.

gbrownmozilla:
Closes #352 

This should be functionally identical before/after. I verified that tests still run and coverage remains the same.


escapewindow (approved):
Nice, thank you!

PR #421, opened 2021-10-04
fix bouncer sub

escapewindow:
Looks like we use `multi` instead of `:lang` for msix -- let's add that
in for devedition + pinebuild.


gbrownmozilla (approved):

bhearsum:
Ack, thanks for the bustage fix.

PR #422, opened 2021-10-08
replace docker tags with corresponding digests

### LINKED ISSUE #396: pin Dockerfile `FROM`s to digests
We point our Dockerfiles at docker hub tags, but those tags can move at any point, and we may hit unexpected bustage. Recently, the `python:3.8` docker tag moved to an image with an incompatible `osslsigncode`, which busted signingscript.

Ben suggested we [pin our Dockerfiles to digests](https://docs.renovatebot.com/docker/#digest-pinning) and explicitly bump them. Renovatebot can bump these (though we may have to do so manually if we specify the digest in an env var). This way we can explicitly decide when to roll out a new base image, after we've tested it for bustage.

gbrownmozilla:
This updates all of the Dockerfiles with FROM lines using constant tags. There remain several Dockerfiles that take an argument - those are left unchanged.

### DIFF
--- a/addonscript/Dockerfile
+++ b/addonscript/Dockerfile
@@ -1,4 +1,5 @@
-FROM python:3.8
+# python:3.8
+FROM python@sha256:037c262134bc0cffa81606421308eb0dbf5c851c2328ec585983f8c18553966f
 
 RUN groupadd --gid 10001 app && \
     useradd -g app --uid 10001 --shell /usr/sbin/nologin --create-home --home-dir /app app
--- a/addonscript/Dockerfile
+++ b/addonscript/Dockerfile
@@ -1,5 +1,4 @@
-# python:3.8
-FROM python@sha256:037c262134bc0cffa81606421308eb0dbf5c851c2328ec585983f8c18553966f
+FROM python:3.8@sha256:037c262134bc0cffa81606421308eb0dbf5c851c2328ec585983f8c18553966f
 
 RUN groupadd --gid 10001 app && \
     useradd -g app --uid 10001 --shell /usr/sbin/nologin --create-home --home-dir /app app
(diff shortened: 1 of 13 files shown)


addonscript/Dockerfile
```diff
@@ -1,4 +1,5 @@
-FROM python:3.8
+# python:3.8
+FROM python@sha256:037c262134bc0cffa81606421308eb0dbf5c851c2328ec585983f8c18553966f
```
escapewindow: I wonder if we need something like `python:3.8@sha256:037c262134bc0cffa81606421308eb0dbf5c851c2328ec585983f8c18553966f`

escapewindow:
`INFO[0001] Invalid repository name (python:3.8), only [a-z0-9-_.] are allowed`

Ah, I wonder if [our version of] kaniko doesn't allow for hash pinning, in which case we might not be able to do this.

gbrownmozilla:
fwiw, that format is fine locally,:
```
$ docker run -t -i python:3.8@sha256:037c262134bc0cffa81606421308eb0dbf5c851c2328ec585983f8c18553966f
Python 3.8.12 (default, Sep 28 2021, 19:06:31) 
[GCC 10.2.1 20210110] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 
```

escapewindow:
Yeah. We use [kaniko](https://github.com/GoogleContainerTools/kaniko) to build our docker images. I believe this is both more secure than running directly on docker, and avoids the whole docker-in-docker issue. But it doesn't support everything that docker supports.

escapewindow:
Hm, https://github.com/GoogleContainerTools/kaniko/blob/0477900febfacb98b88abdbfaf8bbaeb05f51060/integration/dockerfiles/Dockerfile_test_volume_3#L4 looks promising.
I wonder if it's supported in a newer version of kaniko that we're not using yet.

Issue #372, opened 2021-07-14
switch k8s-image tasks to kaniko


escapewindow:
(More details in https://github.com/mozilla-releng/scriptworker-scripts/pull/371#issuecomment-880254355.)

We're using kaniko to build our docker images in gecko and other standalone taskgraph docker-image tasks.
However, in scriptworker-scripts, we're still using the old dind method, which requires an ancient docker version. Kaniko is unable to pull this old docker image.

At some point we should switch the k8s-image tasks (and docker image) to use kaniko to conform. I'm not sure if we'll have to do anything special to also upload to docker hub, since gecko and standalone taskgraph docker-image tasks don't upload to docker hub, aiui.

escapewindow:
We got some more runway with #374 .

PR #428, opened 2021-10-21
upgrade python environment to 3.9, remove 3.7

jmaher:
…r 3.10


addonscript/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.8
+FROM python:3.10.0
```
escapewindow: I don't love that we're hardcoding this here, though this may be the best case until we resolve #372.
The canonical way would be e.g. to [set PYTHON_VERSION](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/taskcluster/ci/docker-image/kind.yml#L19) and [parent](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/taskcluster/ci/docker-image/kind.yml#L26) and using

```
ARG DOCKER_IMAGE_PARENT
FROM $DOCKER_IMAGE_PARENT
```

like we do in e.g. [pushsnapscript](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/taskcluster/docker/pushsnapscript/Dockerfile), but I'm not 100% sure if that's tied into the k8s-image task kind.

We should probably hardcode whatever version of python we're running in prod for each one, so I'm guessing that would be either a version of python3.8 or a python3.9. (Be aware that if we upgrade the entire set of scriptworkers to python 3.9 now, there will be more testing than just signingscript.) We may want to push all of these changes to `dev` and run a `dev` staging release at the end of this :)

addonscript/requirements/base.py37.txt
```diff
@@ -1,415 +0,0 @@
-# SHA1:c2cac962ccf2991d069e31c6e916c06a77d42985
```
escapewindow, line 1: If we're removing py37 support, we can probably remove [importlib-metadata](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/addonscript/requirements/base.in#L4-L6).

addonscript/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.black]
 line-length = 160
-target-version = ["py37", "py38"]
+target-version = ["py38", "py39", "py310"]
```
escapewindow: guessing we're going to remove py310 for now.

escapewindow (commented):
Not sure what's going to happen to these review comments...

addonscript/requirements/base.txt
```diff
@@ -51,9 +51,9 @@ aiomemoizettl==0.0.3 \
     --hash=sha256:07a6becac60f6cd2604b9f2b73bcd9a50079a0b7b55e2a4e45b1eec5a3ea9659 \
```
escapewindow, line 51: I'd love for all the .txt files to have the py3x as part of the name for consistency, but that doesn't have to happen in this PR.

An observation that doesn't require action: our requirements.txt files are the same!
The local and test files are different. I wonder if that means we can someday get down to one set of requirements files though.

```
43500 37 base.py38.txt
43500 37 base.txt
58672 2 local.py38.txt
28338 2 local.txt
40641 14 test.py38.txt
36194 14 test.txt
```

addonscript/tox.ini
```diff
@@ -34,7 +34,7 @@ depends =
 [testenv:report]
 skip_install = true
 commands = coverage report -m
-depends = py38
+depends = py39
```
escapewindow, line 37: We probably want to also update the top level [tox.ini](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/tox.ini#L51) to move the `check` target to the py39 (or whatever is the shipping python minor version) per script as well.

escapewindow (commented):
Nice, this is looking good, thank you!

addonscript/requirements/base.txt
jmaher: I tried this out, it seemed possible until I got to `pip-compile-multi verify`.  Here is it expecting base.in -> base.txt, and there is no option to specify the outfile, it is derived from the .in script.

addonscript/pyproject.toml
jmaher: yeah, 310 is not supported in black, a bit too early.

addonscript/Dockerfile
jmaher: I thought I could do this and fiddled around a lot with this- in the end we are running docker 1.6.2 for the k8s-images, and in order for the ability to pass in --build-arg <ARG>, we need docker 1.9.  For our test images used in tox, etc. we are ok and that is why there is prior art in the repo with PYTHON_VERSION, etc.

addonscript/tox.ini
jmaher: thanks for mentioning this- my latest changes have this!

addonscript/README.rst
```diff
@@ -9,17 +9,45 @@ a standalone script.
...
-    pandoc --from=markdown --to=rst README.md > README.rst
+   pandoc --from=markdown --to=rst README.md > README.rst
 
 This is purely because
 
 1. @Callek prefers writing markdown, and
 2. pypi appears to deal with rst better than markdown.
 
+Update python dependencies
+--------------------------
```
escapewindow, line 27: Looks like we could duplicate this doc in a lot of places -- maybe link to a top level README.rst?
Non-blocking.

addonscript/requirements/base.txt
```diff
@@ -68,89 +68,97 @@ attrs==21.2.0 \
...
-    --hash=sha256:94cc5ed4ceaefcbe5bf38c8fba6a21fc1d365bb8fb826ea1688e3370b2e24a1c \
-    --hash=sha256:94fff993ee9bc1b2440d3b7243d488c6a3d9724cc2b09cdb297f6a886d040ef7 \
-    --hash=sha256:9965c46c674ba8cc572bc09a03f4c649292ee73e1b683adb1ce81e82e9a6a0fb \
-    --hash=sha256:a00cf305f07b26c351d8d4e1af84ad7501eca8a342dedf24a7acb0e7b7406e14 \
-    --hash=sha256:a305600e7a6b7b855cd798e00278161b681ad6e9b7eca94c721d5f588ab212af \
-    --hash=sha256:cd65b60cfe004790c795cc35f272e41a3df4631e2fb6b35aa7ac6ef2859d554e \
-    --hash=sha256:d2a6e5ef66503da51d2110edf6c403dc6b494cc0082f85db12f54e9c5d4c3ec5 \
-    --hash=sha256:d9ec0e67a14f9d1d48dd87a2531009a9b251c02ea42851c060b25c782516ff06 \
-    --hash=sha256:f44d141b8c4ea5eb4dbc9b3ad992d45580c1d22bf5e24363f2fbf50c2d7ae8a7
+cryptography==35.0.0 \
```
escapewindow, line 135: I'm kind of leaning towards landing these changes post-mergeday next week, so we can YOLO a little bit.

iscript/README.md
```diff
@@ -4,14 +4,6 @@ This is designed to be run from scriptworker, but runs perfectly fine as a stand
 
 # Update python dependencies
 
-For python version we use in production:
+The easiest way to do this is to run `pin.sh`:
```
escapewindow: This is actually inaccurate for rolling out dependency updates to production iscript and `notarization_poller`. For these two, we use the script / requirements files [here](https://github.com/mozilla-platform-ops/ronin_puppet/tree/master/modules/signing_worker/files), and the `scriptworker{,_scripts}_revision` [here](https://github.com/mozilla-platform-ops/ronin_puppet/blob/master/data/common.yaml#L125-L143). And I believe we'll keep using python 3.8 for these two for the time being (this is due to the python that's baked into the mac puppet configs, not anything you need to change here).

If you want to update these two READMEs to be accurate, that would be awesome, otherwise we're not any more inaccurate than we were before :)

pushsnapscript/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM python:3.8
+FROM python:3.9.7
```
escapewindow, line 1: One of these push* scripts may have some tricky bits with which OS version we use. I'm not sure which one, and these are trickier to test because of the external stores. I'm hoping it's pushsnap, because we're going to [EOL it](https://mozilla-hub.atlassian.net/browse/RELENG-688).

taskcluster/ci/docker-image/kind.yml
```diff
@@ -13,29 +13,31 @@ transforms:
...
     # k8s-image:
     #     definition: k8s-image
-    python38:
+    python39:
         definition: python
         args:
-            PYTHON_VERSION: "3.8"
-    python37:
+            PYTHON_VERSION: "3.9.7"
+            PYTHON_REQ_SUFFIX: ".py39"
```
escapewindow: Hm, is this right? I see base.txt not base.py39.txt ?

escapewindow (changes requested):
Wow!

First of all, thank you! Amazing work.
Second, it looks like the tests are green here. I have a couple questions and comments in the review. Mostly minor; I'm wondering about the py suffix comment; I'm happy to r+ once we verify what that should be.

Third, how do you want to roll out? I could see:
- wait until after we ship 94? next week, then YOLO out to `production`, and debug everything. There's a chance we'll have to scramble and fix something or back out.
- push to `dev` before rolling out, and run a staging release using all dev scriptworkers. This won't catch everything, but it'll catch at least some issues. This is more yak shaving, but we'll have more confidence when we roll out.
-  roll out to `production-*` one by one as time allows. We could combine this with the staging release option above for even more confidence. The staging release + rolling out one by one is probably the slowest calendar-time wise, but has the highest chance of catching issues.

Do any of those appeal to you more? I'm open to helping you with any of those.

Either way, we probably want to communicate with the team and #releaseduty about these rollouts, and we may want to resurrect the [releng changelog](https://docs.mozilla-releng.net/en/latest/releng_changelog.html) to note when we upgraded the scriptworkers to py39 in case we need to debug things later.

addonscript/README.rst
jmaher: we have this in the top level doc, I would like to update each doc with some testing notes as well, maybe I will save this cleanup for docs later this week.

iscript/README.md
jmaher: fixed this!

iscript/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.black]
 line-length = 160
-target-version = ["py37", "py38"]
+target-version = ["py38"]
```
escapewindow, line 3: Ah, it might be nice to test py39 in iscript even if we don't upgrade yet :)

escapewindow:
The changes look good, though I do wish we could have kept the iscript/notarization_poller py39 tests, just updating the readme to note that we're still on py38 until we upgrade ronin-puppet.

Even though you force-pushed, you can see the diff via something like `git diff 3607de643637d9b2706ab1b83719a09009f2f083 bd2fafe7afe45440eca0cdd6a715066d065890bf`. You can find those revisions above in this PR, or in the reflog, e.g. `.git/logs/refs/remotes/jmaher/py39` (your remote may be named something other than `jmaher`, maybe `origin`)

iscript/pyproject.toml
jmaher: done!

taskcluster/ci/docker-image/kind.yml
jmaher: good catch, I fixed that

iscript/tox.ini
```diff
@@ -8,8 +8,9 @@ usedevelop = false
 depends =
 skip_install = true
 commands =
-    docker build --build-arg PYTHON_VERSION=3.7 --build-arg PYTHON_REQ_SUFFIX=.py37 -t iscript-{envname}-py37-test -f Dockerfile.test .
-    docker run --rm -v {toxinidir}/..:/app -v iscript-{envname}-py37-tox:/app/.tox -w /app/iscript iscript-{envname}-py37-test py37
+# issue with plistlib not having readPList: https://docs.python.org/3.9/whatsnew/3.9.html#removed
+#    docker build --build-arg PYTHON_VERSION=3.9.7 --build-arg PYTHON_REQ_SUFFIX=.py39 -t iscript-{envname}-py39-test -f Dockerfile.test .
```
escapewindow, line 12: Looks like we have PYTHON_REQ_SUFFIX=.py39 again :)

tox.ini
```diff
@@ -48,163 +47,164 @@ passenv =
...
-    tox -e py38
-
-[testenv:iscript-py37]
-changedir = {toxinidir}/iscript
-commands =
-    tox -e py37,check
+    tox -e py38,check
     - tox -e coveralls
 
+#[testenv:iscript-py39]
```
escapewindow, line 108: Do we want to uncomment this?

escapewindow (commented):
The latest diff (`git diff bd2fafe7 1f2e8d71`) looks good except for the spots I commented, thank you!

Do you want help testing on dev, or are you set?

tox.ini
jmaher: no, there is an issue with plistlib:
plist removed: https://docs.python.org/3.9/whatsnew/3.9.html#removed
we see the failure failure:
```
 E           AttributeError: <module 'plistlib' from '/usr/local/lib/python3.9/plistlib.py'> does not have the attribute 'readPlist'
```
/usr/local/lib/python3.9/unittest/mock.py:1378: AttributeError

iscript/tox.ini
jmaher: yes, that is for iscript, I left it in on purpose this time.

escapewindow: Ok. I'll go ahead and approve, and whatever we have left that needs fixing up we can do in followup PRs.
Let's not land this til after we ship 94.0, though.

escapewindow (approved):

PR #432, opened 2021-11-03
add docs for testing scriptworker-scripts

jmaher:


MihaiTabara (commented):
Thanks for adding this. It looks great! One tiny nit, not sure if the rst are rendering properly, I tried looking at the Rich diff and they look a bit funky. Do you think it's worth fixing that to ease the reading or it may be the Github rendering in the diff problematic?

jmaher:
these are rendered on my branch:
https://github.com/jmaher/scriptworker-scripts/tree/testing_docs

MihaiTabara (approved):
Looks great, thank you!

PR #433, opened 2021-11-10
Add pushmsixscript

gbrownmozilla:


pushmsixscript/src/pushmsixscript/script.py
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env python3
+""" Push msix Script main script
+"""
+import logging
+import os
+
+from scriptworker import client
```
escapewindow, line 7: Ideally we'll use [`scriptworker_client`](https://github.com/mozilla-releng/scriptworker-scripts/tree/master/scriptworker_client) instead of importing scriptworker. There are still some scripts that do the latter, but the former is encouraged, and if we are ever able to prioritize it, we'll complete that transition.

pushmsixscript/src/pushmsixscript/task.py
```diff
@@ -0,0 +1,26 @@
+from scriptworker.exceptions import TaskVerificationError
```
escapewindow, line 1: [`scriptworker_client.exceptions`](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/scriptworker_client/src/scriptworker_client/exceptions.py#L48)

pushmsixscript/tests/integration/test_integration_script.py
```diff
@@ -0,0 +1,135 @@
+"""
+import json
+import os
+import tempfile
+from unittest.mock import MagicMock
+
+import pytest
+from scriptworker.utils import makedirs
```
escapewindow, line 8: [scriptworker_client.utils](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/scriptworker_client/src/scriptworker_client/utils.py#L303)

pushmsixscript/src/pushmsixscript/artifacts.py
```diff
@@ -0,0 +1,17 @@
+from scriptworker import artifacts
+from scriptworker.exceptions import TaskVerificationError
+from scriptworker.utils import get_single_item_from_sequence
+
+
+def get_msix_file_path(context):
+    artifacts_per_task_id, _ = artifacts.get_upstream_artifacts_full_paths_per_task_id(context)
```
escapewindow, line 7: We don't have this in `scriptworker_client` yet. It's possible we don't need it, and it's possible we should port it over.

gbrownmozilla:
Let's abandon this PR -- new one coming soon.

PR #434, opened 2021-11-13
Bug 1490598 - [addonscript] Scope, server configuration for Thunderbird

### LINKED ISSUE #264: [addonscript] Langpack upload support to addons.thunderbird.net
ATN is expected to support automatic language pack uploads soon with release-push-langpacks. 

I suspect this will look a lot like the support that Treescript has for working with the Thunderbird HG repos.

@justdave will be working on this.

jfx2006:
Fixes #264.

### DIFF
--- a/addonscript/docker.d/init_worker.sh
+++ b/addonscript/docker.d/init_worker.sh
@@ -15,21 +15,37 @@ test_var_set 'JWT_USER'
 test_var_set 'JWT_SECRET'
 # == END:   this is what we need to configure ==
 
-case $ENV in
-  dev|fake-prod)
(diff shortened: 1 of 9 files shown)


escapewindow:
We restrict the fx scope [here](https://github.com/mozilla-releng/scriptworker/blob/b667e3ffc0c20183b9cffe90b5c59f456e73df64/src/scriptworker/constants.py#L326); let's do something similar with the comm scope before rolling out.

jfx2006:
This is not needed anymore. The differences between AOM and ATN are too much to really support both.

PR #437, opened 2021-11-17
Beetmover prefixes rebased

hneiva:
Rebased #355


beetmoverscript/docker.d/init_worker.sh
```diff
@@ -59,24 +58,41 @@ case $COT_PRODUCT in
...
+        test_var_set 'MAVEN_ID'
+        test_var_set 'MAVEN_KEY'
+        test_var_set 'MAVEN_NIGHTLY_ID'
+        test_var_set 'MAVEN_NIGHTLY_KEY'
+        test_var_set 'DEP_ID'
+        test_var_set 'DEP_KEY'
+        test_var_set 'NIGHTLY_ID'
+        test_var_set 'NIGHTLY_KEY'
+        test_var_set 'RELEASE_ID'
+        test_var_set 'RELEASE_KEY'
```
MihaiTabara, line 82: We should double-check that we export these values properly in relengworker SOPS for the production ones. Otherwise, if we don't, when we land this to the production workers, the GCP workers will barf.

beetmoverscript/docker.d/init_worker.sh
```diff
@@ -59,24 +58,41 @@ case $COT_PRODUCT in
...
-    test_var_set 'MAVEN_NIGHTLY_KEY'
-    export TASKCLUSTER_SCOPE_PREFIX="project:mobile:android-components:releng:${PROJECT_NAME}:"
+    case $ENV in
+      dev|fake-prod)
+        test_var_set 'MAVEN_ID'
+        test_var_set 'MAVEN_KEY'
+        test_var_set 'MAVEN_NIGHTLY_ID'
+        test_var_set 'MAVEN_NIGHTLY_KEY'
+        test_var_set 'DEP_ID'
+        test_var_set 'DEP_KEY'
```
MihaiTabara, line 70: We should double-check we export these values properly, even though failing GCP dev/staging workers are probably less of a problem.

beetmoverscript/docker.d/worker.yml
```diff
@@ -4,7 +4,21 @@ aiohttp_max_connections: 5
...
+taskcluster_scope_prefixes:
+  $flatten:
+    $match:
+      'COT_PRODUCT == "firefox"':
+        - 'project:releng:beetmover:'
+      'COT_PRODUCT == "thunderbird"':
+        - 'project:comm:thunderbird:releng:beetmover:'
+      'COT_PRODUCT == "mobile"':
+        - 'project:mobile:android-components:releng:beetmover:'
+        - 'project:mobile:focus:releng:beetmover:'
```
MihaiTabara: it's nice that we've included focus too. we may need that at some point too in the future 👍

beetmoverscript/docker.d/worker.yml
```diff
@@ -144,6 +166,30 @@ bucket_config:
             key: { "$eval": "MAVEN_NIGHTLY_KEY" }
           buckets:
             nightly_components: 'maven-nightly-s3-upload-bucket-d4zm9oo354qe'
+        nightly:
```
MihaiTabara, line 186: This whole section looks great!

beetmoverscript/src/beetmoverscript/constants.py
```diff
@@ -78,6 +78,8 @@
 
 RELEASE_ACTIONS = ("push-to-releases",)
 
+DIRECT_RELEASE_ACTIONS = ("direct-push-to-releases",)
```
MihaiTabara: I'm not very happy about the way I named this function since we're going to reuse this for Fenix nightlies too. But after all it's just naming thing since the logic is dictated in the task payload. Though we may want to revisit and s/direct-push-to-releases/direct-push-to-bucket` or alike so make sure it's not confusing us in the future.

beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -109,6 +109,26 @@ async def push_to_nightly(context):
     add_checksums_to_artifacts(context)
 
 
+# direct_push_to_releases {{{1
```
MihaiTabara: We may want to rename this function to direct_push_to_bucket at some point since we're going to use it not only to push Fenix into ~releases direcrtly, but also in ~nightly buckets. Not blocking for now, but may help with confusion in the future.

MihaiTabara (approved):
This LGTM to me modulo two things:
* make sure we run successfully a nightly staging release - I have some concerns around the schema of the task payload that might not match. Also make sure the beetmover task scopes are working as expected
* confirm SOPS secrets are well defined so that we don't break the production workers

beetmoverscript/docker.d/init_worker.sh
escapewindow: I think these are exported for all beetmovers: https://github.com/mozilla-services/cloudops-infra/blob/master/projects/relengworker/k8s/charts/beetmover/templates/configmap.yaml#L12-L24 and https://github.com/mozilla-services/cloudops-infra/blob/master/projects/relengworker/k8s/charts/beetmover/templates/secret.yaml#L12-L22

This also means we could test for these vars over both dep and prod, since we'll just have values like `"fake"` in dep/dev.

beetmoverscript/src/beetmoverscript/data/artifactMap_beetmover_task_schema.json
```diff
@@ -57,12 +57,7 @@
                         }
                     },
                     "required": [
-                        "appName",
-                        "buildid",
-                        "appVersion",
-                        "hashType",
-                        "platform",
-                        "branch"
+                        "appName"
```
escapewindow, line 60: Am I right in thinking that these keys may be required in some types of behaviors but not others?
If so, we might want to validate that these keys exist for the behaviors that need them... we can do that with multiple jsonschema files (one per behavior?) or by just making sure the task is well-formed for each behavior in the python code.

beetmoverscript/src/beetmoverscript/task.py
```diff
@@ -34,9 +36,39 @@ def validate_task_schema(context):
     client.validate_task_schema(context, schema_key=schema_key)
 
 
+def _get_bucket_prefixes(script_config):
+    return _get_scope_prefixes(script_config, "bucket")
+
+
+def _get_action_prefixes(script_config):
+    return _get_scope_prefixes(script_config, "action")
```
escapewindow, line 44: We've moved away from action scopes elsewhere, to putting the behavior/action in the task payload, but that shouldn't block here.

escapewindow (commented):
Other than wondering if changing the required schema looks like will affect other actions, this looks good, thank you!

beetmoverscript/src/beetmoverscript/task.py
hneiva: is it an easy migration? Worth doing here? We've got some time still

escapewindow: The steps would probably be:

- support setting the action/behavior from a task payload key/value pair
  - add that to the schema, look there first, fall back to the scope
  - roll that out to production beetmover
- change the in-tree configs to use the task payload instead
- repeat for all branches and repos that use the scope to set the action/behavior
- once all uses of action scopes are in the task payload instead of scope, remove the scope and support for reading the action from the scope

If we introduced the action scope in this PR, that would be pretty easy. I suspect we're using beetmover action scopes all over gecko and other repos, so the in-tree portion might be lengthy.

beetmoverscript/docker.d/worker.yml
ahal: I think `focus` should be `focus-android` now?

beetmoverscript/docker.d/worker.yml
```diff
@@ -131,6 +145,14 @@ bucket_config:
           buckets:
             nightly_components: 'maven-nightly-stage-s3-upload-bucket'
         dummy: {}
+        dep:
+          credentials:
+            id: { "$eval": "DEP_ID" }
+            key: { "$eval": "DEP_KEY" }
+          buckets:
+            fenix: 'net-mozaws-stage-delivery-archive'
+            focus: 'net-mozaws-stage-delivery-archive'
```
ahal, line 154: This might need to be `focus-android` too? I really don't know though.

hneiva:
@escapewindow can you think of any reason not to merge this in? We've been using it for the past week without issues and would like to start working towards prod archive

beetmoverscript/docker.d/worker.yml
escapewindow: let's `s,focus,focus-android,` here before merging if possible, or quick followup?

beetmoverscript/docker.d/worker.yml
escapewindow: yeah, let's maybe leave this til we know and we can adjust if/when we beetmove focus and find bustage.

beetmoverscript/src/beetmoverscript/data/artifactMap_beetmover_task_schema.json
escapewindow: let's file a followup on this.

beetmoverscript/src/beetmoverscript/task.py
escapewindow: let's punt on moving away from action scopes. sometime in the unplanned future.

escapewindow (changes requested):
Looks like I just want the focus -> focus-android scope prefix change, then let's merge!

escapewindow (approved):

PR #439, opened 2021-11-17
adjust cert used for verification via -cafile when in production

jmaher:


signingscript/docker.d/init_worker.sh
```diff
@@ -38,7 +38,7 @@ if [ "$ENV" == "prod" ]; then
   export AUTHENTICODE_TIMESTAMP_STYLE=old
   export AUTHENTICODE_CERT_PATH=/app/signingscript/src/signingscript/data/authenticode_prod_202005.crt
   export AUTHENTICODE_CERT_PATH_202005=/app/signingscript/src/signingscript/data/authenticode_prod_202005.crt
-  export AUTHENTICODE_CA_PATH=/app/signingscript/src/signingscript/data/authenticode_prod_ca_202005.crt
+  export AUTHENTICODE_CA_PATH=/app/signingscript/src/signingscript/data/authenticode_prod_202005.crt
```
escapewindow, line 41: I think this may work, but do we sign anything with the non-202005 cert?

escapewindow (approved):

signingscript/docker.d/init_worker.sh
escapewindow: (If so, we may either want to concatenate the crt files into one, or point at the CERT_PATH that we're using. Also, I'm not sure if the current CA_PATH has additional info that may be needed in the .crt. But this is worth a shot.)

jmaher:
Thanks for the comments, to be honest I don't have any idea what certs we
really use to sign stuff with.  From reading the code, it seems like in
production we use 202005 version of the certs for everything, ideally we
should make that the default and remove the old cert.

I like the idea of cat authenticode_prod_*_202005.crt >
authenticode_prod_bundle.crt and using that.  If we are in a position where
we are backing out again, this would be the next iteration to try.

escapewindow:
I see a bunch of AUTHENTICODE prod password file entries that don't use the keyid 20205, e.g. https://github.com/mozilla-releng/scriptworker-scripts/blob/master/signingscript/docker.d/passwords.yml#L193 . We may want to audit what uses those so we're not blindsided when we roll out... this is across trust domains, e.g. vpn. But I'm not opposed to trying and seeing if it sticks.

I did search and it looks like the contents of `authenticode_prod_ca_202005.crt` is part of `authenticode_prod_202005.crt`, so this won't break anything additional. I am worried in that the dep CA crt isn't part of the dep signing crt, but we'll see if this sticks.

PR #440, opened 2021-11-19
Add focus-android to githubscript docker file

sarah-clements:


jmaher (approved):
I am a bit rusty here, but some grepping in code and double checking a few things gives me confidence that this looks safe.  There might be needs for ci-admin.  and I am not sure if staging will be able to give you the full test you need- key words being not sure.

githubscript/docker.d/worker.yml
```diff
@@ -63,6 +69,6 @@ github_projects:
 taskcluster_scope_prefixes:
   $switch:
     'COT_PRODUCT == "mobile"':
-       ["project:mobile:fenix:releng:github:", "project:mobile:android-components:releng:github:"]
+       ["project:mobile:fenix:releng:github:", "project:mobile:android-components:releng:github:", "project:mobile:focus-android:releng:github:"]
```
MihaiTabara, line 72: Nit: I may be wrong but we may need to whitelist this scope in https://github.com/mozilla-releng/scriptworker/blob/master/src/scriptworker/constants.py#L356 CC @escapewindow

sarah-clements: Interesting, I didn't add the "project:mobile:focus-android:releng:googleplay:product:" to that module when modifying the pushapk scriptworker. Is this specific to the githubscript?

sarah-clements:
Do I need to change anything else for version-bump? It looks like thats done via treescript, right?

githubscript/docker.d/worker.yml
escapewindow: @sarah-clements ideally we make all production-only scopes restricted there, so if we missed pushapk, let's add it.

e.g., we have

```
# scope                                                 repo/branch alias
"project:mobile:fenix:releng:googleplay:product:fenix": "fenix-repo",
```

[here](https://github.com/mozilla-releng/scriptworker/blob/master/src/scriptworker/constants.py#L360) to limit pushing the fenix apk from the fenix repo itself, and https://github.com/mozilla-releng/scriptworker/blob/master/src/scriptworker/constants.py#L356 for limiting the a-c github scope to the a-c repo as @MihaiTabara mentioned.

escapewindow:
> Do I need to change anything else for version-bump? It looks like thats done via treescript, right?

It looks like we have mobile pools for treescript already https://firefox-ci-tc.services.mozilla.com/provisioners/scriptworker-k8s/ . If we're not able to land changes, we may have to adjust the github bots' permissions in the focus-android repo, but I think that's doable without a patch. It does look like we restrict the `push` scope for treescript for desktop [here](https://github.com/mozilla-releng/scriptworker/blob/master/src/scriptworker/constants.py#L339), but we seem to be missing it for mobile. If we can identify and add a mobile scope, that would be good.

githubscript/docker.d/worker.yml
sarah-clements: Made changes in scriptworker for this: https://github.com/mozilla-releng/scriptworker/pull/522

githubscript/docker.d/worker.yml
```diff
@@ -15,6 +15,12 @@ github_projects:
         github_owner: mozilla-mobile
         github_repo_name: android-components
         contact_github: true
+      focus-android:
```
escapewindow, line 18: If we want to add a staging-focus-android below (in fake-prod and dev) that may help in some testing scenarios; otherwise we can use the `mock` repo.

escapewindow (approved):
This looks good, thank you!
scriptworker#522 has some changes requested. Once that gets merged+released, we'll want a `maintenance/pin.sh` run to update the pinned python deps. That could belong in this PR, or it could potentially go in another PR depending on our timing. Rolling out the restricted scopes along with this change is best practice, but the latter may allow us to unblock ourselves if there are any urgent deadlines or external requirements here.

sarah-clements:
Scriptworker bump has been merged first (#447)

PR #446, opened 2021-12-02
remove remaining scriptworker imports from pushmsixscript

### LINKED ISSUE #443: [pushmsixscript] cleanup remaining scriptworker imports
pushmsixscript mostly relies on scriptworker-client rather than scriptworker directly, but there are some remaining uses of scriptworker. I should have another try at resolving those.

gbrownmozilla:
Adds scriptworker_client.artifacts (a copy of part of scriptworker.artifact) and other changes to eliminate direct use of scriptworker by pushmsixscript.

Closes #443


escapewindow (approved):
Awesome, thank you for following up!

PR #451, opened 2021-12-10
tidy init_worker.sh and worker.yml

### LINKED ISSUE #445: [pushmsixscript] update init_worker.sh and worker.yml
From https://github.com/mozilla-releng/scriptworker-scripts/pull/442, follow-up on aki's comments about init_worker.sh:

```
In this file we'd test_var_set the various env vars that Cloudops will populate the k8s pods with.
We create those variable names in cloudops-infra, and we populate the secret values in our relengworker sops.
I imagine we'll need to at least define whatever creds we need to authenticate with the microsoft store.
```

and for worker.yml

```
All of these $eval variables would be good to either explicitly set in init_worker.sh (for non-secrets) or to pull from the environment in init_worker.sh (for secrets).

(For user/pass combinations, one could argue that the username isn't a secret, but having to update the user in one repo and the password in another, with different release flows, is a pain. Rotating an entire sops file full of dozens or hundreds of secrets, and not knowing which username they belong to, is also a massive pain. I've settled on pairing the user/pass combinations together in sops to avoid both of those pain points.)
```

gbrownmozilla:
Treat TENANT_ID, CLIENT_ID, and CLIENT_SECRET as secrets; remaining parameters are not sensitive and expected to be fairly constant: set them in init_worker.sh.

Closes #445


gbrownmozilla:
See also https://github.com/mozilla-services/cloudops-infra/pull/3635

pushmsixscript/docker.d/init_worker.sh
```diff
@@ -10,10 +10,24 @@ test_var_set() {
   fi
 }
 
+export REQUEST_TIMEOUT_SECONDS=30
+export LOGIN_URL=https://login.microsoftonline.com
+export TOKEN_RESOURCE=https://manage.devcenter.microsoft.com
+export STORE_URL=https://manage.devcenter.microsoft.com/v1.0/my/applications/
+export MOCK_APPLICATION_ID=MOCK-APP-ID
+export BETA_APPLICATION_ID=9NZW26FRNDLN
+export RELEASE_APPLICATION_ID=9NZVDKPMR9RD
```
escapewindow, line 19: This placement makes sense. (We may end up moving some of these into the `if` blocks below if we ever ship more than 1 product, but that may never happen.)

pushmsixscript/docker.d/worker.yml
```diff
@@ -6,12 +6,12 @@ request_timeout_seconds: { "$eval": "REQUEST_TIMEOUT_SECONDS" }
 login_url: { "$eval": "LOGIN_URL" }
 token_resource: { "$eval": "TOKEN_RESOURCE" }
 store_url: { "$eval": "STORE_URL" }
-tenant_id: { "$eval": "TENANT_ID" }
-client_id: { "$eval": "CLIENT_ID" }
-client_secret: { "$eval": "CLIENT_SECRET" }
 application_ids:
```
escapewindow, line 9: do we want to add the mock id, or is that covered elsewhere?

escapewindow (approved):
Thank you!

pushmsixscript/docker.d/worker.yml
gbrownmozilla: Other than the prod environment, push_to_store = False (afaik, there is only The One Store, nothing for dev/staging purposes), so credentials won't be used.

PR #453, opened 2021-12-16
treescript: make do_merge idempotent

jcristau:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1744721


treescript/tests/test_merges.py
```diff
@@ -243,36 +269,54 @@ async def test_do_merge(mocker, config, task, repo_context, merge_info, add_merg
     if l10n_bump:
         task["payload"]["l10n_bump_info"] = {"foo": "bar"}
 
+    orig_run_hg_command = merges.run_hg_command
```
jcristau: flake8 gets unhappy:
```
check run-test: commands[3] | flake8 /app
./tests/test_merges.py:264:1: C901 'test_do_merge' is too complex (11)
async def test_do_merge(mocker, config, task, repo_context, merge_info, add_merge_info, raises, expected_calls, l10n_bump, expected_return):
^
ERROR: InvocationError for command /app/.tox/check/bin/flake8 . (exited with code 1)
```

escapewindow: We've previously gotten around this by moving blocks of complex logic to helper functions. Also, I'm not too worried about the complexity of a *test* function as long as it's not over the top, so we could also skip this flake8 test on this particular function.

jcristau: Thanks, I moved some of the setup for this test to a helper function.

escapewindow (approved):
Thank you! Let's add this to the [changelog](https://github.com/mozilla/build-relengdocs/blob/main/releng_changelog.md) for the next mergeduty folk.

PR #454, opened 2021-12-17
Update pushmsixscript based on real-world testing against the Store

gbrownmozilla:
- s/upload_url/fileUploadUrl/
 - add some useful defaults to the submission request
 - force lower-case true/false in the submission request


jmaher (approved):
a lot of working around things, might need to revisit over time as the store changes, etc.

PR #457, opened 2022-01-17
Allow multiple msix uploads in one Store submission

gbrownmozilla:
It turns out we need to upload both 32-bit and 64-bit msix files in one submission to the Microsoft Store.

### DIFF
--- a/pushmsixscript/src/pushmsixscript/artifacts.py
+++ b/pushmsixscript/src/pushmsixscript/artifacts.py
@@ -1,17 +1,13 @@
 from scriptworker_client import artifacts
 from scriptworker_client.exceptions import TaskVerificationError
-from scriptworker_client.utils import get_single_item_from_sequence
 
 
-def get_msix_file_path(config, task):
+def get_msix_file_paths(config, task):
     artifacts_per_task_id, _ = artifacts.get_upstream_artifacts_full_paths_per_task_id(config, task)
 
     all_artifacts = [artifact for artifacts in artifacts_per_task_id.values() for artifact in artifacts]
 
-    return get_single_item_from_sequence(
-        all_artifacts,
-        condition=lambda artifact: artifact.endswith(".store.msix"),
-        ErrorClass=TaskVerificationError,
-        no_item_error_message="No upstream artifact is a store msix",
-        too_many_item_error_message="Too many msix artifacts found",
-    )
+    filtered_artifacts = [item for item in all_artifacts if item.endswith(".store.msix")]
+    if len(filtered_artifacts) == 0:
+        raise TaskVerificationError("No upstream artifact is a store msix")
+    return filtered_artifacts
--- a/pushmsixscript/src/pushmsixscript/microsoft_store.py
+++ b/pushmsixscript/src/pushmsixscript/microsoft_store.py
@@ -35,16 +35,16 @@ DEFAULT_LISTINGS = {
 DEFAULT_ALLOW_TARGET = {"Desktop": True, "Mobile": False, "Holographic": False, "Xbox": False}
 
 
-def push(config, msix_file_path, channel):
-    """Publishes a msix onto a given channel.
+def push(config, msix_file_paths, channel):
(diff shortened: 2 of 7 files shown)


pushmsixscript/src/pushmsixscript/artifacts.py
```diff
@@ -1,17 +1,13 @@
...
     all_artifacts = [artifact for artifacts in artifacts_per_task_id.values() for artifact in artifacts]
 
-    return get_single_item_from_sequence(
-        all_artifacts,
-        condition=lambda artifact: artifact.endswith(".store.msix"),
-        ErrorClass=TaskVerificationError,
-        no_item_error_message="No upstream artifact is a store msix",
-        too_many_item_error_message="Too many msix artifacts found",
-    )
+    filtered_artifacts = [item for item in all_artifacts if item.endswith(".store.msix")]
```
escapewindow, line 10: Do we have any upstreamArtifacts that don't end in `.store.msix` ?
Just curious.

pushmsixscript/src/pushmsixscript/microsoft_store.py
```diff
@@ -148,12 +148,13 @@ def _update_submission(config, channel, session, submission_request, headers, fi
     response = session.put(url, submission_request, headers=headers)
     _log_response(response)
     response.raise_for_status()
-    with open(file_path, "rb") as f:
-        upload_headers = {"x-ms-blob-type": "BlockBlob"}
-        response = requests.put(upload_url, f, headers=upload_headers, timeout=config["request_timeout_seconds"])
-        response.raise_for_status()
-        _log_response(response)
-        # no response body expected
+    for file_path in file_paths:
```
escapewindow, line 151: This is going to be synchronous. We may want to at least test to see if we can submit files concurrently to the store. Are you familiar with async python? The switch to coroutines should be fairly simple; we'd go from

```
async def async_main():
    foo()


def foo():
    bar()


def bar():
    # do stuff
```

to

```
async def async_main():
    await foo()


async def foo():
    await bar()


async def bar():
    # do stuff
```

Mainly, it's easiest if the calling function of your async coroutine is also an async coroutine; it's ok to call synchronous functions if we intend for those steps to be synchronous. This also means we want to go from requests to aiohttp, which is close to a drop-in, with the exception of the extra `await`s. We don't have to make this change in this PR, but this might be a good followup.

pushmsixscript/src/pushmsixscript/script.py
```diff
@@ -12,14 +12,15 @@
 
 
 async def async_main(config, task_dict):
-    msix_file_path = artifacts.get_msix_file_path(config, task_dict)
-    log.info(f"found msix at {msix_file_path}")
-    manifest.verify_msix(msix_file_path)
+    msix_file_paths = artifacts.get_msix_file_paths(config, task_dict)
+    for msix_file_path in msix_file_paths:
+        log.info(f"found msix at {msix_file_path}")
+        manifest.verify_msix(msix_file_path)
```
escapewindow, line 18: If this takes a while to run, we could look at running this async as well as the push.

escapewindow (approved):

pushmsixscript/src/pushmsixscript/artifacts.py
gbrownmozilla: I don't think so, but there are other artifacts produced alongside .store.msix (including other .msix artifacts), and other scripts do similar filtering: https://github.com/mozilla-releng/scriptworker-scripts/blob/master/pushapkscript/src/pushapkscript/script.py#L30. Maybe we are all just being extra careful.

pushmsixscript/src/pushmsixscript/script.py
gbrownmozilla: This part is quite fast.

pushmsixscript/src/pushmsixscript/microsoft_store.py
gbrownmozilla: Thanks. I'll look into async as a followup.

PR #458, opened 2022-01-18
adds xpi config to balrog & beetmover workers

gabrielbusta:
Adds xpi config to balrog & beetmover workers.

### DIFF
--- a/beetmoverscript/docker.d/init_worker.sh
+++ b/beetmoverscript/docker.d/init_worker.sh
@@ -40,6 +40,21 @@ case $COT_PRODUCT in
         ;;
     esac
     ;;
+  xpi)
+    case $ENV in
+      dev|fake-prod)
+        test_var_set 'DEP_ID'
+        test_var_set 'DEP_KEY'
+        ;;
+      prod)
+        test_var_set 'RELEASE_ID'
+        test_var_set 'RELEASE_KEY'
+        ;;
+      *)
+        exit 1
+        ;;
+    esac
+    ;;
   thunderbird)
     case $ENV in
       dev|fake-prod)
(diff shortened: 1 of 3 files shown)


beetmoverscript/docker.d/init_worker.sh
```diff
@@ -40,6 +40,21 @@ case $COT_PRODUCT in
...
     ;;
+  xpi)
+    case $ENV in
+      dev|fake-prod)
+        test_var_set 'DEP_ID'
+        test_var_set 'DEP_KEY'
+        ;;
+      prod)
+        test_var_set 'RELEASE_ID'
+        test_var_set 'RELEASE_KEY'
```
escapewindow, line 51: These match up with sops 👍

beetmoverscript/docker.d/worker.yml
```diff
@@ -99,6 +99,22 @@ bucket_config:
...
             geckoview: 'maven-default-s3-upload-bucket-1705kkmo2aj3c'
         dummy: {}
+      'COT_PRODUCT == "xpi" && (ENV == "dev" || ENV == "fake-prod")':
+        dep:
+          credentials:
+            id: { "$eval": "DEP_ID" }
+            key: { "$eval": "DEP_KEY" }
+          buckets:
+            xpi: 'net-mozaws-stage-delivery-archive'
+          url_prefix: 'https://ftp.stage.mozaws.net'
```
escapewindow, line 109: We may want to add a `dummy: {}` line -- I'm not 100% sure but I think this lets us test basic sanity checks without actually pushing to a bucket. I *think* we have that for every pool, but we'd at least want the dev/fake-prod pools.

escapewindow (approved):
r+ with changes, thank you!

beetmoverscript/docker.d/worker.yml
gabrielbusta: OK. I added it.

PR #459, opened 2022-01-20
[signingscript] Add 'autograph_apk' format to mozillavpn

ahal:
The VPN team is standing up a signing task for their Android app and need autograph_apk support.

### DIFF
--- a/signingscript/docker.d/init_worker.sh
+++ b/signingscript/docker.d/init_worker.sh
@@ -205,6 +205,8 @@ case $ENV in
       mozillavpn)
         test_var_set 'AUTOGRAPH_AUTHENTICODE_PASSWORD'
         test_var_set 'AUTOGRAPH_AUTHENTICODE_USERNAME'
+        test_var_set 'AUTOGRAPH_MOZILLAVPN_PASSWORD'
+        test_var_set 'AUTOGRAPH_MOZILLAVPN_USERNAME'
         test_var_set 'AUTHENTICODE_CERT_PATH'
         test_var_set 'AUTHENTICODE_CA_PATH'
         test_var_set 'AUTHENTICODE_CA_TIMESTAMP_PATH'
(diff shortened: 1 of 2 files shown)


signingscript/docker.d/init_worker.sh
```diff
@@ -205,6 +205,8 @@ case $ENV in
       mozillavpn)
         test_var_set 'AUTOGRAPH_AUTHENTICODE_PASSWORD'
         test_var_set 'AUTOGRAPH_AUTHENTICODE_USERNAME'
+        test_var_set 'AUTOGRAPH_MOZILLAVPN_PASSWORD'
+        test_var_set 'AUTOGRAPH_MOZILLAVPN_USERNAME'
```
ahal, line 209: @escapewindow I can't remember where these variables are supposed to come from or how to generate the passwords (that's why this is still a draft).

ahal: I found where this is set, trying to figure out how to request the creds.

escapewindow: Anything public would go [here](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/relengworker/k8s/values/signing.yaml), but nothing here is public, since we want to keep the user+pass together in sops.

We declare secrets [here](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/relengworker/k8s/charts/signing/values.yaml#L35) but define them in k8s-sops. We then translate the secret from a yml entry into an env var [here](https://github.com/mozilla-services/cloudops-infra/blob/master/projects/relengworker/k8s/charts/signing/templates/secret.yaml#L12).

escapewindow: If there are multiple pools that may reference some of these values, we tend to specify fake values in k8s-sops to avoid bustage related to missing secrets.

So it's possible that gecko signingscript pools may not refer to the mozillavpn values, but if adhoc-signing dep and dev refer to them, it may be best to specify the key/value pairs with a value of `"fake"`.

ahal: Here's the cloudops-infra PR: https://github.com/mozilla-services/cloudops-infra/pull/3747

ahal:
cloudops-infra PR has merged, this should be ready to land.

escapewindow (approved):
It will be good if we add a dep signing something so we can test e.g. staging releases, but this should unblock for now.

Let's have them sign something and verify that this is the same key, otherwise we need an Autograph-side config change.

Issue #463, opened 2022-01-27
pushmsixscript fails: TypeError: '<=' not supported between instances of 'str' and 'int'


gbrownmozilla:
https://firefoxci.taskcluster-artifacts.net/FnIDi5a1T0SJk8UULyWcEw/0/public/logs/live_backing.log failed with:

```
2022-01-27 22:24:43,483 - urllib3.util.retry - DEBUG - Converted retries value: 5 -> Retry(total=5, connect=None, read=None, redirect=None, status=None)
Traceback (most recent call last):
  File "/app/lib/python3.9/site-packages/urllib3/util/timeout.py", line 150, in _validate_timeout
    if value <= 0:
TypeError: '<=' not supported between instances of 'str' and 'int'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/app/bin/pushmsixscript", line 8, in <module>
    sys.exit(main())
  File "/app/lib/python3.9/site-packages/pushmsixscript/script.py", line 42, in main
    return client.sync_main(async_main, config_path=config_path, default_config=get_default_config())
  File "/app/lib/python3.9/site-packages/scriptworker_client/client.py", line 127, in sync_main
    loop.run_until_complete(_handle_asyncio_loop(async_main, config, task))
  File "/usr/local/lib/python3.9/asyncio/base_events.py", line 642, in run_until_complete
    return future.result()
  File "/app/lib/python3.9/site-packages/scriptworker_client/client.py", line 182, in _handle_asyncio_loop
    await async_main(config, task)
  File "/app/lib/python3.9/site-packages/pushmsixscript/script.py", line 23, in async_main
    microsoft_store.push(config, msix_file_paths, channel)
  File "/app/lib/python3.9/site-packages/pushmsixscript/microsoft_store.py", line 55, in push
    access_token = _store_session(config)
  File "/app/lib/python3.9/site-packages/pushmsixscript/microsoft_store.py", line 83, in _store_session
    response = session.post(url, body, headers=headers, timeout=config["request_timeout_seconds"])
  File "/app/lib/python3.9/site-packages/requests/sessions.py", line 590, in post
    return self.request('POST', url, data=data, json=json, **kwargs)
  File "/app/lib/python3.9/site-packages/requests/sessions.py", line 542, in request
    resp = self.send(prep, **send_kwargs)
  File "/app/lib/python3.9/site-packages/requests/sessions.py", line 655, in send
    r = adapter.send(request, **kwargs)
[...]

PR #465, opened 2022-01-28
[pushmsixscript] Allow for type(config["request_timeout_seconds"]) == str

gbrownmozilla:
request_timeout_seconds is always accessed as an int, but the type is dependent on whether or not quotes are used in the config file: guard against accidental use of quotes for this one and only int in the config.


gbrownmozilla:
Closes #463

escapewindow (approved):

PR #468, opened 2022-01-28
[pushmsixscript] Fail if a pending submission is found on the Microsoft Store

### LINKED ISSUE #467: [pushmsixscript] fail if another submission is pending
On the Microsoft Store, there can only be one pending submission at a time. If there is a pending submission and a new submission is made through the API, the new submission will fail. pushmsix currently avoids this (as recommended by the Store docs) by deleting the pending submission. An advantage of this is that if pushmsix is interrupted (network outage during upload), it will likely leave behind a pending submission; on re-run, pushmsix will delete the partial submission and likely succeed on the new submission.

ryanvm points out that, in addition to submissions made for new release builds, Product people make submissions to update content, and those submissions go through the same time-consuming certification process: they are typically pending for hours. To avoid possibly over-writing a Product submission with a CI-initiated submission, it's probably better for pushmsix to fail in this case, with a clear error message.

gbrownmozilla:
This is a requested change from release management (ryanvm): We want to be careful not to accidentally discard human-crafted submissions, even if that means the release task sometimes fails.

Closes #467

### DIFF
--- a/pushmsixscript/src/pushmsixscript/microsoft_store.py
+++ b/pushmsixscript/src/pushmsixscript/microsoft_store.py
@@ -89,7 +89,7 @@ def _push_to_store(config, channel, msix_file_paths, access_token):
     headers = {"Authorization": f"Bearer {access_token}", "Content-type": "application/json", "User-Agent": "Python"}
     with requests.Session() as session:
         session.mount("https://", requests.adapters.HTTPAdapter(max_retries=MAX_RETRIES))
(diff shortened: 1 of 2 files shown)


escapewindow (approved):
Looks like a good change.
This is the standard post-initial-rollout set of edge cases, robustness fixes, and user requests :) These will slow down after a while, and you'll know what kinds of things to expect if/when you roll out another scriptworker type.

PR #469, opened 2022-01-31
[pushmsixscript] Add publishMode to task payload, to specify Store's targetPublishMode

### LINKED ISSUE #466: [pushmsixscript] override targetPublishMode
Microsoft Store submissions have various settings; pushmsix currently uses the settings from the previous submission for the new submission. For instance, the publishing mode, targetPublishMode, for our Beta application is currently set to "immediate", so pushmsix Beta submissions will be immediate; the publishing mode for our Firefox (release) application is currently set to "manual", so pushmsix Release submissions will be manual. Any changes made through the Partner Center to an existing submission will be re-used for the next pushmsix submission.

However, ryanvm and jcristau prefer that the publishing mode specifically is made part of the task payload, always set to "immediate" for Beta and "manual" for Release.

I still think the existing re-use-the-previous-setting is a good thing (it's no accident it produces exactly the desired behavior!) so I'd like to also preserve that as an option.

https://docs.microsoft.com/en-us/windows/uwp/monetize/update-an-app-submission#request-body

gbrownmozilla:
Closes #466

### DIFF
--- a/pushmsixscript/src/pushmsixscript/data/push_msix_task_schema.json
+++ b/pushmsixscript/src/pushmsixscript/data/push_msix_task_schema.json
@@ -55,7 +55,7 @@
                 },
                 "publishMode": {
                   "type": "string"
-                },
+                }
             },
             "required": ["upstreamArtifacts", "channel"],
             "additionalProperties": false
--- a/pushmsixscript/src/pushmsixscript/data/push_msix_task_schema.json
+++ b/pushmsixscript/src/pushmsixscript/data/push_msix_task_schema.json
@@ -50,28 +50,14 @@
                     "minItems": 1,
                     "uniqueItems": true
                 },
-                "target_store": {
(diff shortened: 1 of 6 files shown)


escapewindow (approved):
These look like good changes, thank you!
My only question: should we limit the publish mode to a set of known/supported publish modes, or is it preferable to error out when we hit the store?

gbrownmozilla:
Good question, thanks. This "level" of the Store api - the data format of the submission api, and especially the options and choices - seems to be the least-defined and most ephemeral. Rather than revisiting validation code each time a new option becomes available, I'd like to pass through the configured value. In my experience, the Store rejects invalid values with a good error message, which should show up in the task log.

escapewindow:
Sounds good. It may be worth having some documentation somewhere about the standard operating procedure, once pushmsix is mature enough to have an SOP :)

PR #471, opened 2022-02-02
Add vpn dep signing

hneiva:


iscript/src/iscript/macvpn.py
```diff
@@ -168,17 +172,19 @@ async def notarize_vpn_behavior(config, task):
     zip_path = os.path.join(config["work_dir"], "notarization.zip")
     await _create_notarization_zipfile(config["work_dir"], top_app.pkg_path, zip_path)
```
escapewindow: You can probably put these two lines inside the notarize block?

iscript/src/iscript/script.py
```diff
@@ -8,52 +8,79 @@
...
-    elif behavior == "mac_sign_and_pkg":
-        # For staging releases; or should we mac_notarize but skip notarization
-        # for dep?
-        await sign_and_pkg_behavior(config, task)
-        return
-    raise IScriptError("Unknown iscript behavior {}!".format(behavior))
+
+    # Check for dep behaviors (skips notarization, only signs with dep keys)
+    # Raises if behavior not supported
+    behavior = check_dep_behavior(task, behavior, sign_config["supported_behaviors"])
```
escapewindow, line 90: It looks like we call this and fall back to the dep behavior whether we're dep signing or not. Was that intentional? We should only fall back if the original behavior isn't in the supported behaviors, or if we're dep.

Per [this block](https://github.com/mozilla-platform-ops/ronin_puppet/blob/master/data/common.yaml#L89-L99), this would have turned off notarization for all of our prod pools.

iscript/src/iscript/script.py
```diff
@@ -8,52 +8,79 @@
...
+    return functions[behavior]
+
+
 async def async_main(config, task):
     """Sign all the things.
 
     Args:
         config (dict): the running config.
         task (dict): the running task.
-
```
escapewindow: Historically, docstring linting has said we have to end with an empty line, but this is ok if that rule has changed.

escapewindow (changes requested):

iscript/src/iscript/script.py
hneiva: You'd think our tests would catch that. Nice find :)

hneiva: (I probably broke the tests btw, so still on me hahahah)

hneiva:
Tested here: https://github.com/mozilla-releng/adhoc-signing/pull/76

escapewindow (approved):
Thank you!

PR #475, opened 2022-02-11
Bug 1613283 Poller log rotation

hneiva:


escapewindow (approved):
Thank you!
We'll roll this out via updating the [scriptworker_scripts_revision](https://github.com/mozilla-platform-ops/ronin_puppet/blob/master/data/common.yaml#L125-L142) for at least both prod pools.

PR #476, opened 2022-02-11
Bug 1751808 Intermittent iscript failure

hneiva:
Added retry_async to run_command in create_pkg_files

### DIFF
--- a/iscript/src/iscript/mac.py
+++ b/iscript/src/iscript/mac.py
@@ -992,20 +992,22 @@ async def create_pkg_files(config, sign_config, all_paths, requirements_plist_pa
             asyncio.ensure_future(
                 semaphore_wrapper(
                     semaphore,
-                    run_command(
-                        [
-                            "pkgbuild",
-                            "--install-location",
-                            "/Applications",
-                        ]
-                        + cmd_opts
-                        + [
-                            "--component",
-                            app.app_path,
-                            app.tmp_pkg_path1,
-                        ],
-                        cwd=app.parent_dir,
-                        exception=IScriptError,
+                    retry_async(
+                        func=run_command,
+                        kwargs={
+                            "cmd": [
+                                "pkgbuild",
+                                "--install-location",
+                                "/Applications",
+                                *cmd_opts,
+                                "--component",
+                                app.app_path,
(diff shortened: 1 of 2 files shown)


hneiva:
🤔 
`tox -e py38` works locally. I'll try re-triggering, looks like TC had an exception or something

iscript/src/iscript/mac.py
```diff
@@ -992,20 +992,22 @@ async def create_pkg_files(config, sign_config, all_paths, requirements_plist_pa
...
-                            "--component",
-                            app.app_path,
-                            app.tmp_pkg_path1,
-                        ],
-                        cwd=app.parent_dir,
-                        exception=IScriptError,
+                    retry_async(
+                        func=run_command,
+                        kwargs={
+                            "cmd": [
```
escapewindow, line 998: We might want `cmd` to be in `"args": ("pkgbuild", ...),`

hneiva: Same difference, no?

hneiva: Now that I think about it, having it in `args`, makes it more confusing.
We want it to be:
```
run_command(["pkgbuild", "--install-location"....], cwd="foo", exception=Bar)
```
But having `"args": ("pkgbuild", "--install-location"....),` means we'd be calling it as:
```
run_command("pkgbuild", "--install-location"...., cwd="foo", exception=Bar)
```
Which isn't the same, (but could be solved by double wrapping the args) so in order to keep the code more readable, I'd vote for using just kwargs.

escapewindow (approved):
Thanks!

For extra credit, we can add a retry_async test:
1. add a counter
2. fail on first attempt
3. succeed on second attempt
4. verify that we've succeeded, and the counter is 2

iscript/src/iscript/mac.py
escapewindow: These are supposed to essentially mirror `*args` and `**kwargs` https://www.geeksforgeeks.org/args-kwargs-python/

escapewindow: and yeah, I suppose it would be `"args": (["pkgbuild", ...])`

PR #477, opened 2022-02-14
balrog & beetmover scriptworkers system addons support

gabrielbusta:
These changes modify the balrog & beetmover scriptworkers to support system add-on releases.

I tested these changes in staging by shipping the balrog-dryrun add-on.

**Links**
* [beetmover task ✅](https://firefox-ci-tc.services.mozilla.com/tasks/NegBVXwVR4anezRhr4PXKQ)
  * [balrog manifest](https://firefoxci.taskcluster-artifacts.net/NegBVXwVR4anezRhr4PXKQ/0/public/manifest.json)
  * [artifact archive](https://ftp.stage.mozaws.net/pub/system-addons/balrog-dryrun/balrog-dryrun-3.0.0-build7/)
* [balrog task ✅](https://firefox-ci-tc.services.mozilla.com/tasks/LxmBLOfNT5WW0LGbYEDVSg)
  * [staging release](https://balrog-admin-static-stage.stage.mozaws.net/releases#balrog-dryrun-3.0.0-build7) 
  * [balrog payload](https://firefox-ci-tc.services.mozilla.com/tasks/LxmBLOfNT5WW0LGbYEDVSg/runs/1/logs/public/logs/live_backing.log#L9)

### DIFF
--- a/balrogscript/docker.d/worker.yml
+++ b/balrogscript/docker.d/worker.yml
@@ -69,3 +69,21 @@ server_config:
       - 'esr'
       - 'esr-localtest'
       - 'esr-cdntest'
+  staging:
+    api_root: { "$eval": "STAGE_API_ROOT" }
+    auth0_domain: { "$eval": "AUTH0_DOMAIN"}
+    auth0_client_id: { "$eval": "AUTH0_CLIENT_ID" }
+    auth0_client_secret: { "$eval": "AUTH0_CLIENT_SECRET" }
+    auth0_audience: { "$eval": "AUTH0_AUDIENCE" }
+    allowed_channels:
+      - 'nightly'
+      - 'aurora'
+      - 'beta'
+      - 'beta-localtest'
+      - 'beta-cdntest'
+      - 'release'
+      - 'release-localtest'
+      - 'release-cdntest'
+      - 'esr'
+      - 'esr-localtest'
+      - 'esr-cdntest'
--- a/balrogscript/pyproject.toml
+++ b/balrogscript/pyproject.toml
@@ -23,6 +23,7 @@ exclude = '''
 [tool.isort]
 line_length = 160
 known_first_party = 'balrogscript'
+profile = "black"
 
 [tool.towncrier]
 package = "balrogscript"
--- a/balrogscript/src/balrogscript/submitter/cli.py
+++ b/balrogscript/src/balrogscript/submitter/cli.py
@@ -6,6 +6,7 @@ from balrogclient import Release, ReleaseState, Rule, ScheduledRuleChange, Singl
 from deepmerge import always_merger
 from redo import retry
 from requests.exceptions import HTTPError
+from balrogscript.constants import SYSTEM_ADDONS_PLATFORMS
 
 from .release import buildbot2bouncer, buildbot2ftp, buildbot2updatePlatforms, getPrettyVersion, getProductDetails, makeCandidatesDir
 from .util import recursive_update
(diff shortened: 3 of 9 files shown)


balrogscript/docker.d/worker.yml
```diff
@@ -69,3 +69,21 @@ server_config:
       - 'esr'
       - 'esr-localtest'
       - 'esr-cdntest'
+  staging:
```
escapewindow, line 72: Hm, how is this different from `dep` ? Other than a more intuitive name? :)

balrogscript/pyproject.toml
```diff
@@ -23,6 +23,7 @@ exclude = '''
 [tool.isort]
 line_length = 160
 known_first_party = 'balrogscript'
+profile = "black"
```
escapewindow, line 26: 👍

balrogscript/docker.d/worker.yml
```diff
@@ -69,3 +69,21 @@ server_config:
...
+    api_root: { "$eval": "STAGE_API_ROOT" }
+    auth0_domain: { "$eval": "AUTH0_DOMAIN"}
+    auth0_client_id: { "$eval": "AUTH0_CLIENT_ID" }
+    auth0_client_secret: { "$eval": "AUTH0_CLIENT_SECRET" }
+    auth0_audience: { "$eval": "AUTH0_AUDIENCE" }
+    allowed_channels:
```
escapewindow, line 78: For staging, I wonder if there's a way to say "all channels are allowed".

Not necessary. But for system addons we may at least want to add `*-sysaddon` since that's how we named our test channels there https://github.com/mozilla-releng/systemaddons-wip/tree/main/existing/SystemAddons/rules .

balrogscript/src/balrogscript/submitter/cli.py
```diff
@@ -538,3 +538,49 @@ def __init__(self, api_root, auth0_secrets=None):
...
+    def run(self, manifest):
+        release_blob = {
+            "addons": {},
+            "hashFunction": manifest["hashType"],
+            "name": manifest["releaseName"],
+            "product": "SystemAddons",
+            "schema_version": 5000,
+        }
+        for addon in manifest["addons"]:
+            platforms = {
```
escapewindow, line 560: We may want to move the platform aliases out into a module constant or config var, since we update them every time we add a new arch, compiler type, or platform. Similar to [`PLATFORMS`](https://github.com/mozilla/morgoth-cli/blob/master/morgoth/xpi.py#L12-L46). We could then deepcopy that into `platforms` before updating it with the `default` dict.

balrogscript/src/balrogscript/submitter/cli.py
```diff
@@ -538,3 +538,49 @@ def __init__(self, api_root, auth0_secrets=None):
     def run(self, productName, version, build_number):
         name = get_release_blob_name(productName, version, build_number)
         ReleaseState(name, api_root=self.api_root, auth0_secrets=self.auth0_secrets).set_readonly()
+
+
+class SystemAddonsReleaseCreator(object):
+    def __init__(self, api_root, auth0_secrets):
+        self.auth0_secrets = auth0_secrets
+        self.api_root = api_root
+        self.url = "{}/{}".format(self.api_root, "releases")
```
escapewindow, line 549: I think this means we'll be using v1 releases for now, which is the right call, since we want to be compatible with morgoth-cli as long as we're still using it.

beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -69,6 +70,25 @@
 log = logging.getLogger(__name__)
 
 
+# push_to_system_addons {{{1
+async def push_to_system_addons(context):
```
escapewindow, line 74: Side note: we've been moving away from using the god object `context` in other scriptworker-scripts, in favor of passing in more specific arguments to functions. Refactoring beetmoverscript to stop using `context` is definitely massive scope creep, though, so no need to here.

Examples:
- https://github.com/mozilla-releng/scriptworker-scripts/commit/402e25db439a0e26cc047f3711d73d91810b80ca
- https://github.com/mozilla-releng/scriptworker-scripts/commit/3ef1c5c27a6f83eb9d7a7fdf57b9a2fcd0a4a54b

probably others where the commit log doesn't mention `context` :)

escapewindow (approved):
The `PLATFORMS` and `staging` comments might be worth addressing sooner, though they're non-blocking. The `context` comment is totally future-able. I think every other comment is just informative.

Good job! Thank you!

balrogscript/docker.d/worker.yml
gabrielbusta: It matches the xpi scope in [grants.yml#l1813](https://hg.mozilla.org/ci/ci-configuration/file/2458e282d4e38fe75c237d47d2ab1332b3a0676d/grants.yml#l1813)

escapewindow: aha :)
In the future we probably want to standardize our naming, but this works, ty.

balrogscript/src/balrogscript/submitter/cli.py
gabrielbusta: It does. I kept the behavior the same as morgoth-cli

balrogscript/docker.d/worker.yml
gabrielbusta: Should we add `*-sysaddon` to the other servers? This might be relevant to the balrog rules

escapewindow: I think so.

`*-sysaddon` only really applies to the xpi balrog scriptworkers, so this is also a place where we could have a different set of allowed channels depending on the trust domain. If that's easy to do, let's do that; if that's trickier to do, it's probably not a huge worry if we're able to push to the other products' test channels.

PR #478, opened 2022-02-15
Bug 1751808 - Retry async productsign

hneiva:


iscript/src/iscript/mac.py
```diff
@@ -84,6 +84,23 @@ def check_required_attrs(self, required_attrs):
                 raise IScriptError("Missing {} attr!".format(att))
 
 
+def _retry_run_command_semaphore(semaphore, cmd, cwd, exception=IScriptError):
```
escapewindow: Nice :) Do we want to update the pkgbuild call to also use this helper?

escapewindow (approved):
The tox failure appears to be a git clone issue. We may want to update our taskgraph revision to something more recent, something that includes https://hg.mozilla.org/ci/taskgraph/rev/7ea98e00363a16c8233c931ee6e9e8b60771b18e .

iscript/tests/test_mac.py
```diff
@@ -786,16 +786,17 @@ async def fake_raise_future_exceptions(futures):
...
     """
 
-    async def fake_run_command(cmd, **kwargs):
+    async def fake_retry_async(**kwargs):
+        cmd = kwargs["kwargs"]["cmd"]
         assert cmd[0] in ("pkgbuild", "productbuild", "productsign")
-        if raises:
-            raise IScriptError("foo")
+        if should_raise:
+            raise kwargs["kwargs"]["exception"]("foo")
```
escapewindow, line 799: Ok, I was looking hard where this `exception` kwarg came from, and this is `run_command`'s `exception`.

Note that that kwarg is optional, so in theory this could hit a `KeyError` if not specified. Since we know the call stack and we always set `exception` in `_retry_run_cmd_semaphore`, this can work.

escapewindow (approved):
Thanks!

PR #479, opened 2022-02-16
Update taskgraph version/revision

hneiva:


gbrownmozilla (approved):
I think this should be okay, but this is mostly a rubberstamp.

Issue #481, opened 2022-02-17
ensure we have GITHUB_OAUTH_TOKEN set for level 3 github trust domains


escapewindow:
This token increases our Github API limits from 60 to [5000](https://docs.github.com/en/rest/overview/resources-in-the-rest-api#rate-limiting).

This is likely:

- find the worker pools where we have a level 3 (prod) pool pointed at a github repo, e.g. mobile, xpi, adhoc,
- ensure each of those pools has GITHUB_OAUTH_TOKEN set: `test_var_set 'GITHUB_OAUTH_TOKEN'` in `*script/docker.d/init_worker.sh`
- possibly double-check the relengworker sops repo to make sure we're setting `githubOauthToken` for those pools, to reduce bustage in the next step,
- deploy to production. These deploys will work if we have the oauth token set (and nothing else is broken), and fail if we don't. If we fail due to a missing oauth token, update the sops repo and redeploy.

escapewindow:
For the first two steps, we could even put this check in [this block](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/docker.d/init.sh#L13-L24): check for the `$COT_PRODUCT` and `"$ENV" == "prod"` .

hneiva:
Do we want to change the variable name here?
https://github.com/mozilla-releng/scriptworker-scripts/blob/master/githubscript/docker.d/init_worker.sh#L20

escapewindow:
> Do we want to change the variable name here? https://github.com/mozilla-releng/scriptworker-scripts/blob/master/githubscript/docker.d/init_worker.sh#L20

No, that's a separate read-write token for creating Github releases. The oauth token is a read-only token to prevent hitting API limits.

jcristau:
#482 added a check for l3 pools, #722 extended it to l1.  Closing.

PR #482, opened 2022-02-17
Add check for github oauth token

hneiva:


escapewindow (approved):
For posterity: this is to avoid hitting Github API limits during CoT verification and the like; we did so for xpi and several mobile graphs recently, and adding the oauth token to those pools resolved the issue. #481

PR #484, opened 2022-03-03
RELENG-847: Add Mozilla VPN scope prefix to push apk script

gabrielbusta:
Adds Mozilla VPN scope prefix to `pushapkscript/docker.d/worker.yml`


hneiva (approved):
Not sure what `$flatten:` does there. But looks good

gabrielbusta:
> Not sure what `$flatten:` does there. But looks good

Got it from here
https://github.com/mozilla-releng/scriptworker-scripts/blob/master/beetmoverscript/docker.d/worker.yml#L8

PR #485, opened 2022-03-03
Fix mozilla vpn scope prefix

gabrielbusta:
After trying it out with push apk script I realized it's getting the product name from the scope.


jcristau (approved):
We also need to add the scope to https://github.com/mozilla-releng/scriptworker/blob/master/src/scriptworker/constants.py#L385

PR #487, opened 2022-03-10
[VPN] Use local provisionprofile instead of URL

hneiva:


iscript/src/iscript/macvpn.py
```diff
@@ -60,12 +59,19 @@ async def _sign_app(config, sign_config, app, entitlements_url, provisioning_pro
...
         entitlements_url (str): URL for entitlements file
-        provisioning_profile_url (str): URL for provisioning profile
+        provisionprofile_filename (str): .provisionprofile filename in <config.provisionprofile_folder>
 
     """
     # We mock the task for downloading entitlements and provisioning profiles
     entitlements_path = await download_entitlements_file(config, sign_config, {"payload": {"entitlements-url": entitlements_url}})
-    provisioning_profile_path = await download_provisioning_profile(config, {"payload": {"provisioning-profile-url": provisioning_profile_url}})
+    provisioning_profile_path = None
+    if config.get("provisionprofile_folder"):
```
ahal: nit: `provisionprofile_dir`?

iscript/src/iscript/macvpn.py
```diff
@@ -60,12 +59,19 @@ async def _sign_app(config, sign_config, app, entitlements_url, provisioning_pro
...
+        provisioning_profile_path = os.path.join(config.get("provisionprofile_folder"), provisionprofile_filename)
```
ahal: nit: don't need to use `config.get` here

ahal (changes requested):
I think we should remove the `download_provisioning_profile` function as well? Or do you think we'll use that for other projects?

Otherwise LGTM!

iscript/src/iscript/macvpn.py
```diff
@@ -60,12 +59,19 @@ async def _sign_app(config, sign_config, app, entitlements_url, provisioning_pro
...
+        if not os.path.exists(provisioning_profile_path):
+            log.error(f"Could not find provisionprofile file: {provisioning_profile_path}")
+            raise IScriptError("Something")
```
ahal: Did you mean to update this to something more informative? :)

ahal (approved):
Thanks!

hneiva:
> I think we should remove the `download_provisioning_profile` function as well? Or do you think we'll use that for other projects?
> 
> Otherwise LGTM!

I'll write a ticket for it so we can check with gecko/other products.

Issue #474, opened 2022-02-10
[pushmsixscript] package not updated in new submissions


gbrownmozilla:
pushmsixscript is successfully creating new submissions for the Microsoft Store, but the new submission seems to re-use the package from the previous submission: The new msix files that we upload seem to disappear, and the new submission keeps serving up the old msix files.

gbrownmozilla:
pushmsixscript uses the REST API documented at 
https://docs.microsoft.com/en-us/windows/uwp/monetize/manage-app-submissions
and uses the python examples for inspiration:
https://docs.microsoft.com/en-us/windows/uwp/monetize/python-code-examples-for-the-windows-store-submission-api

Our application package currently consists of two msix files, one 32-bit and one 64-bit. pushmsixscript creates and updates the submission, then uploads each of the two msix files to the fileUploadUrl; each upload receives a 201 response code and no response body. pushmsixscript proceeds to commit the submission and monitors the commit status, typically seeing the status proceed from CommitStarted to PreProcessing. In the Partner Center, we can see the new submission and certification begins. However, the Package information for the new submission continues to reflect the original package -- not the new files uploaded.

gbrownmozilla:
https://docs.microsoft.com/en-us/windows/uwp/monetize/python-code-examples-for-the-windows-store-submission-api#create-app-submission states:
`# Upload images and packages in a zip file. Note that large file might need to be handled differently`
We are only interested in updating our app, with the 2 msix files. Should we put the msix files in a zip file and upload the zip? I've tried that but the upload consistently fails: "Connection reset by peer". Might that be related to the file size (275 MB)?

gbrownmozilla:
I also wonder whether the content of the submission request used to update the submission is correct. Currently pushmsixscript basically uses the submission request data returned from the create submission call, and sends that back to update submission. Is that okay, or should the submission request data be modified somehow?

Seann-Murdock:
Can you please provide a product id from one of your test runs so that I can take a look at the requests from the service side?

gbrownmozilla:
I have been testing with Store ID / Application ID 9NFQ5K7N2T6C ("Firefox Nightly"):

Package/Identity/Name | Mozilla.64253760D188F
Package/Identity/Publisher | CN=082E9164-EE6C-4EC8-B62C-441FAE7BEFA1

Seann-Murdock:
>I also wonder whether the content of the submission request used to update the submission is correct. 

[...]

PR #488, opened 2022-03-14
[pushmsixscript] Upload one zip file containing all msix files, and use azure blob storage client

gbrownmozilla:


gbrownmozilla:
This works great with small files, but doesn't work with the real firefox application.

gbrownmozilla:
> This works great with small files, but doesn't work with the real firefox application.

With the azure changes, it now works fine with the real firefox (verified with adhoc test against Nightly channel on the Store).

escapewindow (approved):
Thanks!

gbrownmozilla:
Closes #474

PR #490, opened 2022-03-15
Remove old notarization_poller requirements files

escapewindow:
We aren't using `py39.txt`-suffixed files anymore; let's clean them up
so dependabot doesn't ping me about them :)


jcristau (approved):

jcristau:
Do we use 3.8 anywhere still or could we drop that and add 3.10, at some point?

escapewindow:
We do use python 3.8 on the mac signers, though we point at `base.in` and pin requirements in [ronin_puppet](https://github.com/mozilla-platform-ops/ronin_puppet/blob/master/modules/signing_worker/files/update-requirements.sh). Because of relops turnover, the likelihood of us upgrading from 3.8 anytime soon is probably low.

Because we use py38 in production environments, that tells me we may want to keep some testing, but we could also add 3.10.

PR #493, opened 2022-03-24
iscript: remove 'NativeMessaging' inner app from Mac VPN signing

ahal:
The VPN client is being re-structured such that the NativeMessaging
module will no longer be a sub-package of the main app. Instead this
functionality is being provided by a binary file.

Jira: RELENG-878

### DIFF
--- a/iscript/src/iscript/macvpn.py
+++ b/iscript/src/iscript/macvpn.py
@@ -141,23 +141,6 @@ async def vpn_behavior(config, task, notarize=True):
         provisionprofile_filename="firefoxvpn_loginitem_developerid.provisionprofile",
     )
 
-    # NativeMessaging inner app
-    nativemessaging_app = App(
-        orig_path=os.path.join(submodule_dir, "NativeMessaging/MozillaVPNNativeMessaging.app"),
-        parent_dir=os.path.join(submodule_dir, "NativeMessaging"),  # Using main app as reference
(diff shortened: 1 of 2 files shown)


bakulf (approved):

bakulf:
We still need to cover the `wireguard-go` and `mozillavpnnp`, at least understand why it works :)

escapewindow:
Is this app removed from all shipping branches, or will we need to support signing apps with- and without- the inner app concurrently? (If the latter, then we probably want to key this off of a task definition flag.)

ahal:
> We still need to cover the `wireguard-go` and `mozillavpnnp`, at least understand why it works :)

Is it possible they're being signed as part of the build task? I saw you had linked to some `codesign` commands in Slack, would that not be doing it?

> Is this app removed from all shipping branches, or will we need to support signing apps with- and without- the inner app concurrently? (If the latter, then we probably want to key this off of a task definition flag.)

Would like @bakulf to chime in, but I believe there's only the one shipping branch. VPN has done forced updates for security fixes in the past.

ahal:
Before we merge this, I have an [adhoc-signing PR](https://github.com/mozilla-releng/adhoc-signing/pull/92) up to test this. The adhoc mac signer is already updated with the code in this PR. Once that's done, I'll pass the signed build on to @bakulf  and he can test whether `wireguard-go` and `mozillavpnnp` are working properly.

Assuming they work, I'll also leave the *why they're working* question up to him :)

PR #494, opened 2022-03-29
bug 1590599 - wrap winsign.sign.sign_file calls in retry_async

escapewindow:
I think we contact timestamp.digicert.com in this function, so we need
retries to be able to handle network/service blips.

### DIFF
--- a/signingscript/src/signingscript/sign.py
+++ b/signingscript/src/signingscript/sign.py
@@ -1271,6 +1271,12 @@ async def merge_omnija_files(orig, signed, to):
 
 
 # sign_authenticode_file {{{1
+async def _winsign_helper(error_message, *args, **kwargs):
+    """Raise an exception if winsign.sign.sign_file returns False to enable retries."""
+    if not await winsign.sign.sign_file(*args, **kwargs):
+        raise SigningScriptError(error_message)
+
+
 @time_async_function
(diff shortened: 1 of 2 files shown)


escapewindow:
Related: https://mozilla-hub.atlassian.net/browse/SVCSE-484

jcristau (approved):
lgtm

signingscript/src/signingscript/sign.py
```diff
@@ -1328,20 +1334,19 @@ async def signer(digest, digest_algo):
...
-        cafile=cafile,
-        timestampfile=timestampfile,
-        url=url,
-        comment=authenticode_comment,
-        crosscert=crosscert,
-        timestamp_style=timestamp_style,
-    ):
-        raise IOError(f"Couldn't sign {orig_path}")
+    # Retry winsign.sign.sign_file, because the timestamp server can hiccup
+    await retry_async(
```
gbrownmozilla, line 1338: You are happy with default args for retry_async attempts, sleeptime, etc? (Just checking - not suggesting a change.)

gbrownmozilla (approved):

signingscript/src/signingscript/sign.py
escapewindow: Yeah, I'm not sure which retry args to specify, and I generally default to the default unless I have an opinion.

PR #495, opened 2022-04-04
Fix VPN signing

hneiva:
Added provisioning profile
Added custom packaging logic for VPN
Added wireguard code signing


iscript/src/iscript/macvpn.py
```diff
@@ -44,7 +45,7 @@ async def _create_notarization_zipfile(work_dir, source, dest):
 
     """
     await run_command(
-        ["ditto", "-c", "-k", "--sequesterRsrc", "--keepParent", source, dest],
+        ["ditto", "-c", "-k", "--sequesterRsrc", source, dest],
```
ahal: The doc string says "keeps parent directory in zip" but I guess that's no longer true? Please update.

iscript/src/iscript/macvpn.py
```diff
@@ -64,20 +65,123 @@ async def _sign_app(config, sign_config, app, entitlements_url, provisionprofile
...
     entitlements_path = await download_entitlements_file(config, sign_config, {"payload": {"entitlements-url": entitlements_url}})
+
+    # TODO: Add provisionprofile_dir to scriptworker config
+    # https://github.com/mozilla-releng/scriptworker/blob/master/src/scriptworker/constants.py#L34
     provisioning_profile_path = None
-    if config.get("provisionprofile_dir"):
-        provisioning_profile_path = os.path.join(config["provisionprofile_dir"], provisionprofile_filename)
+
+    if provisionprofile_filename:
+        provisioning_profile_path = os.path.join("/builds/scriptworker/provisionprofiles", provisionprofile_filename)
```
ahal: Why does this need to be hardcoded again?

iscript/src/iscript/macvpn.py
```diff
@@ -64,20 +65,123 @@ async def _sign_app(config, sign_config, app, entitlements_url, provisionprofile
...
+    await retry_async(
+        run_command,
+        args=(sign_command,),
+        kwargs={"cwd": os.path.dirname(binary_path), "exception": IScriptError, "output_log_on_exception": True},
+        retry_exceptions=(IScriptError,),
+    )
+
+
+async def _create_pkg_files(config, sign_config, app):
+    """Create .pkg installer from the .app file. VPN specific behavior.
```
ahal: Could you elaborate what is different between this function and the one in `mac.py`? Is there any way we could have this one wrap that one instead of outright replacing it?

ahal (changes requested):
Thanks, this looks good to me! Just have a couple minor nits and questions I'd like answers to before it lands.

iscript/src/iscript/macvpn.py
hneiva: AIUI this is necessary for gecko because of the multiple-apps-single-zip method. Each source (work/0, work/1, work/2, etc) goes into it's own folder inside the zip.
I was trying to remove as many customized things as possible to make it work.
I'll revert this one to make it the same as gecko.

iscript/src/iscript/macvpn.py
hneiva: Ideally we'd add it to the [schema](https://github.com/mozilla-releng/scriptworker/blob/master/src/scriptworker/constants.py#L63) and modify script_config.yml with it. I'll make this a "default" instead

iscript/src/iscript/macvpn.py
hneiva: The arguments for the underlying commands are quite different. `mac.py` has options that work for gecko, and making it "generic" doesn't mean the next app will be able to use it either.

escapewindow: One way we solved for this was using [MAC_PRODUCT_CONFIG](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/iscript/src/iscript/constants.py#L4-L26); we could get our commands / arguments from that config dictionary, if our workflow is otherwise similar.

iscript/src/iscript/macvpn.py
ahal: I wasn't commenting on whether this is needed or not.. I'm just saying that the docstring (which is outside of this diff, you need to expand context), says "Keeps parent directory in zip".

The fact that you are removing this flag I think means that it is no longer true, so the docstring should be updated :)

iscript/src/iscript/constants.py
```diff
@@ -18,7 +18,7 @@
     "mozillavpn": {
         "designated_requirements": """=designated => certificate leaf[subject.OU] = "%(subject_ou)s" """,
         "sign_dirs": ("MacOS", "Frameworks", "Resources"),
-        "skip_dirs": ("MozillaVPNLoginItem.app",),
+        "skip_dirs": ("MozillaVPNLoginItem.app", "utils"),
```
ahal, line 21: We should probably at least file a follow-up issue to fix things so that `skip_dirs` normalizes all paths to be relative to the root of the package. Otherwise, this will skip all dirs named `utils` and there will be no way to specify which one.

iscript/src/iscript/macvpn.py
```diff
@@ -64,20 +65,133 @@ async def _sign_app(config, sign_config, app, entitlements_url, provisionprofile
     """
     # We mock the task for downloading entitlements and provisioning profiles
     entitlements_path = await download_entitlements_file(config, sign_config, {"payload": {"entitlements-url": entitlements_url}})
+
+    # TODO: Add provisionprofile_dir to scriptworker config
+    # https://github.com/mozilla-releng/scriptworker/blob/master/src/scriptworker/constants.py#L34
```
ahal: Looks like Aki might be pushing back on this. Just remember to remove this `TODO` if it doesn't end up landing there.

iscript/src/iscript/macvpn.py
```diff
@@ -64,20 +65,133 @@ async def _sign_app(config, sign_config, app, entitlements_url, provisionprofile
...
-            log.info(f"Using provisionprofile {provisioning_profile_path}")
+        log.info(f"Using provisionprofile {provisioning_profile_path}")
 
     await sign_all_apps(config, sign_config, entitlements_path, [app], provisioning_profile_path)
 
     log.info(f"Done signing app {app.app_name}")
 
 
+async def _sign_util(sign_config, binary_path):
+    """Sign command for VPN util binaries.
```
ahal, line 88: This function isn't actually specific to either the `utils` dir or binaries for that matter. Maybe it should just be called `_codesign`?

iscript/src/iscript/macvpn.py
```diff
@@ -64,20 +65,133 @@ async def _sign_app(config, sign_config, app, entitlements_url, provisionprofile
...
+    ]
+    await retry_async(
+        run_command,
+        args=(sign_command,),
+        kwargs={"cwd": os.path.dirname(binary_path), "exception": IScriptError, "output_log_on_exception": True},
+        retry_exceptions=(IScriptError,),
+    )
+
+
+async def _create_pkg_files(config, sign_config, app):
```
ahal, line 114: Sorry I forgot, but I why do we need a separate `create_pkg_files` again? If the `skip_dirs` thing is working, then shouldn't we be able to continue using the main one?

ahal (changes requested):
Looks good! Just a minor nit, and a question I wouldn't mind answered before this lands.

iscript/src/iscript/macvpn.py
ahal: Oh, I see you answered this in a previous review iteration... but in that case, how were we using it before? Wouldn't it be better to have a single function that can handle both?

iscript/src/iscript/constants.py
hneiva: https://mozilla-hub.atlassian.net/browse/RELENG-895 :)

iscript/src/iscript/macvpn.py
hneiva: I couldn't find an elegant way of making `create_pkg_files` generic and support both use cases.
The original function is tailored for gecko signing and VPN uses a different method of `pkgbuild` and `productbuild`.
Ideally we would split the logic so each external command is a separate function and manipulates the command input as needed based on configs/constants for each app.

ahal (approved):
Thanks!

PR #496, opened 2022-04-18
RELENG-790: add debsign support to signingscript

gabrielbusta:
These changes add debsign support to signingscript.
This is a [staging Linux signing task.](https://firefox-ci-tc.services.mozilla.com/tasks/TZGxVoEXSWaLQXN6uHXlog/runs/0)


signingscript/src/signingscript/sign.py
```diff
@@ -927,6 +931,44 @@ def make_signing_req(input_file, fmt, keyid=None, extension_id=None):
     return sign_req
 
 
+@time_function
+def make_files_signing_req(input_files, fmt, keyid=None, options=None):
+    """Make a signing request to pass to the /sign/files endpoint in autograph."""
+    signing_req = {"keyid": keyid, "options": options}
+    encoded_signing_req = BytesIO()
```
ahal: Might be cleaner to use `StringIO` here and then encode to bytes in the return statement?

signingscript/src/signingscript/sign.py
```diff
@@ -927,6 +931,44 @@ def make_signing_req(input_file, fmt, keyid=None, extension_id=None):
...
+    encoded_signing_req.write(b"[{")
+    for k, v in signing_req.items():
+        encoded_signing_req.write(json.dumps(k).encode("utf8"))
+        encoded_signing_req.write(b":")
+        encoded_signing_req.write(json.dumps(v).encode("utf8"))
+        encoded_signing_req.write(b",")
```
ahal: Why are we building a JSON object string by string here? Can't we build up `signing_req` as a dictionary and then return `json.dumps(signing_req).encode("utf-8")` at the end?

signingscript/src/signingscript/sign.py
```diff
@@ -860,12 +861,15 @@ def get_hawk_header(url, user, password, content_type, content_hash):
...
+async def call_autograph(session, url, user, password, sign_req, skip_encoding=False):
     """Call autograph and return the json response."""
     content_type = "application/json"
 
     request_body = tempfile.TemporaryFile("w+b")
-    write_signing_req_to_disk(request_body, sign_req)
+    if skip_encoding:
+        request_body.write(sign_req)
+    else:
+        write_signing_req_to_disk(request_body, sign_req)
```
ahal: Looks like `make_signing_req` just returns a non encoded dictionary. If `make_files_signing_req` also returned this, would we be able to drop this `skip_encoding` parameter?

ahal (changes requested):
I have a few questions here

signingscript/src/signingscript/sign.py
gabrielbusta: I'd have to change the code in write_signing_req_to_disk. [It's writing it to the file assuming we are using one of the single file endpoints](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/signingscript/src/signingscript/sign.py#L814)

signingscript/src/signingscript/sign.py
gabrielbusta: This is equivalent. Building the byte string this way is one of the things I tried when I couldn't figure out why I was getting 400s from autograph (I think it was an issue with "." in the file names after all)

[it's close to what write_signing_req_to_disk is doing](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/signingscript/src/signingscript/sign.py#L814)... I can simplify it... I think they only diff is that this is chunking the files to save some memory

signingscript/src/signingscript/sign.py
gabrielbusta: @ahal now that I think about it maybe [write_signing_req_to_disk](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/signingscript/src/signingscript/sign.py#L808) could be refactored to take a signing request and serialize it regardless of how it's formatted... but it is doing the file encoding... [which is why it needs to now which fields need to be base64 encoded](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/signingscript/src/signingscript/sign.py#L818)

signingscript/src/signingscript/sign.py
ahal: Ah, that makes sense.. I guess we sign some pretty big blobs and autograph needs the whole contents of the file in the request. By building it piece by piece we don't have the whole file loaded in memory at once.

signingscript/src/signingscript/sign.py
ahal: Yeah I think I would prefer this approach. The manual building of JSON feels a bit hacky, so the fewer places we do this the better.

On the other hand, modifying the function could potentially break other signing cases. On the other other hand, I think being too cautious about that and always duplicating code will make things less maintainable, and be more harmful in the long run.

So I'd vote we try and get `write_signing_req_to_disk` to support this signing req format. But we should probably wait for Aki to chime in here.

signingscript/src/signingscript/sign.py
ahal: Makes sense. As per my earlier comment I think that's the way to go.. But even if we don't go that route, it feels like we should make a duplicate `write_files_signing_req_to_disk` function rather than have your `make_files_signing_req` doing all the JSON building.

ahal:
Also, despite my comments, forgot to say thanks for tackling this! Looks like this is on the right track.

signingscript/src/signingscript/sign.py
gabrielbusta: Yea, that's was my thought too (that changing it too much might break other signing use cases)

signingscript/src/signingscript/sign.py
gabrielbusta: I made this func return a dict instead of the byte string to make it more orthogonal with `make_signing_req`

signingscript/src/signingscript/sign.py
gabrielbusta: I added a `write_files_signing_req_to_disk` function, I think this is more consistent with the rest of the code

signingscript/src/signingscript/sign.py
```diff
@@ -927,6 +939,15 @@ def make_signing_req(input_file, fmt, keyid=None, extension_id=None):
     return sign_req
 
 
+@time_function
+def make_files_signing_req(input_files, fmt, keyid=None, options=None):
+    """Make a signing request to pass to the /sign/files endpoint in autograph."""
+    signing_req = {"keyid": keyid, "options": options, "files": []}
+    for input_file in input_files:
+        signing_req["files"].append({"name": input_file.name, "content": b64encode(input_file.read())})
+    return signing_req
```
ahal: Maybe we can just merge this with `make_signing_req` now?
```
if  len(input_files) > 1:
    sign_req = {"input": input_files}
else:
    sign_req = {"files": []}
    for input_file in input_files:
        ...

Comment (ahal) [reply signingscript/src/signingscript/sign.py]: Though I'm not sure what `options` is for.. If that's only something used by the files endpoint, then maybe keeping them separate is better...

File: signingscript/src/signingscript/sign.py [hunk 5]
```diff
@@ -834,6 +835,14 @@ def write_signing_req_to_disk(fp, signing_req):
     fp.write(b"}]")
 
 
+def write_files_signing_req_to_disk(fp, signing_req):
+    """Write files/ signing_req to fp.
+    The files content should be ecoded as base64 strings
+    """
+    encoded_signing_req = json.dumps(signing_req).encode("utf-8")
+    fp.write(encoded_signing_req)
```
ahal: Won't we potentially hit the memory issues that `write_signing_req_to_disk` was trying to solve now? Maybe there's a good reason to be less concerned in this case?

gabrielbusta: True. I replaced this with a chunking version

signingscript/src/signingscript/sign.py
gabrielbusta: `/sign/files/` does have an option field I combined these

gabrielBusta:
The diff on `write_signing_req_to_disk` is a bit convoluted [here's a link to the function](https://github.com/mozilla-releng/scriptworker-scripts/blob/4809009023eeb03bf74887de763c59590be1e444/signingscript/src/signingscript/sign.py#L815)

I didn't change the existing logic, but I added support for requests to /sign/files

signingscript/src/signingscript/sign.py
```diff
@@ -811,27 +812,46 @@ def write_signing_req_to_disk(fp, signing_req):
...
-                e = b64encode(block).encode("utf8")
-                fp.write(e)
-            fp.write(b'"')
-        else:
-            fp.write(json.dumps(v).encode("utf8"))
-        fp.write(b",")
-    fp.seek(-1, 1)
-    fp.write(b"}]")
+                encoded_block = b64encode(block)
+                encoded_file_content.write(encoded_block)
```
ahal: I see a bit of potential for factoring out here, maybe just a `def _encode_single_file()` inner function that can be shared with below. But not a big deal to leave as is either.

ahal (approved):
Nice thanks! Let's also block on @escapewindow's review though.

signingscript/docker.d/passwords.yml
```diff
@@ -144,6 +144,12 @@ in:
              {"$eval": "AUTOGRAPH_AUTHENTICODE_PASSWORD"},
              ["autograph_authenticode", "autograph_authenticode_stub"]
             ]
+          - ["https://autograph-external.stage.autograph.services.mozaws.net",
```
escapewindow: We should use Production Autograph for dep/dev unless we're specifically testing the Staging Autograph instance.
Essentially, the Autograph team reserves the right to bring the Staging Autograph instance down for hours or days at a time if needed, but we need PRs, CI signing tasks, and staging releases to work more reliably than that. Therefore we have a "fake", "dep", self-signed key on the production cluster.

signingscript/docker.d/passwords.yml
```diff
@@ -144,6 +144,12 @@ in:
...
+             {"$eval": "AUTOGRAPH_MOZILLAVPN_DEBSIGN_USERNAME"},
+             {"$eval": "AUTOGRAPH_MOZILLAVPN_DEBSIGN_PASSWORD"},
+             ["autograph_debsign"],
+             "dummy_gpg2_pgpsubkey_debsign"
```
escapewindow: I believe the keyid is default, so we don't need to specify, but we can. If we do specify, this would be `release_at_mozilla_debsign_dep`.

signingscript/src/signingscript/sign.py
```diff
@@ -811,27 +812,49 @@ def write_signing_req_to_disk(fp, signing_req):
     Does proper base64 and json encoding.
     Tries not to hold onto a lot of memory.
     """
```
escapewindow: As mentioned in channel: I'm not 100% sure what we're doing in `write_signing_req_to_disk` and why. I believe this affects all autograph signing formats, and when we roll out we'll affect all trains/products/releases, so we probably a) want to do more testing than just debsign testing before rolling this out, and b) choose a non-release non-nightly time and do some doublechecking before the next nightly/release.

This may be worth pushing to `dev-signingscript` if we haven't already (I'm guessing you have already :)
When we do, we can also try some other test signing tasks, e.g. point some gecko Windows or other signing tasks at the dev pool.

Also, if we're able to add some unit tests to this function before and after the refactor, that may help us reason about the changes here a bit better. (I'm thinking we have some example signing reqs, and some expected file contents to diff against.)

escapewindow (changes requested):

signingscript/src/signingscript/sign.py
gabrielbusta: I am not sure why it builds the byte string this way. So I didn't change it. I refactored the code (renamed `write_signing_req_to_disk` to `_encode_single_file`) so we can see the logic hasn't changed in the diff. The new code I had was getting 400s from autograph... I reverted the logic to be equivalent to the original commit (build the byte string in a similar way as `_encode_single_file`). Autograph is happy now.

`dev-signingscript` is in sync with this branch ([here's a task completed by the latest version of the code](https://firefox-ci-tc.services.mozilla.com/tasks/cPfKdVwvQzuI8gW9aWKcSg))

I will add a unit test. From the coverage report it seems most of the uncovered code it's the new code I added.

signingscript/src/signingscript/sign.py
gabrielbusta: This cleaned up nicely thanks :)

signingscript/src/signingscript/sign.py
```diff
@@ -834,6 +835,56 @@ def write_signing_req_to_disk(fp, signing_req):
     fp.write(b"}]")
 
 
+def _encode_multiple_files(fp, signing_req):
```
escapewindow, line 838: If you've figured these functions out, it might be helpful for the next person if you add a docstring here (and possibly flesh out the docstring above in `_encode_single_file`) describing what it does. (The `why` might be super helpful too, but I'm not sure if we understand that yet?)

escapewindow (approved):
Thanks!
Tests for `write_signing_req_to_disk` specifically may be helpful if we ever want to touch these `_encode_*` functions again -- I'm thinking:

- tiny file(s) to sign
- call the `write_signing_task_to_disk`
- diff against a known signing req (possibly written to the data dir)

I think that would help us; what do you think?

signingscript/src/signingscript/sign.py
gabrielbusta: I added a doc string

gabrielbusta:
> Thanks! Tests for `write_signing_req_to_disk` specifically may be helpful if we ever want to touch these `_encode_*` functions again -- I'm thinking:
> 
>     * tiny file(s) to sign
> 
>     * call the `write_signing_task_to_disk`
> 
>     * diff against a known signing req (possibly written to the data dir)
> 
> 
> I think that would help us; what do you think?

I think this would be helpful. I added the test. The files end up as `BufferedRandom` objects. So I created instances of those to simulate tiny files.

PR #503, opened 2022-05-09
Add autograph_authenticode_sha2_rfc3161_stub format

gbrownmozilla:


gbrownmozilla:
This is ugly, but seems expedient and safe.

escapewindow (approved):
Nice :)

We may want to also be able to specify different rfc3161 timestamp servers, which may either require `:` delimited format strings or even more defined formats, but this should work for us for just testing the one timestamp server.

PR #504, opened 2022-05-10
Use certum for timestamp server, prod firefox only, temporarily

gbrownmozilla:
This is a temporary, experimental change, to be landed in production for 1 nightly, then backed out before the next beta.


escapewindow (approved):
I'm guessing you'll roll this out after today's beta finishes?

gbrownmozilla:
Hoping to do it before today's beta, but will be cautious about that and make sure everyone is on board first.

PR #507, opened 2022-05-12
Add a config option to insert extra intermediate cert with authentico…

jcristau:
…de signature

If the authenticode_add_digicert_cross option is set (initially only for
adhoc), after signing a file for authenticode we check whether the
embedded certificates chain up to "DigiCert Trusted Root G4" instead of
"DigiCert Global Root CA", and if so, insert the extra cert, because the
G4 root is not trusted by some Windows 7 installs, so it complains about
lack of trust path to the file's timestamp.


jcristau:
I ended up making this a config option instead of a task param, that might be silly though...

escapewindow:
> I ended up making this a config option instead of a task param, that might be silly though...

If this means what I think it means, then we'll only be able to roll out by pool. So all of adhoc prod, and/or all of gecko prod. If we want to ride the trains, we'll want a task param.

signingscript/docker.d/init_worker.sh
```diff
@@ -42,6 +42,11 @@ if [ "$ENV" == "prod" ]; then
   export AUTHENTICODE_CA_PATH=/app/signingscript/src/signingscript/data/authenticode_prod_202005.crt
   export AUTHENTICODE_CA_TIMESTAMP_PATH=/usr/lib/ssl/certs/ca-certificates.crt
 fi
+if [ "$COT_PRODUCT" = "adhoc" ]; then
```
escapewindow, line 45: We could also put this in the `case $COT_PRODUCT in` below.
One potential downside here is we won't be able to adhoc-sign without this option, but since we're only appending the additional cert when we have a specific root, maybe that's not too bad?

gbrownmozilla (approved):
I liked the digicert approach of using a proxy server. It had the advantage of not needing osslsigncode. But that came with the cost of managing the server, plus I couldn't get it to work for all cases...probably this is better!

escapewindow (approved):
Thank you!
Let's try this :)
If we want to move this to the task.payload, that can be a followup. If we need to adhoc-sign something without this, we can roll out the previous revision to production-signingscript.

signingscript/docker.d/init_worker.sh
gbrownmozilla, line 45: Oops! = vs == here??

PR #527, opened 2022-06-10
Bug 1773380 - use same character encoding for submission data as found in response when getting data

gbrownmozilla:


gbrownmozilla:
I tested this approach on the release channel and checked the pending submission descriptions for various locales: checked the U+2019 in en-US, checked accented characters for french, and checked that the greek description looked greek to me.

gbrownmozilla:
response.encoding seems to always be utf-8 for this Store request; I don't imagine it would be anything else.

escapewindow (approved):
This looks sensible given the failure cases we know about, thank you!

(I'm guessing you're learning a lot by having to deal with these failure cases in production; do you agree? :)

pushmsixscript/src/pushmsixscript/microsoft_store.py
```diff
@@ -200,7 +201,7 @@ def _update_submission(config, channel, session, submission_request, headers, fi
     submission_request = str(submission_request)
     submission_request = submission_request.replace("True", "true").replace("False", "false")
     url = _store_url(config, f"{application_id}/submissions/{submission_id}")
-    response = session.put(url, submission_request, headers=headers)
+    response = session.put(url, submission_request.encode(encoding), headers=headers)
```
jcristau, line 204: json is always utf-8, can we just hardcode the encoding here and leave everything else alone?

gbrownmozilla: Is it always utf-8? rfc 7159 says it is utf-8 by default but can be utf-16 or utf-32.

I don't mind assuming utf-8 and in practice I suspect that will be fine, but this way seems possibly-safer.

jcristau: my source is wikipedia so yeah, your fix is safer :)

gbrownmozilla:
> (I'm guessing you're learning a lot by having to deal with these failure cases in production; do you agree? :)

I wouldn't mind learning a little less! :)

PR #530, opened 2022-06-20
Adjust pre-stop polling duration for treescript (bug 1771735)

jcristau:
Some treescript tasks (merges) take longer than 20 minutes so it has a
long `terminationGracePeriodSeconds` value; pre-stop needs to be
adjusted to match, otherwise tasks get killed too eagerly.

### DIFF
--- a/docker.d/pre-stop.sh
+++ b/docker.d/pre-stop.sh
@@ -2,15 +2,23 @@
 
 # scriptworker runs as PID 1 in Kubernetes
 SCRIPTWORKER_PID=${SCRIPTWORKER_PID:-1}
-# See https://kubernetes.io/docs/concepts/workloads/pods/pod/?source=post_page---------------------------#termination-of-pods
+# See https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination
 # for the details
 # Kubernetes executes this script synchronously and waits for it to finish. If
 # the script doesn't finish in `terminationGracePeriodSeconds`, it sends
 # SIGTERM, waits 2 seconds and kills the container. To prevent this scenario
 # in case the task takes too long, this script exits 2 minutes before
-# `terminationGracePeriodSeconds` to let scriptworker upload files and report
-# `machine-shutdown` to Taskcluster.
-POLL_DURATION=1080
+# `terminationGracePeriodSeconds` (as set e.g. in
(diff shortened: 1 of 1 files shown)


docker.d/pre-stop.sh
```diff
@@ -10,7 +10,14 @@ SCRIPTWORKER_PID=${SCRIPTWORKER_PID:-1}
 # in case the task takes too long, this script exits 2 minutes before
 # `terminationGracePeriodSeconds` to let scriptworker upload files and report
 # `machine-shutdown` to Taskcluster.
-POLL_DURATION=1080
+case ${PROJECT_NAME} in
+    tree)
+        POLL_DURATION=3480
```
JohanLorenzo, line 16: Nit: let's add a comment to remind ourselves about the value duplication

```suggestion
        # Keep this value in sync with [1] so that it's 2 minutes shorter than the terminationGracePeriodSeconds
        # [1] https://github.com/mozilla-services/cloudops-infra/blob/d94d5fd6a7704ffd2c829d870206f5c0ed8d75e7/projects/relengworker/k8s/charts/tree/templates/deployment.yaml#L35
        POLL_DURATION=3480
```

docker.d/pre-stop.sh
```diff
@@ -10,7 +10,14 @@ SCRIPTWORKER_PID=${SCRIPTWORKER_PID:-1}
...
+        ;;
+    *)
+        POLL_DURATION=1080
```
JohanLorenzo, line 19:
```suggestion
        # Keep this value in sync with all `templates/deployment.yaml` under [2]
        # [2] https://github.com/mozilla-services/cloudops-infra/blob/d94d5fd6a7704ffd2c829d870206f5c0ed8d75e7/projects/relengworker/k8s/charts
        POLL_DURATION=1080
```

JohanLorenzo (approved):
LGTM! Let's add a few more comments to let our future selves know about the link between `POLL_DURATION` and `terminationGracePeriodSeconds`

docker.d/pre-stop.sh
jcristau: That feels like a bit of a repeat of the existing comment?

JohanLorenzo: A little bit. Although, the comment above doesn't say where `terminationGracePeriodSeconds` lives and is not super clear that we must manually keep them in sync. We could update the comment above if that helps having a single bloc of text.

PR #534, opened 2022-06-28
[pushmsixscript] Bug 1776696 - skip commit on release channel

gbrownmozilla:


pushmsixscript/src/pushmsixscript/microsoft_store.py
```diff
@@ -109,10 +109,15 @@ def _push_to_store(config, channel, msix_file_paths, publish_mode, access_token)
         submission_id = submission_request.get("id")
         log.info(">> updating the submission...")
         _update_submission(config, channel, session, submission_request, headers, msix_file_paths, publish_mode, encoding)
-        log.info(">> committing the submission...")
-        _commit_submission(config, channel, session, submission_id, headers)
-        log.info(">> waiting for completion...")
-        _wait_for_commit_completion(config, channel, session, submission_id, headers)
+        if channel == "release":
```
escapewindow, line 112: We can hardcode this, but a good followup may be defining products, channels, and their respective behaviors in the script_config.yml (worker.yml in docker.d). e.g.

```
# script_config.yml
supported_channels:
    release:
        do_commit: false

# code
if channel_config.get("do_commit"):
    ...
```

escapewindow (approved):

pushmsixscript/src/pushmsixscript/microsoft_store.py
gbrownmozilla: Yes, I'll keep that in mind for a follow-up. For now, I am not sure if this change is going to work out (can envision another relman request in a few weeks) so keeping it simple.

PR #535, opened 2022-06-29
Bump dependencies

jcristau:


jcristau:
There's a major version bump for protobuf (https://developers.google.com/protocol-buffers/docs/news/2022-05-06#python-updates) but we don't actually use that AFAICT.
Other than that, s3transfer dropped python3.6 support, and everything else is minor version bumps.

jcristau:
I should note, the main reason for this is picking up new scriptworker.

JohanLorenzo (approved):
LGTM! l'm surpised no new transitive dependencies were added even though many deps have been bumped!

PR #539, opened 2022-07-11
RELENG-11 - Bump Android-Components versions

JohanLorenzo:
Fixes this error[1]:

```
mozilla_version.errors.PatternNotMatchedError: "104.0.0" does not match the pattern: Minor number and patch number cannot be both equal to 0
```

The error itself is caused by [bug 1777255](https://bugzilla.mozilla.org/show_bug.cgi?id=1777255). However, this is isn't the root cause. Android-Components follow the Maven version number scheme. Thus, I made the changes to follow that. 

Tested at [2]. By the way, `outgoing.diff` is now exposed and contains:

```diff
diff --git a/version.txt b/version.txt
index 18cf10229f..5fa45906ac 100644
--- a/version.txt
+++ b/version.txt
@@ -1 +1 @@
-104.0.0
+104.0.1
```

I also took the liberty of pruning `FennecVersion` since Fennec doesn't exist anymore. 

[1] https://firefox-ci-tc.services.mozilla.com/tasks/fMYm7gRiTgCpGLNzUyo8Xg/runs/0/logs/public/logs/live_backing.log
[2] https://firefox-ci-tc.services.mozilla.com/tasks/I07q0GGOQPSy-KLV5otYtw/runs/2

### DIFF
--- a/treescript/src/treescript/versionmanip.py
+++ b/treescript/src/treescript/versionmanip.py
@@ -5,7 +5,7 @@ import logging
 import os
 
 from mozilla_version.fenix import FenixVersion
-from mozilla_version.gecko import FennecVersion, FirefoxVersion, GeckoVersion, ThunderbirdVersion
+from mozilla_version.gecko import FirefoxVersion, GeckoVersion, ThunderbirdVersion
 
 from treescript.exceptions import TaskVerificationError, TreeScriptError
 from treescript.task import DONTBUILD_MSG, get_dontbuild, get_vcs_module, get_version_bump_info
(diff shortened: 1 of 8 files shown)


treescript/src/treescript/versionmanip.py
```diff
@@ -19,27 +20,41 @@
...
+    "android-components": MavenVersion,
+}
+
+
+def _find_what_version_parser_to_use(file_, source_repo):
+    version_classes = [cls for path, cls in _VERSION_CLASS_PER_BEGINNING_OF_PATH.items() if file_.startswith(path)]
+
+    number_of_version_classes = len(version_classes)
+    if number_of_version_classes > 0:
+        if number_of_version_classes > 1:
```
jcristau: Maybe reorder to avoid nesting the `if`s

jcristau (approved):

treescript/src/treescript/versionmanip.py
JohanLorenzo: Sounds good, done!

PR #540, opened 2022-07-14
[iscript] Bug 1774669 - Add mac_single_file (openh264, geckodriver) notarization

gbrownmozilla:


gbrownmozilla:
Do I need to avoid notarization on dep, like https://github.com/mozilla-releng/scriptworker-scripts/blob/master/iscript/src/iscript/script.py#L35-L39 ?

gbrownmozilla:
I think we can add geckodriver notarization fairly easily, but we'll need to untar, zip and notarize the zip. Is that desired? (I would handle that as a separate bug.)

escapewindow:
> Do I need to avoid notarization on dep, like https://github.com/mozilla-releng/scriptworker-scripts/blob/master/iscript/src/iscript/script.py#L35-L39 ?

Yes, Apple will only notarize files signed with valid keys. Dep signing uses a self-signed cert, so to Apple, it's not a valid key.

> I think we can add geckodriver notarization fairly easily, but we'll need to untar, zip and notarize the zip. Is that desired? (I would handle that as a separate bug.)

Huh, I was under the impression we needed a .dmg for notarization to work properly, but the doc in the bug says differently.

I do think that's desired, per https://bugzilla.mozilla.org/show_bug.cgi?id=1588081, as long as it's not too much extra work.

escapewindow (changes requested):
Thanks!
Let's block on not notarizing if we're dep-signing.

Were you able to test this on a mac signer?
Did this result in a working openh264 if we release-signed and notarized? If so, amazing :)

gbrownmozilla:
Yes, I have been able to test on a mac signer, release signed and notarized. Logs look good: "Status Message: Package Approved", etc. What else should I do to check that the result is "working"?

I'll sort out dep signing tomorrow.

escapewindow:
> Yes, I have been able to test on a mac signer, release signed and notarized. Logs look good: "Status Message: Package Approved", etc. What else should I do to check that the result is "working"?
> 
> I'll sort out dep signing tomorrow.

I'm not sure what all :jrmuizel and :csasca did to test; it may be worth sending them a signed file and asking them if it works.

gbrownmozilla:
I believe I'll need to add the new behavior to ronin_puppet and then update the openh264 behaviors in taskgraph to start using openh264 notarization.

escapewindow (approved):
> I believe I'll need to add the new behavior to ronin_puppet and then update the openh264 behaviors in taskgraph to start using openh264 notarization.

+1.
We probably want to add this to gecko and possibly adhoc-signing once those are rolled out.
Did we verify the signature works in real life?

gbrownmozilla:
> Did we verify the signature works in real life?

I'm waiting to hear back from jrmuizel.

gbrownmozilla:
> > Did we verify the signature works in real life?
> 
> I'm waiting to hear back from jrmuizel.

jrmuizel confirms: "Yep, the binaries in the zip work, even with the quarantine attribute set."

PR #543, opened 2022-07-22
[iscript] Bug 1588081 - Add notarization of single file geckodriver

gbrownmozilla:


gbrownmozilla:
This is a lot like the openh264 notarization. We can't directly notarize either the tar file or the geckodriver executable, but wrapping geckodriver in a zip file works fine.

Similar to the wireguard case ([PR 326](https://github.com/mozilla-releng/scriptworker-scripts/pull/326)) hardened_runtime_only_files needs to be used, to sign with "-o runtime", else notarization fails.

I have successfully notarized geckodriver manually on a mac signer with this code.

This won't change behavior for mac_geckodriver, but adds notarization for mac_notarize_geckodriver. I want to check in with whimboo (on pto for a while) before enabling.

iscript/src/iscript/constants.py
```diff
@@ -14,6 +14,7 @@
         "skip_dirs": tuple(),
         "zipfile_cmd": "zip",
         "create_pkg": True,
+        "hardened_runtime_only_files": "geckodriver",
```
escapewindow, line 17: I was wondering if we might break something for mac Firefox signing, but [looking at the code](https://github.com/mozilla-releng/scriptworker-scripts/blob/9946f96ac5dc203b096c66dc76886447a0d0d2d3/iscript/src/iscript/mac.py#L175) we're probably good.

iscript/src/iscript/mac.py
```diff
@@ -194,7 +194,7 @@ async def sign_single_files(config, sign_config, all_paths):
     """
     identity = sign_config["identity"]
     keychain = sign_config["signing_keychain"]
-    sign_command = _get_sign_command(identity, keychain, sign_config)
+    sign_command = _get_sign_command(identity, keychain, sign_config, file_=all_paths[0].single_file_globs[0])
```
escapewindow: Hm, are we sure we'll never have multiple `all_paths`?
If so, we may want to move this down into the 2nd `for app in all_paths:` below.

iscript/src/iscript/script.py
```diff
@@ -39,6 +39,8 @@ def check_dep_behavior(task, behavior, supported_behaviors):
         behavior = "mac_sign_and_pkg_vpn"
     if behavior == "mac_notarize_single_file" and "mac_single_file" in supported_behaviors:
         behavior = "mac_single_file"
+    if behavior == "mac_notarize_geckodriver" and "mac_geckodriver" in supported_behaviors:
+        behavior = "mac_geckodriver"
```
escapewindow, line 43: We could change this to a

```
{
    "orig_format": "fallback_format",
    ...
}
```

tyep config and have a single `if` statement, but this works :)

escapewindow (approved):

iscript/src/iscript/mac.py
gbrownmozilla: I am pretty sure -- this is the single_file case after all. But I still like the idea - thanks!

iscript/src/iscript/script.py
gbrownmozilla: Maybe if we add one more case? For now, I'm happy with the if's.

iscript/src/iscript/constants.py
gbrownmozilla: I think so too.

PR #545, opened 2022-07-25
Bump dependencies

gbrownmozilla:


JohanLorenzo (approved):
Mostly a rubber stamp from me. I just looked at what files are changed and glanced over some of the version bumps. It looks good. Thanks for taking the time to update these deps! 👍

PR #547, opened 2022-07-29
[pushmsixscript] Bug 1779435 - Always commit; specify a default rollout percent for release submissions

gbrownmozilla:


escapewindow (approved):
Not necessarily a blocker, but a suggestion:
I find this is useful to have in a config file, in this case `script_config.yml` (or `worker.yml` as it's called in `docker.d`).

E.g.

```yaml
channel_config:
    release:
        package_rollout: true
        package_rollout_percentage: 25
    default:
        commit_submission: true
```

If these config items are available and documented and the code handles them in an intuitive/logical manner, this makes it easier for the rest of the team to make these changes when you're not around. Otherwise they need to reason about what portions of code need to change and make changes that may or may not be sufficient for everything.

gbrownmozilla:
Good point. But commit_submission really should not be an option: we know now that it doesn't work. 

I've gone with a slight short-cut for the rollout %, since we don't have per-channel config, and I suspect rollout % will always be release-only.

PR #549, opened 2022-08-04
Fix #548 - Update signingscript to use the "browser_specific_settings"

### LINKED ISSUE #548: Signing a privileged add-on fails if deprecated manifest `applications` key is missing
For privileged extensions, the signing script expects an `applications` key in the extension manifest. However, that key has been deprecated and renamed to `browser_specific_settings`. The script should check for the latter first and fall back to `applications`. Only if neither are present, we should fail.

Ref: https://firefox-ci-tc.services.mozilla.com/tasks/groups/ZS6EBHxEQZaDZbEigcKIaQ

gabrielbusta:
Update signingscript to use the "browser_specific_settings" key in the xpi manifest. Fall back to deprecated "applications" key.

### DIFF
--- a/signingscript/src/signingscript/sign.py
+++ b/signingscript/src/signingscript/sign.py
@@ -527,13 +527,14 @@ def _extension_id(filename, fmt):
                 break
         except KeyError:
             log.debug("{} doesn't exist in {}...".format(manifest_name, filename))
-    if not manifest.get("applications", {}).get("gecko", {}).get("id"):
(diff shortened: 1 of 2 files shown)


escapewindow (approved):
Perfect, thank you!
It would be good practice to add a set of parameters to test the browser_specific_settings [here](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/signingscript/tests/test_sign.py#L983-L1012).

Once we land, let's roll out a new set of production signing scriptworkers at a quiet time.

PR #555, opened 2022-08-18
[pushmsixscript] Bug 1785907 - Fix publishMode lookup

gbrownmozilla:


pushmsixscript/src/pushmsixscript/script.py
```diff
@@ -20,7 +20,11 @@ async def async_main(config, task_dict):
     channel = task.get_msix_channel(config, task_dict)
     _log_warning_forewords(config, channel)
 
-    publish_mode = task_dict.get("publishMode")
+    payload = task_dict.get("payload")
+    if payload:
```
escapewindow: I'm guessing payload isn't specified in tests, or sometimes we get real tasks without payloads?

We could also spell this as

```
publish_mode = task_dict.get("payload", {}).get("publishMode")
```

which would be `None` if either `payload` or `payload.publishMode` weren't defined.

escapewindow (approved):

pushmsixscript/src/pushmsixscript/script.py
gbrownmozilla: Yeah, no "payload" in some tests. I like your more compact form, will update - thanks!

PR #556, opened 2022-08-23
[RELENG-966] Add config to test VPN addon signing against autograph s…

ahal:
…tage

### DIFF
--- a/signingscript/docker.d/init_worker.sh
+++ b/signingscript/docker.d/init_worker.sh
@@ -211,6 +211,8 @@ case $ENV in
         test_var_set 'AUTOGRAPH_AUTHENTICODE_USERNAME'
         test_var_set 'AUTOGRAPH_MOZILLAVPN_PASSWORD'
         test_var_set 'AUTOGRAPH_MOZILLAVPN_USERNAME'
+        test_var_set 'AUTOGRAPH_MOZILLAVPN_ADDONS_PASSWORD'
+        test_var_set 'AUTOGRAPH_MOZILLAVPN_ADDONS_USERNAME'
         test_var_set 'AUTHENTICODE_CERT_PATH'
         test_var_set 'AUTHENTICODE_CA_PATH'
         test_var_set 'AUTHENTICODE_CA_TIMESTAMP_PATH'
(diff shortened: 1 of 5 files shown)


ahal:
This is a WIP and not yet ready to land. We'll need a corresponding change in cloudops-infra and to add the keys in SOPS first.

ahal:
Creds have been added to relengworker-sops and here's the cloudops-infra PR: https://github.com/mozilla-services/cloudops-infra/pull/4291

I believe that has to land first, and then we can deploy the signingscript change.

ahal:
I'll leave this as a draft for now, but Aki.. is this what you had in mind in https://mozilla.slack.com/archives/CEMMGTZJ5/p1660761121876099 ?

escapewindow:
> I'll leave this as a draft for now, but Aki.. is this what you had in mind in https://mozilla.slack.com/archives/CEMMGTZJ5/p1660761121876099 ?

This looks good!

We probably don't need this format to test against autograph-stage for the long term, just to make sure that the signer config+code work. So we could push this commit to dev-signingscript to test, without merging to master, if we want to keep it clean. Otherwise we can merge to master, push to dev-signingscript, and when we're done, we can remove it from master. (For the long term, we would add autograph-prod dep creds to dev+dep signingscript, and autograph-prod release creds to prod signingscript.)

ahal:
Good idea, I'll try keeping this out of master. This PR will still be needed once the signer merges to autograph prod.. Will just need to s/stage/prod.

ahal:
Still pointing at autograph stage for now, but other than changing that when the signer merges to autograph prod, this should be what I'll be submitting for review.

ahal:
I'd like to get the review process started here even though the signer isn't moved over to autograph prod yet. We can hold off landing until this happens (though I think it technically *could* land ahead of time.. it's just any tasks using this format would fail).

escapewindow (commented):
Looks good!
For multi-format signing, I might slot `autograph_rsa` between `macapp` and `gpg` in [_sort_formats](https://github.com/mozilla-releng/scriptworker-scripts/blob/23d0314ca171e3b719b22b296f3da2566baccb19/signingscript/src/signingscript/task.py#L165-L185) in case we ever do create a detached rsa signature after doing some other signing that modifies the binary. That's my only real thought. With that, looks like we're ready to land, though we could hold off til we add the prod/dep script_config.yml (aka worker.yml) changes if wanted.

escapewindow (approved):

gabrielbusta (approved):

PR #558, opened 2022-08-26
bouncerscript: remove some leftovers from fennec

jcristau:
Fennec was EOL mid-2020.


escapewindow (approved):
Looks like the decision task is green, and Github didn't pick it up for some reason.

PR #560, opened 2022-08-31
Add MIME types for Qt .rcc and .sig files in beetmoverscript

gabrielbusta:
The VPN addons are Qt `.rcc` files ([see](https://doc.qt.io/qtforpython/overviews/resources.html#the-qt-resource-compiler-rcc)) These are compiled binary files so `application/octet-stream` seems appropriate. 

The `.sig` file is the detached signature for the addons manifest. Looking at [@ahal's WIP](https://github.com/mozilla-releng/scriptworker-scripts/commit/7ecceb928257aed0a96ca8f372b3b354885663f7), I think `application/octet-stream` is the right MIME type for that too.

### DIFF
--- a/beetmoverscript/src/beetmoverscript/constants.py
+++ b/beetmoverscript/src/beetmoverscript/constants.py
@@ -18,9 +18,11 @@ MIME_MAP = {
     ".msix": "application/msix",
     ".pkg": "application/x-newton-compatible-pkg",
     ".pom": "application/xml",
+    ".rcc": "application/octet-stream",
     ".sha1": "text/plain",
     ".sha256": "text/plain",
     ".sha512": "text/plain",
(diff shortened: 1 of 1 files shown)


beetmoverscript/src/beetmoverscript/constants.py
```diff
@@ -23,6 +23,8 @@
     ".sha512": "text/plain",
     ".snap": "application/octet-stream",
     ".xpi": "application/x-xpinstall",
+    ".rcc": "application/octet-stream",
+    ".sig": "application/octet-stream",
```
ahal: nit: looks like this is alphabetized, please put these in alphabtetic order.

ahal (approved):

PR #561, opened 2022-09-02
[pushmsixscript] Bug 1788793 - Fail task if commit reports CommitFailed

gbrownmozilla:


gbrownmozilla:
I found that test_get_submission_status was not actually checking the status, because mocked_response was redefined - fixed.

JohanLorenzo (approved):
🚢  it! 👍

pushmsixscript/src/pushmsixscript/microsoft_store.py
```diff
@@ -265,4 +265,12 @@ def _wait_for_commit_completion(config, channel, session, submission_id, headers
         time.sleep(COMMIT_POLL_WAIT_SECONDS)
         response_json = _get_submission_status(config, channel, session, submission_id, headers)
         log.info(response_json.get("status"))
+    if response_json.get("status") == "CommitFailed":
```
jcristau: Are there other status values that may be unexpected?  https://docs.microsoft.com/en-us/windows/uwp/monetize/get-status-for-an-app-submission#response-body has a long list (but I guess some of those aren't relevant for us), I was wondering whether we could assert the actual status is what we expect.

jcristau: Also, we may want to log the statusDetails?

gbrownmozilla: We decided a while ago not to wait around for commit completion: completion times vary a lot, and can take a long time (at least tens of minutes). With the short polling wait, we usually just see "CommitStarted", but there are various other non-failure possibilities. Of course there are various failure possibilities too: Let's go with something like, if "Failed" in response_json.get("status").

gbrownmozilla: The whole response body is logged in the polling loop (in _get_submission_status), so I think there's no need to log the statusDetails (or status for that matter...but I find it a convenient/reassuring reminder).

jcristau: Thanks for the explanations, and sorry for the noise! :+1:

gbrownmozilla: I appreciate the extra feedback - thanks for taking the time!

PR #564, opened 2022-09-19
Bug 1791366 - Temporarily increase treescript timeout to 4 hours

JohanLorenzo:
Must be merged at the same time as https://github.com/mozilla-services/cloudops-infra/pull/4364


jcristau:
Things ended up working out without this for the big webrtc merge, so I'll close the PR now; this is still a problem waiting to happen though, probably until we fix k8s-autoscale to account for running tasks.

PR #567, opened 2022-09-19
Add copy-artifacts action

gabrielbusta:
Adds `copy-artifacts` action to `beetmoverscript` to move away from re-uploading artifacts from builds (useful in projects that don't use balrog and for non-app artifacts)

Like direct-push-to-bucket but it copies.

### DIFF
--- a/beetmoverscript/src/beetmoverscript/script.py
+++ b/beetmoverscript/src/beetmoverscript/script.py
@@ -167,6 +167,37 @@ async def push_to_partner(context):
     add_checksums_to_artifacts(context)
 
 
+# copy_artifacts {{{1
+async def copy_artifacts(context):
+    """A un-opinionated action to copy artifacts from one location to another. It WONT overwrite existing artifacts in the destination."""
+    context.artifacts_to_beetmove = {}
+    product = context.task["payload"]["product"]
+    context.bucket_name = get_bucket_name(context, product, "aws")
+    copy_from = context.task["payload"]["copy_from"]
+    copy_to = context.task["payload"]["copy_to"]
+    exclude_files = context.task["payload"].get("exclude_files")
+
+    creds = get_creds(context)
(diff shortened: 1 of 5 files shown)


beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -167,6 +167,37 @@ async def push_to_partner(context):
...
+    creds = get_creds(context)
+    s3_resource = boto3.resource("s3", aws_access_key_id=creds["id"], aws_secret_access_key=creds["key"])
+
+    from_keys = list_bucket_objects(context, s3_resource, copy_from)
+    to_keys = list_bucket_objects(context, s3_resource, copy_to)
+
+    if not from_keys:
+        raise ScriptWorkerTaskException("No artifacts to copy from {} so there is no reason to continue.".format(copy_from))
+
+    if to_keys:
```
escapewindow, line 189: <s>It may be nice to know if we have something different in `copy_from` than exists in `copy_to`, or if it's the same file.

We could maybe move [copy_key](https://github.com/mozilla-releng/scriptworker-scripts/blob/8fea664a6d732d66767e16c8ce5c984a70db7e23/beetmoverscript/src/beetmoverscript/script.py#L249-L261) into a top level helper function, and use that.</s>

N/m, you're using `copy_beets` below :)

escapewindow (approved):
I'm not a beetmover expert, but I suspect you'll chat with @hneiva about this since he needs to merge this into his GCS PR. Since this a) passes the eyeball test, b) will likely be limited to VPN for now, and c) we'll go back over this code for GCS support, r+.

gabrielbusta:
I have a pretty good idea what the code that supports GCS would look like ([something like this](https://github.com/mozilla-releng/scriptworker-scripts/commit/eccfb24662bd501e3b96a1892554329743ddef09)). I can add it once @hneiva lands the GCS support.

gabrielbusta:
Now that I tried it on the client side... it would have to take in a list of destinations... it should be more like the artifact map schema (rather than `push-to-releases`)

PR #569, opened 2022-09-27
treescript: don't append "esr" to all files on version bump

### LINKED ISSUE #349: treescript's version bump appended "esr" to version.txt which should only be append in version_display.txt
Today treescript automatically bumped the version of mozilla-esr78 to 78.10.1esr https://hg.mozilla.org/releases/mozilla-esr78/rev/ebc0ee4b3262cfc54cbeb809e9ceee39c3f08966#l1.4

version.txt should have `78.10.1` and version_display.txt `78.10.1esr`.

```
it's probably worth noting that this might have been broken for awhile
since we normally do the major version bump on Monday and ship on Tuesday (so the automated bump is a no-op)
what I'm saying is that we generally don't exercise this code because we do a 78.X bump on Monday so the 78.X-1.1 bump on Tuesday doesn't happen
so this could have been broken for awhile without us noticing because the merge day version bump Just Works
```

jcristau:
On ESR, the display version (in browser/config/version_display.txt) has an "esr" suffix, but the milestone (in config/milestone.txt) and application version (in browser/config/version.txt) don't.  Our post-release version-bump task has a "next_version" with the "esr" suffix, so we were adding it to all version files, and breaking expectations.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1786807

### DIFF
--- a/treescript/src/treescript/versionmanip.py
+++ b/treescript/src/treescript/versionmanip.py
@@ -162,6 +162,9 @@ async def do_bump_version(config, repo_path, files, next_version, source_repo):
             )
         ):
             next_version = VersionClass.parse("{}esr".format(next_version))
+        if not is_esr and next_version.is_esr:
+            # if the file had no esr prefix then don't add it
+            next_version = VersionClass.parse(str(next_version)[:-3])
 
         if next_version < curr_version:
             log.warning("Version bumping skipped due to conflicting values: " "(next version {} is < current version {})".format(next_version, curr_version))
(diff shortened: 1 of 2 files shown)


escapewindow (approved):
Hm, this works. I suggested a different task-definition-driven approach in https://github.com/mozilla-releng/scriptworker-scripts/issues/349, but this could be simpler.

jcristau:
@JohanLorenzo sorry for re-requesting review on this, but aki's no longer counts because I didn't merge this last year :(

JohanLorenzo (approved):
No worries, looks good to me!

PR #575, opened 2022-10-14
[beetmover] Add GCS support - fix platform

hneiva:
Bug fix is on the last commit


beetmoverscript/src/beetmoverscript/task.py
```diff
@@ -211,6 +211,8 @@ def get_release_props(task, platform_mapping=STAGE_PLATFORM_MAP):
     if not payload_properties:
         raise ScriptWorkerTaskException("could not determine release props file from task payload")
 
+    payload_properties = payload_properties.copy()
+
```
escapewindow: I slightly prefer

```
from copy import deepcopy

payload_properties = deepcopy(task).get("payload", {}).get("releaseProperties", None)
```

just in case payload properties ever contains lists or dicts.

```
#!/usr/bin/env python

from copy import deepcopy

a = {
"one": [1, 2, 3],
"two": ["a", "b", "c"],
}

b = deepcopy(a)["two"]
print(b)

print(b == a["two"])
print(b is a["two"])
```

results in 

```
['a', 'b', 'c']
True
False
```

escapewindow (approved):

PR #579, opened 2022-10-27
[pushflatpakscript] Bug 1744410 - Update Firefox screenshots for flathub

gbrownmozilla:
See discussion in bug 1744410. These screenshots were approved by Romain Testard.


ahal (approved):
Looks like the new ones aren't as tall, but I guess that's fine.

gbrownmozilla:
Right. It was hard to get the dimensions exactly the same and I don't think that's necessary.

PR #581, opened 2022-11-03
quick update to pushflatpakscript README

gbrownmozilla:
Noting a few minor things I recently learned about this script.


escapewindow (approved):
I wonder if the images should live in-tree at some point.
Thanks for the doc update!

PR #584, opened 2022-11-16
Bug 1780787 - signingscript: remove digicerthack

jcristau:
DigiCert's timestamp server now includes an intermediate in its response that serves our purposes (chains to a root that's included in Windows 7), so we don't need to mess with the certificates ourselves anymore.


signingscript/tox.ini
```diff
@@ -19,7 +19,6 @@ setenv =
     PYTHONPATH = {toxinidir}/tests
 deps=
     py39: -rrequirements/test.txt
-    py38: -rrequirements/test.py38.txt
```
JohanLorenzo, line 22: Nit: this line seems still needed to make `tox-signingscript-python38` happy.

```
[task 2022-11-16T14:23:58.249Z] py38 run-test: commands[0] | py.test --cov-config=tox.ini --cov-append --cov=/builds/worker/checkouts/vcs/signingscript/src/signingscript --cov-report term-missing tests
[task 2022-11-16T14:23:58.250Z] ERROR: InvocationError for command could not find executable py.test
```

[1] https://firefox-ci-tc.services.mozilla.com/tasks/H7A8tE0sS_Cx5WJCr_6lUA/runs/0/logs/public/logs/live.log#L79

JohanLorenzo (approved):
This looks good to me based on the fact that we're just reverting commits and based on the thorough explanation in [bug 1780787 comment 1](https://bugzilla.mozilla.org/show_bug.cgi?id=1780787#c1). 

Do we have a way to test this on try and nightly first? If we could make this change ride the trains, that would give us some leeway to ensure the timing is right and we don't break anything. That said, I'm not going to block on this 🙂

signingscript/tox.ini
jcristau: Right, thanks, looks like I didn't get the rebase quite right.  Fixed now.

jcristau:
> This looks good to me based on the fact that we're just reverting commits and based on the thorough explanation in [bug 1780787 comment 1](https://bugzilla.mozilla.org/show_bug.cgi?id=1780787#c1).
> 
> Do we have a way to test this on try and nightly first? If we could make this change ride the trains, that would give us some leeway to ensure the timing is right and we don't break anything. That said, I'm not going to block on this slightly_smiling_face

There's a few things we could do.  Off the top of my head:
- turn off the hack for adhoc, get a binary signed that way, and test it, before we remove the code
- add a "turn off the hack for me please" param in the task payload, and land a patch in-tree to set it, so it can ride the trains; once we're happy we can remove the code and ignore the flag
- turn off the hack for nightly and dep signing, and test on nightly, before we remove the code

It all depends on the amount of effort we want to spend vs the risk.  I'm reasonably comfortable landing this directly after the checks I've done, but I could be convinced to go for a more cautious approach.

JohanLorenzo:
Thanks for thinking these options through, @jcristau! To me, the third one is the most pragmatic. We can test this on the Nightly population without landing many patches in various places.

jcristau:
> Thanks for thinking these options through, @jcristau! To me, the third one is the most pragmatic. We can test this on the Nightly population without landing many patches in various places.

Cool.  Doing this in #588

signingscript/src/signingscript/sign.py
```diff
@@ -1436,29 +1435,21 @@ async def signer(digest, digest_algo):
...
-    winsign_kwargs = {
-        "cafile": cafile,
-        "timestampfile": timestampfile,
-        "url": url,
-        "comment": authenticode_comment,
-        "crosscert": crosscert,
-        "timestamp_style": timestamp_style,
-        "timestamp_url": timestamp_url,
-    }
-    log.info(f"running winsign.sign.sign_file with kwargs {winsign_kwargs}...")
```
escapewindow: It may be useful to have this logging info in the future, no? If not, then it's fine to remove.

jcristau: Added back.

PR #586, opened 2022-11-17
test: initialize `context.task["scopes"]` in the context test fixture

jcristau:
A bunch of tests need this, so might as well do this once and for all.


JohanLorenzo (approved):
LGTM!

PR #588, opened 2022-11-17
signing: disable digicerthack for nightly and dep (bug 1780787 - part 1)

jcristau:
Depends on #586


JohanLorenzo (approved):
:shipit:! Let's see how this behaves with nightly users

PR #610, opened 2022-12-16
Run pin.sh to bump requirements

gbrownmozilla:


JohanLorenzo (approved):

ahal:
I was going to ask if we could re-generate this again because I need scriptworker 46.0.0 for the CoT removal.. but maybe it would be better to do that in a separate commit in case we need to back it out. So look forward to another dep upgrade soon :)

ahal:
Oh nevermind, this *does* include scriptworker 46.0.0, nice :)

PR #615, opened 2023-01-11
[RELENG-1011] - Add AuthenticodeEV signing vars

ahal:


ahal:
I'm not sure if any different configuration is required for an EV cert rather than an OV one, or whether it should be using sha256 or not (probably should?).

Suffice to say there are some unknowns for me here, so I think I'm just going to have to land something, try it out in adhoc-signing, see what fails and go from there.

ahal:
I believe we need to wait for https://github.com/mozilla-services/cloudops-infra/pull/4620 to land before this can merge.

signingscript/src/signingscript/sign.py
```diff
@@ -1405,10 +1405,10 @@ async def signer(digest, digest_algo):
 
     infile = orig_path
     outfile = orig_path + "-new"
-    if "authenticode_sha2" in fmt:
-        digest_algo = "sha256"
-    else:
+    if fmt in ["autograph_authenticode", "autograph_authenticode_stub"]:
```
JohanLorenzo, line 1408: Looks like we don't use these formats anymore on central[1]:

image

Same results with ESR 102 and 91. ESR 78 was still referencing them[2], though:

image

I wonder if it's worth retiring these signing format in a followup PR.

[1] https://searchfox.org/mozilla-central/search?path=&q=autograph_authenticode
[2] https://searchfox.org/mozilla-esr78/search?path=&q=autograph_authenticode

signingscript/docker.d/passwords.yml
```diff
@@ -413,6 +413,11 @@ in:
              {"$eval": "AUTOGRAPH_AUTHENTICODE_SHA2_PASSWORD"},
              ["autograph_authenticode_sha2", "autograph_authenticode_sha2_stub"]
              ]
+          - ["https://autograph-external.prod.autograph.services.mozaws.net",
+             {"$eval": "AUTOGRAPH_AUTHENTICODE_EV_USERNAME"},
+             {"$eval": "AUTOGRAPH_AUTHENTICODE_EV_PASSWORD"},
+             ["autograph_authenticode_file"]
```
JohanLorenzo: Nit (no need to fix this in this current PR): I wonder if `autograph_authenticode_file` can confuse future readers knowing that `autograph_authenticode_sha2` and `autograph_authenticode_sha2_stub` also exist. That said, I know this change requires coordination so no need to address it right away.

JohanLorenzo (approved):
Looks good to me! 👍

signingscript/src/signingscript/task.py
```diff
@@ -53,6 +54,7 @@
         "autograph_authenticode_sha2": sign_authenticode_zip,
         "autograph_authenticode_sha2_stub": sign_authenticode_zip,
         "autograph_authenticode_sha2_rfc3161_stub": sign_authenticode_zip,
+        "autograph_authenticode_file": sign_authenticode_file,
```
jcristau: I doubt we can re-use sign_authenticode_file as-is.  It is currently set to use `context.config["authenticode_ca"]`, which is not the same intermediate that the EV cert chains to, and one of the `authenticode_cert*` config options as signing cert, which also point at the non-EV cert.
I'm seeing 2 options from a quick look:
- a new sign_authenticode_ev function that passes in the right EE and intermediate certs to winsign directly
- a new "keyid" for sign_authenticode_file, plus making it possible to pass in a different intermediate cert in e.g. `authenticode_ca_{keyid}` instead of always `authenticode_ca`
We'll also want to add both certs to signingscript/src/signingscript/data.

jcristau (changes requested):
Requesting changes because I *think* this isn't quite right.

bhearsum:
> I'm not sure if any different configuration is required for an EV cert rather than an OV one, or whether it should be using sha256 or not (probably should?).

I'm pretty sure we should use a sha256 digest (at th every least we should try with it). As far as I know, the only reason we stuck with sha1 digest for so long is because old versions of Windows 7 didn't support sha256 digests.

signingscript/docker.d/passwords.yml
ahal: Maybe it's worth creating a single entrypoint that can handle both single files as well as zip files...

signingscript/src/signingscript/task.py
ahal: Thanks, appreciate the pointer! I was pretty sure this was going to fail, but wasn't sure what I needed to do. This will help save me a lot of time investigating :)

signingscript/src/signingscript/sign.py
ahal: +1 to simplification! This stuff is already too complex for me. Though I'm definitely not going to tackle it in this PR.

signingscript/docker.d/passwords.yml
```diff
@@ -413,6 +413,12 @@ in:
              {"$eval": "AUTOGRAPH_AUTHENTICODE_SHA2_PASSWORD"},
              ["autograph_authenticode_sha2", "autograph_authenticode_sha2_stub"]
              ]
+          - ["https://autograph-external.prod.autograph.services.mozaws.net",
+             {"$eval": "AUTOGRAPH_AUTHENTICODE_EV_USERNAME"},
+             {"$eval": "AUTOGRAPH_AUTHENTICODE_EV_PASSWORD"},
+             ["autograph_authenticode"],
+             "ev"
```
ahal, line 420: I see the "202005" keyid is used above, but I'm a bit confused about how this works..

From what I can tell, these keyids should end up getting parsed into [this property](https://github.com/mozilla-releng/scriptworker-scripts/blob/ffe4fd9391d8857d17b4840d32acd9d98aa0d994/signingscript/src/signingscript/utils.py#L25). But from the looks of it, that property is only ever accessed in [this one place](https://github.com/mozilla-releng/scriptworker-scripts/blob/ffe4fd9391d8857d17b4840d32acd9d98aa0d994/signingscript/src/signingscript/sign.py#L1020).

But `sign_authenticode_file` doesn't ever call into that function and instead extracts keyids out of the format [over here](https://github.com/mozilla-releng/scriptworker-scripts/blob/ffe4fd9391d8857d17b4840d32acd9d98aa0d994/signingscript/src/signingscript/sign.py#L1397). And I can't seem to find anywhere that is joining a `keyid` to the format!

So how do authenticode keyids work??

ahal: I traced the signing format backwards, and afaict, it looks like it might just need to come from the [task payload](https://github.com/mozilla-releng/scriptworker-scripts/blob/ffe4fd9391d8857d17b4840d32acd9d98aa0d994/signingscript/src/signingscript/task.py#L224)? However, if that's the case, what's the point of putting the signing format in `passwords.yml` (like we do with the "202005" one)? Maybe that's actually a no-op and when :catlee landed #218 he didn't actually switch anything over... Fwiw there are no references to "202005" in Gecko.

ahal: Sigh nevermind. My initial statement was faulty and `sign_authenticode_file` does indeed ultimately call into `sign_with_autograph`.

On the bright side, I've learned a lot about how this works. And on the even brighter side, the PR looks like it might actually work.

jcristau (approved):
This looks good to me, thanks!.  Not a guarantee it'll work but worth giving it a try.

PR #618, opened 2023-01-17
[RELENG-1011] - Add Authenticode EV signing vars and certificates

ahal:
Take two. The previous patch failed because of missing json-e context values when rendering `worker.yml`. I believe it happened because I was only exporting `AUTHENTICODE_{CERT|CA}_PATH_EV` in the prod environment maybe? Though in that case, why did level 3 pools get busted? I added the values for non-prod anyway set to empty strings (as we don't have a dep EV cert to use).

I'm a bit unsure how to test this without landing again, though I'll definitely pick a more opportune time.


ahal:
I'm also planning to see if I can add a test that would have caught the failure.

jcristau:
Would it be worth creating a fake "ev" cert/ca for the non-prod case?

ahal:
> Would it be worth creating a fake "ev" cert/ca for the non-prod case?

How would that work? Like just a ca file with some junk in it? Or do you mean purchase an actual EV cert for dep usage? If the latter I think we'd want to avoid it because these things are pertty expensive.

This iteration just stores an empty string in the env vars for non-prod. I think that should more or less have the same effect as using a fake cert.

jcristau:
I meant the former.

bhearsum (approved):

signingscript/src/signingscript/sign.py
```diff
@@ -137,11 +137,11 @@ def get_autograph_config(autograph_configs, cert_type, signing_formats, raise_on
 
     """
     for a in autograph_configs.get(cert_type, []):
-        if a and (set(a.formats) & set(signing_formats)):
+        if a and (set(a.formats) & set(signing_formats)) and a.key_id == keyid:
```
jcristau: I think this breaks e.g. xpi dep signing, where the keyid is in the autograph config but not in the request.

Might need something like `(keyid is None or a.key_id == keyid)` if we don't want to reshuffle the config too much?

jcristau (changes requested):
Sorry for the delay, got distracted last week :/

signingscript/src/signingscript/sign.py
ahal: Good catch! Though why wouldn't it be in the request? Just because it works?

I wonder if it would be worth forcing requests to be explicit about keyids in the future. Definitely not a problem to solve in this PR though.

bhearsum: Historically, we've had "default" keyids for each format IIRC. I think we may have only support 1 per format for awhile. I'm totally onboard with being explicit here in the future. (Aki might have some historical perspective to add here as well.)

jcristau: Taking a step back, trying to sum up how this works, as far as I can tell...

The tasks can specify two things:
- scopes (typically dep/nightly/release)
- a set of "formats" for each file to sign
Those two things then map (via config) to an autograph url, credentials, format and optional key id.  The format can optionally include a key id that currently overrides the one set in the config.

The key id is sent to autograph, and it lets a single username (hawk_id?) use different keys, so AIUI that's what it's for in the xpi case.  By default if we don't set the key id, the username is the key id (I think?).

Here if I understand correctly we currently have a single key id for the ev user, so we don't need to pass the key id to autograph (although we probably can if we want to).  One reason to use a keyid in signingscript is to let us reuse `sign_authenticode_file` while using a different cert and CA than it currently does.  That looks like it requires setting the key id in the task payload though, so that's maybe not great when we need to rotate keys.  I'll ponder some more and think of possible alternatives...

ahal: > That looks like it requires setting the key id in the task payload though, so that's maybe not great when we need to rotate keys.

Wouldn't we just rotate the key in-place in `signingscript` and leave the keyid the same in the task?

Edit: Ah, I think I see what you mean with the current set up. We could possibly invent some kind of alias system to allow for this, but it's getting complex.

> The key id is sent to autograph, and it lets a single username (hawk_id?) use different keys, so AIUI that's what it's for in the xpi case. By default if we don't set the key id, the username is the key id (I think?).

Instead of using an `ev` keyid, should I instead create a new signing format (that still uses `sign_authenticode_file`), but then have some hardcoded logic that says:
```
if signing_format == "x":
    use this key
else:
    use that key
```
?

I thought the keyid approach was cleaner, but maybe it's using it for something that wasn't intended?

jcristau: > Instead of using an `ev` keyid, should I instead create a new signing format (that still uses `sign_authenticode_file`), but then have some hardcoded logic that says:
> 
> ```
> if signing_format == "x":
>     use this key
> else:
>     use that key
> ```
> 
> ?

That seems like the easiest way forward at this point.

> I thought the keyid approach was cleaner, but maybe it's using it for something that wasn't intended?

Agreed, it seemed promising but that doesn't seem to quite do what we want here :/

ahal:
@jcristau Ok, let's give this another shot. I *think* we still want the change in the first commit? Though I could pull that out to a separate PR to keep things clean if you like

signingscript/docker.d/passwords.yml
```diff
@@ -413,6 +413,12 @@ in:
              {"$eval": "AUTOGRAPH_AUTHENTICODE_SHA2_PASSWORD"},
              ["autograph_authenticode_sha2", "autograph_authenticode_sha2_stub"]
              ]
+          - ["https://autograph-external.prod.autograph.services.mozaws.net",
+             {"$eval": "AUTOGRAPH_AUTHENTICODE_EV_USERNAME"},
+             {"$eval": "AUTOGRAPH_AUTHENTICODE_EV_PASSWORD"},
+             ["autograph_authenticode_sha2"],
+             "ev"
```
jcristau, line 420: I think you need to remove the keyid here and change the format to `autograph_authenticode_ev`, plus add to `FORMAT_TO_SIGNING_FUNCTION`

jcristau (approved):
Let's split out the first commit, since we don't need it here.

r=me with the comment about the new format addressed.

Thank you!

PR #619, opened 2023-01-18
Add better support for `browser_specific_settings` (with `applications` as fallback)

willdurand:
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1810985


willdurand:
(I am looking into the test failures between meetings..)

willdurand:
@jcristau green 😎

jcristau (approved):
Thanks Will!

jcristau:
A couple of notes:
- we may want to duplicate addonscript/tests/fake_target.langpack.xpi to use the modern key (or just update it)
- the beetmoverscript case looks like it's not covered?

willdurand:
I added coverage for `addonscript` and I am trying to add new tests for `beetmoverscript`

willdurand:
The new tests should pass now :)

PR #620, opened 2023-01-19
Add a test to execute `docker.d/init.sh` with various configurations

ahal:


ahal:
Phew, I believe the intermittents are fixed. This should be ready for review!

tests/test_init.py
```diff
@@ -0,0 +1,288 @@
...
+
+import pytest
+
+here = Path(__file__).parent
+
+
+# These are variables that would typically be set by cloudops-infra.
+# Be careful not to add anything here that would normally be set within
+# the `init.sh` or `init_worker.sh` scripts, or else the purpose of the
+# test will be defeated.
```
bhearsum, line 18: I'm not necessarily recommending this at all, and definitely not recommending it as part of this -- but we could, in theory, pull the k8s configs to populate this. (It would require a github token and might be totally overengineering this -- but I wanted to toss it out there at least.)

bhearsum (approved):
I'm having trouble digging up the exact failure we're trying to regression test here now. Assuming it was `init.sh` failing, this looks great! (I don't see why it wouldn't have been, just calling it out because I can't verify this myself.)

tests/test_init.py
ahal: Yeah, that's a good idea! I'll file an issue so we remember it

ahal:
So I had added a new context value to `worker.yml` but only exported the environment variable when `ENV == prod`. Basically if you delete one of the existing envs in signing script like `AUTHENTICODE_CA_PATH`, you should be able to reproduce. Though any bug that causes a mismatch between what's in `worker.yml` and the envs should trigger this test.

PR #631, opened 2023-01-25
Bug 1812302: fix version bumps on mobile

jcristau:
Don't look at next_version.is_esr unless we're dealing with gecko, since that attribute doesn't exist elsewhere.


treescript/src/treescript/versionmanip.py
```diff
@@ -147,6 +147,10 @@ async def do_bump_version(config, repo_path, files, next_version, source_repo):
             is_esr = curr_version.is_esr
         except AttributeError:  # Fenix does not expose the is_esr attribute
             is_esr = False
+        else:
```
JohanLorenzo, line 150: TIL `try/except/else` in Python!

JohanLorenzo (approved):
LGTM! Thanks for adding a test for mobile! 👍

PR #587, opened 2022-11-17
signingscript: update vendored mozbuild

jcristau:
~TODO: replace the revert with a test fix~

PR #632, opened 2023-01-25
signingscript: update vendored mozbuild, take 2

jcristau:
The previous attempt in #587 broke production because mozbuild now imports `packaging` and I didn't add that to the requirements file.  CI didn't notice because packaging is pulled in by test dependencies.


bhearsum:
> The previous attempt in #587 broke production because mozbuild now imports `packaging` and I didn't add that to the requirements file. CI didn't notice because **packaging is pulled in by test dependencies**.

:rage:

bhearsum (approved):

PR #640, opened 2023-02-03
Only make `k8s_image` tasks depend on `tox` tasks when pushing to production

ahal:
We presumably have this dependency in place to help ensure we aren't pushing busted images to Docker hub. However in the context of PRs, or pushing to `dev` branches, we don't really care as much and this dependency just slows the pipeline down.

Let's only make the image tasks block when they need to.

### DIFF
--- a/taskcluster/scriptworker_taskgraph/parameters.py
+++ b/taskcluster/scriptworker_taskgraph/parameters.py
@@ -17,6 +17,17 @@ PROJECT_SPECIFIC_PREFIXES = {
 
 PUSH_TAGS = ("dev", "production")
 
+
+def get_defaults(_):
+    return {
+        "docker_tag": None,
+        "push_docker_image": False,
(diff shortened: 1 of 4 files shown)


ahal:
This worked, but looks like we were relying on that dependency link to pull in the `tox` tasks. I'll need to make sure those are scheduled independently.

ahal:
Hm, though they have https://github.com/mozilla-releng/scriptworker-scripts/blob/master/taskcluster/ci/tox/kind.yml#L20 and show up when I run `taskgraph target -p taskcluster/test/params/pull-request.yml`.. Could they have been optimized? I don't see any optimizations though..

ahal:
Oh nvm, they're using `cached_tasks`. So I guess this is working as expected :)

bhearsum (approved):
Thanks for doing this!

PR #641, opened 2023-02-07
Convert EV code signing cert's line endings from CRLF to LF

jcristau:
winsign expects unix line endings here.


JohanLorenzo (approved):
Github UI shows no changes if I request to hide whitespace ones. Thus, there's no change in the certificate. LGTM!

PR #643, opened 2023-02-07
signingscript: pass the signing certificate as authenticode_ca_ev

jcristau:
This is confusing, because it's called "cafile" in winsign and passed as "-CAfile" to osslsigncode, so you'd expect it to want an actual CA cert, but apparently passing the end entity cert is what we do for the existing authenticode signing, and passing the intermediate CA instead seems to fail.


JohanLorenzo (approved):
Discussed with @jcristau. This fix ensures parity with what we used to have. I confess I don't understand why we don't use a CA anymore. That said, let's green up the signing task.

PR #644, opened 2023-02-07
signingscript: add full certificate chain to authenticode_prod_ev.crt

jcristau:
Mimic authenticode_prod_202005.crt by adding the CA certificates along with the signing cert, to hopefully let osslsigncode attach-signature work.


jcristau:
The patch here is essentially `cat signingscript/src/signingscript/data/authenticode_prod_ca_ev.crt /usr/share/ca-certificates/mozilla/DigiCert_Trusted_Root_G4.crt >> signingscript/src/signingscript/data/authenticode_prod_ev.crt`

ahal:
This looks good to me! Though I wonder why we both setting all those `AUTHENTICODE_CA_PATH` environment variables if we just point them at the same thing as `AUTHENTICODE_CERT_PATH` anyway. Should we clean those up in a follow-up?

ahal (approved):

jcristau:
Yeah, I'll try to spend a bit more time to understand exactly what osslsigncode/winsign expect and then clean things up so it's slightly less confusing next time.

PR #657, opened 2023-02-08
signingscript: shuffle authenticode certs around

jcristau:
* The authenticode_cert* variables point at a file that contains the
  signing certificate plus any intermediate CA certificates.  These
  certs are added to the signed file so the client can verify
  signatures.

* The authenticode_ca* variables point at a (set of) root
  certificate(s).  We pass that to osslsigncode (via winsign) so it can
  verify the chain of certificates included in the binary.  The roots
  do not need to be included in the signed file since they're part of
  the platform's trust store.

This also deletes a couple of obsolete files related to an expired signing certificate.


jcristau:
One thing I should note, I believe this will result in the root cert no longer being included in our binaries.  Which should be fine, because the platform has it, but I've been wrong before, so I'll do some checks...

ahal (approved):
I can't say I'm confident this is going to work, but I'll trust your judgement.

jcristau:
I'm going to close this.  Safer to do any reshuffling as part of the cert rotation, and this was done in #979 for the new cert and chain.

PR #662, opened 2023-02-10
chore: Standardize syntax, black and isort configs

hneiva:


bouncerscript/pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [tool.black]
-line-length = 180
+line-length = 160
```
ahal, line 2: I've always found this super jarring, coming from most projects that set it to 100 or 80 :p.. Maybe as a follow-up we can ask the team if they're ok with 100 or something a little more common.

ahal (approved):
Thanks! Longer term I think we should try and have a single set of all these configuration files across all scripts. But this is a good first step.

hneiva:
Added this in #666

PR #666, opened 2023-02-14
feat: Move syntax check to ruff

hneiva:
Includes changes from #662


hneiva:
Pass 1: [green] https://firefox-ci-tc.services.mozilla.com/tasks/Gg-8iNBrR4qqtVGq2RLbiQ/runs/0/logs/public/logs/live.log
Pass 2: [expected failure] https://firefox-ci-tc.services.mozilla.com/tasks/PZS-E1QgTR23bchP4wmGEw/runs/0/logs/live/public/logs/live.log
Pass 3: [back to green] https://firefox-ci-tc.services.mozilla.com/tasks/cRMBTq6_T-CYBt6L4XFVOQ/runs/0/logs/live/public/logs/live.log

gabrielbusta (approved):

gabrielbusta:
This PR might be cursed https://tenor.com/bQnr2.gif

taskcluster/ci/ruff/kind.yml
```diff
@@ -0,0 +1,38 @@
...
+# Copy of tox job, but doesn't use caching
+
+transforms:
+    - scriptworker_taskgraph.transforms.python_version:transforms
+    - scriptworker_taskgraph.transforms.tox:transforms
+    - taskgraph.transforms.job:transforms
+    - taskgraph.transforms.task:transforms
+
+jobs:
+    check:
```
ahal: I'd prefer to call the kind `check` and the task `ruff` (so it'll be `check-ruff` instead of `ruff-check`). That way if we want to add more of these kinds of tasks in the future, there will be a good place for them.

taskcluster/ci/ruff/kind.yml
```diff
@@ -0,0 +1,38 @@
...
+---
+loader: taskgraph.loader.transform:loader
+
+kind-dependencies:
+    - docker-image
+
+# Copy of tox job, but doesn't use caching
+
+transforms:
+    - scriptworker_taskgraph.transforms.python_version:transforms
```
ahal: Maybe we don't need these transforms? Would we ever want to run this on more than one Python version?

taskcluster/ci/check/kind.yml
```diff
@@ -0,0 +1,35 @@
...
+    - docker-image
+
+transforms:
+    - taskgraph.transforms.job:transforms
+    - taskgraph.transforms.task:transforms
+
+jobs:
+    ruff:
+        description: "ruff-py39"
+        run-on-tasks-for: ["action", "github-pull-request", "github-push"]
```
jcristau: `action` here seems wrong?

hneiva: Good catch
🤔 does that mean tox/unit tests should also not have "action" listed?

ahal: It's probably a no-op. The `run-on-tasks-for` is only used if the target tasks method uses it, and actions don't even use target tasks.

So I'd say it's cleaner to remove both of them.

ahal (approved):

PR #667, opened 2023-02-15
[ci] Upgrade to taskgraph 4.2.0

hneiva:


jcristau:
Missing an update to `taskcluster/requirements.*`?  Looks like this is still using 1.1.7 (and so not creating any tasks).

taskcluster/requirements.txt
```diff
@@ -79,9 +87,9 @@ slugid==2.0.0 \
     --hash=sha256:a950d98b72691178bdd4d6c52743c4a2aa039207cf7a97d71060a111ff9ba297 \
     --hash=sha256:aec8b0e01c4ad32e38e12d609eab3ec912fd129aaf6b2ded0199b56a5f8fd67c
     # via taskcluster-taskgraph
-taskcluster-taskgraph==1.1.7 \
-    --hash=sha256:648e2cd5a92ce3412d830d1a28fb41ba3415956f44fe6013a53d920fc313ebdf \
-    --hash=sha256:7025856ed543c8e4a42fae580d579feadb53a54ce91b6e6b994e678449be3aee
+taskcluster-taskgraph==4.1.1 \
```
jcristau: Let's also version the dependency in requirements.in, for documentation purposes if nothing else.  The incompatible changes mean we need at least 3.x.

taskcluster/scriptworker_taskgraph/transforms/python_version.py
```diff
@@ -42,37 +42,38 @@ def _resolve_replace_string(item, field, subs):
...
         "run.command",
         "worker.command",
         "worker.docker-image",
     ]
-    for job in jobs:
-        for python_version in job.pop("python-versions"):
-            task = deepcopy(job)
-            subs = {"name": job["name"], "python_version": python_version}
+    for task_raw in tasks:
+        task = deepcopy(task_raw)
```
jcristau: Moving the deepcopy outside the loop doesn't seem safe.  AFAICT `_resolve_replace_string` would end up making changes to `task` in the first iteration that persist for later ones.
E.g. https://firefox-ci-tc.services.mozilla.com/tasks/NW_4tQxDSLqsr28eOp1azg is supposed to be using python 3.9 but ends up with 3.8.

taskcluster/scriptworker_taskgraph/transforms/python_version.py
```diff
@@ -42,37 +42,38 @@ def _resolve_replace_string(item, field, subs):
...
 
 
 @transforms.add
-def update_name_with_python_version(config, jobs):
-    for job in jobs:
-        job["name"] = "{}-python{}".format(job["name"], job["attributes"]["python-version"])
-        yield job
+def update_name_with_python_version(config, tasks):
+    for task_raw in tasks:
+        task = deepcopy(task_raw)
```
jcristau: Why copy?

jcristau (changes requested):

hneiva:
I just realized I never merged this.
@ahal should we upgrade to 4.2.0 instead?

ahal:
Up to you, doesn't hurt I suppose.

ahal (approved):
Thanks, lgtm!

gabrielbusta (approved):

PR #673, opened 2023-02-22
addonscript: rely on strict_min_version from the manifest

### LINKED ISSUE #672: Fix get_stripped_version to handle the langpack versions of Firefox 112+ (bug 1796531)
In https://bugzilla.mozilla.org/show_bug.cgi?id=1796531 , the langpack version format changed from `111.0buildid20230221.190142` to something like `112.0.20230222.123456`.

But the current logic still expects the "actual version" to be separated from the numeric build ID by the non-numeric "buildid" string,

https://github.com/mozilla-releng/scriptworker-scripts/blob/f4a16602b04ea6412882c7849ad110c76413d6b7/addonscript/src/addonscript/xpi.py#L39
https://github.com/mozilla-releng/scriptworker-scripts/blob/f4a16602b04ea6412882c7849ad110c76413d6b7/addonscript/src/addonscript/xpi.py#L10-L21

Note: the new version number only includes major.minor, not major.minor.patch. For a long while, the langpack format consisted of "major.minor" only, and only briefly "major.minor.patch" since https://bugzilla.mozilla.org/show_bug.cgi?id=1732676 until In https://bugzilla.mozilla.org/show_bug.cgi?id=1796531. So if the updated logic only checks major.minor, we should be good.

jcristau:
Remove the get_stripped_version fallback, since we expect all langpacks to have strict_min_version set.

Fixes #672

### DIFF
--- a/addonscript/src/addonscript/xpi.py
+++ b/addonscript/src/addonscript/xpi.py
@@ -1,26 +1,7 @@
 """Resources that operate on an XPI file."""
 
 import json
-import re
 from zipfile import ZipFile
-
-from addonscript.exceptions import BadVersionError
-
-
-def get_stripped_version(version):
-    """Strip out buildid or other extraneous info from the version.
-
-    Args:
-        version (string): the full version string from manifest.json
-                          e.g. `81.0buildid20200914232702`
-    Returns:
-        string: semver version
-    """
-    m = re.match(r"""[0-9\.]+""", version)
-    if m:
-        return m.group(0)
-    else:
-        raise BadVersionError(f"Can't determine stripped version from `{version}`!")
 
 
 def get_langpack_info(path):
@@ -36,6 +17,6 @@ def get_langpack_info(path):
         "version": manifest_info["version"],
         "id": browser_specific_settings["gecko"]["id"],
         "unsigned": path,
-        "min_version": browser_specific_settings["gecko"].get("strict_min_version", get_stripped_version(manifest_info["version"])),
+        "min_version": browser_specific_settings["gecko"]["strict_min_version"],
     }
     return langpack_info
--- a/addonscript/src/addonscript/exceptions.py
+++ b/addonscript/src/addonscript/exceptions.py
@@ -33,12 +33,6 @@ class FatalSignatureError(ScriptWorkerTaskException):
     pass
 
 
-class BadVersionError(ScriptWorkerTaskException):
-    """Fatal error when XPI's version is invalid against AMO sanity checks."""
-
-    pass
-
-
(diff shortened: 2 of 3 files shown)


addonscript/src/addonscript/xpi.py
```diff
@@ -36,6 +17,6 @@ def get_langpack_info(path):
         "version": manifest_info["version"],
         "id": browser_specific_settings["gecko"]["id"],
         "unsigned": path,
-        "min_version": browser_specific_settings["gecko"].get("strict_min_version", get_stripped_version(manifest_info["version"])),
+        "min_version": browser_specific_settings["gecko"]["strict_min_version"],
```
Rob--W, line 20: strict_min_version is not a required key in the langpack manifest. Is there a need to add an assertion for the validity here?

The current langpack generation script does always insert the key: https://searchfox.org/mozilla-central/rev/611660bff9e6d52f1769bf216a7fbd12ece4d2d2/python/mozbuild/mozbuild/action/langpack_manifest.py#482, but an extra check may still be useful.

willdurand: Breaking things loudly if `langpack_manifest.py` suddenly drops the `strict_min_version` prop is what we want here, but that kind of changes would break other things anyway.

jcristau: I don't think asserting the presence of strict_min_version is necessary, it doesn't really matter whether we're raising a KeyError or something else there.
I don't know if we want to assert something about what the version looks like, before we send it to AMO as a new firefox version, e.g. "it looks like `{major}.0`".
@diox does AMO do any validation on what the application version string looks like (in https://addons-server.readthedocs.io/en/latest/topics/api/v4_frozen/applications.html#put--api-v4-applications-(string-application)-(string-version)- specifically)?

diox: Yes, for the applications versions endpoint AMO wants the version to match https://github.com/mozilla/addons-server/blob/9ba4d6c404a6ab513aba8799582e213b181206c2/src/olympia/versions/compare.py#L10-L21

diox: In addition, it generates 2 additional special versions from the one you pass using that regexp, for internal needs:
```python
release_version = '%d.%d' % (version_data['major'], version_data['minor1'] or 0)
star_version = '%d.*' % version_data['major']
```

jcristau: Thanks, that seems like it should be good enough.

gabrielbusta (approved):

PR #612, opened 2022-12-21
Sanitize input to tar extractall

gbrownmozilla:
Poached the flatpak changes from #593 to get this moving.

PR #674, opened 2023-02-22
Bug 1806899 - [flatpakscript] Detect link escapes in safe_extract

gbrownmozilla:
Copy of latest changes from tooltool.

Follow-up on #612

### DIFF
--- a/pushflatpakscript/src/pushflatpakscript/flathub.py
+++ b/pushflatpakscript/src/pushflatpakscript/flathub.py
@@ -117,7 +117,7 @@ def check_and_extract_tar_archive(context, tar_file_path):
                     raise Exception("Attempted path traversal in tar file: " + member.name)
                 if member.issym():
                     link_path = os.path.join(os.path.dirname(member_path), member.linkname)
-                    if not _is_within_directory(path, link_path):
+                    if not is_within_directory(path, link_path):
                         raise Exception("Attempted link path traversal in tar file: " + member.name)
                 if member.mode & (stat.S_ISUID | stat.S_ISGID):
                     raise Exception("Attempted setuid or setgid in tar file: " + member.name)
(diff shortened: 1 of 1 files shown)


pushflatpakscript/src/pushflatpakscript/flathub.py
```diff
@@ -105,16 +105,20 @@ def check_and_extract_tar_archive(context, tar_file_path):
...
+            real_target = os.path.realpath(target)
+            prefix = os.path.commonprefix([real_directory, real_target])
+            return prefix == real_directory
 
         def safe_extract(tar, path=".", members=None, *, numeric_owner=False):
             for member in tar.getmembers():
                 member_path = os.path.join(path, member.name)
                 if not is_within_directory(path, member_path):
                     raise Exception("Attempted path traversal in tar file: " + member.name)
+                if member.issym():
```
jcristau, line 118: We can't do this check up-front (the realpath might be different before extract vs mid extract), we have to move it to the members param of extractall somehow.

gbrownmozilla: Sorry, I meant to go back and check for that - sloppy of me.

pushflatpakscript/src/pushflatpakscript/flathub.py
```diff
@@ -104,20 +104,29 @@ def check_and_extract_tar_archive(context, tar_file_path):
...
+                        raise Exception("Attempted link path traversal in tar file: " + member.name)
                 if member.mode & (stat.S_ISUID | stat.S_ISGID):
                     raise Exception("Attempted setuid or setgid in tar file: " + member.name)
-            tar.extractall(path, members, numeric_owner=numeric_owner)
+            def files(tar, path):
+                for member in tar:
+                    validate_tar_member(member, path)
+                    yield member
+
+            tar.extractall(path, members=files(tar, path), numeric_owner=numeric_owner)
```
jcristau, line 130: We should either remove the `members` param from safe_extract, or use it.  (AFAICT the single caller doesn't pass it, so removing is maybe the easier/saner option.)

jcristau (changes requested):

pushflatpakscript/src/pushflatpakscript/flathub.py
gbrownmozilla: Thanks.

jcristau (approved):

PR #675, opened 2023-02-23
addonscript: switch to v5 addon submission API (RELENG-935)

jcristau:
Replace the v4 flow of:
- upload new version
- poll for validation/signing
- download signed xpi

With:
- upload new xpi
- poll for processing/validation
- create new version using that upload
- poll for signing/publishing
- download signed xpi

This makes it possible to submit entirely new language packs and have them listed on AMO, instead of having to manually go in and add the required metadata.

### DIFF
--- a/addonscript/src/addonscript/api.py
+++ b/addonscript/src/addonscript/api.py
@@ -4,23 +4,30 @@ from aiohttp.client_exceptions import ClientResponseError
 
 from addonscript.exceptions import AMOConflictError, AuthFailedError, AuthInsufficientPermissionsError, FatalSignatureError, SignatureError
 from addonscript.task import get_channel
-from addonscript.utils import amo_download, amo_get, amo_put, get_api_url
+from addonscript.utils import amo_download, amo_get, amo_post, amo_put, get_api_url
 
-# https://addons-server.readthedocs.io/en/latest/topics/api/signing.html#uploading-a-version
-UPLOAD_VERSION = "api/v4/addons/{id}/versions/{version}/"
+# https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#upload-create
+UPLOAD_CREATE = "api/v5/addons/upload/"
 
-# https://addons-server.readthedocs.io/en/latest/topics/api/signing.html#checking-the-status-of-your-upload
-UPLOAD_STATUS = "api/v4/addons/{id}/versions/{version}/"
-UPLOAD_STATUS_PK = "api/v4/addons/{id}/versions/{version}/uploads/{upload_pk}/"
+# https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#upload-detail
+UPLOAD_DETAIL = "api/v5/addons/upload/{uuid}/"
 
+# https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#versions-list
+VERSION_LIST = "api/v5/addons/addon/{id}/versions/?filter=all_with_unlisted"
 
-# https://addons-server.readthedocs.io/en/latest/topics/api/applications.html
+# https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#version-detail
+VERSION_DETAIL = "api/v5/addons/addon/{id}/versions/{version}/"
+
+# https://addons-server.readthedocs.io/en/latest/topics/api/addons.html#put-create-or-edit
+ADDON_CREATE_OR_EDIT = "api/v5/addons/addon/{id}/"
+
+# https://addons-server.readthedocs.io/en/latest/topics/api/v4_frozen/applications.html
 # XXX when we want to support multiple products, we'll need to unhardcode the
 #     `firefox` below
-ADD_VERSION = "api/v4/applications/firefox/{version}/"
+ADD_APP_VERSION = "api/v4/applications/firefox/{version}/"
 
 
-async def add_version(context, version):
+async def add_app_version(context, version):
     """Add a new version to AMO.
 
     Use the `min_version` here, rather than the string with the buildid etc.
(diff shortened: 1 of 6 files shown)


addonscript/src/addonscript/api.py
```diff
@@ -45,29 +52,65 @@ async def add_version(context, version):
...
+    langpack_id = locale_info["id"]
+    url = get_api_url(context, ADDON_CREATE_OR_EDIT, id=langpack_id)
+    data = {
+        "categories": {"firefox": ["language"]},
+        "name": locale_info["name"],
+        "summary": locale_info["description"],
+        "version": {
+            "license": "MPL-2.0",
+            "upload": upload_uuid,
+        },
```
jcristau: I'm hardcoding the application name and license here, because that's not explicit in the xpi.
I think this is fine for right now because there's a few other places we already assume firefox, and it seems unlikely we'd need to change the license for specific langpacks (and if it happens we can reconsider then and add that as part of the task payload or something).

gabrielbusta: It might be worthwhile to not assume it's Firefox. Now that we can use the newer API, maybe we can automate upload for the web-extensions in the add-ons pipeline? IIRC there's a few add-ons in mozilla-extensions that are uploaded to AMO manually. This is fine for now, add-on script is only used by langpacks in Gecko.

addonscript/src/addonscript/api.py
```diff
@@ -45,29 +55,68 @@ async def add_version(context, version):
...
+    """
+    locale_info = context.locales[locale]
+    langpack_id = locale_info["id"]
+    url = get_api_url(context, ADDON_CREATE_OR_EDIT, id=langpack_id)
+    data = {
+        "categories": {"firefox": ["general"]},
+        "name": {"en-US": locale_info["name"]},
+        "summary": {"en-US": locale_info["description"]},
+        "version": {
+            "license": "MPL-2.0",
```
jcristau, line 103: There's a few instances currently of langpacks with different licenses set on AMO, but checking with @flodolo, those are "old" locales with incorrect metadata rather than intentional deviations.

addonscript/src/addonscript/api.py
```diff
@@ -45,29 +55,68 @@ async def add_version(context, version):
...
+        AMOConflictError: If the version already exists for this addon
+        FatalSignatureError: If AMO returned a permanent error
+    """
+    locale_info = context.locales[locale]
+    langpack_id = locale_info["id"]
+    url = get_api_url(context, ADDON_CREATE_OR_EDIT, id=langpack_id)
+    data = {
+        "categories": {"firefox": ["general"]},
+        "name": {"en-US": locale_info["name"]},
+        "summary": {"en-US": locale_info["description"]},
```
jcristau, line 101: This will reset the name/summary for the addon at every upload, and change them from their current values for existing langpacks, the first time this code runs.  Overall this seems like a good thing because it gets us some consistency, and if we ever want to update the template for these fields we can do it in one place (https://searchfox.org/mozilla-central/rev/3002762e41363de8ee9ca80196d55e79651bcb6b/python/mozbuild/mozbuild/action/langpack_manifest.py#161-214)

gabrielbusta: Yeah that seems like a good thing, along with making them all "MPL-2.0" when it first runs

jcristau:
I've tested this on try against addons.allizom.org, with both "listed" and "unlisted" channel.

https://firefox-ci-tc.services.mozilla.com/tasks/AVDnxWzdQMmauyrWcJzZ3g/runs/1 created the langpack-ca-valencia@firefox.mozilla.org addon, with a listed version.

I've verified that a force-rerun of a successful task is handled correctly and can download the existing xpi.

One issue that we hit during testing is that some langpack ids were in a "denied" list and couldn't be created (https://github.com/mozilla/addons/issues/9112).

addonscript/src/addonscript/api.py
```diff
@@ -1,26 +1,36 @@
...
+
+# https://addons-server.readthedocs.io/en/latest/topics/api/v4_frozen/applications.html
 # XXX when we want to support multiple products, we'll need to unhardcode the
 #     `firefox` below
-ADD_VERSION = "api/v4/applications/firefox/{version}/"
+ADD_APP_VERSION = "api/v4/applications/firefox/{version}/"
 
 
-async def add_version(context, version):
+async def add_app_version(context, version):
```
hneiva, line 33: At first glance, **_app_** seems misleading, I feel like it should be **_addon_** instead?

addonscript/src/addonscript/api.py
```diff
@@ -79,54 +128,61 @@ async def get_signed_addon_url(context, locale, pk):
...
+    file_detail = version_detail["file"]
+    status = file_detail["status"]
 
-    if not signed_data.get("download_url"):
-        raise SignatureError('Expected XPI "download_url" parameter. Got: {}'.format(signed_data))
+    if status == "disabled":
+        raise FatalSignatureError("XPI disabled on AMO")
+    if status != "public":
+        raise SignatureError("XPI not public")
+    return file_detail["url"], file_detail["size"], file_detail["hash"]
```
hneiva, line 143: I recommend using `namedtuple` for returning more a single value such as this, as it helps on readability and debugging.

hneiva (approved):
Thanks for updating this!
Just a couple non-blocking comments, totally optional

addonscript/src/addonscript/api.py
jcristau: No this is actually the app (firefox) version. :)

addonscript/src/addonscript/api.py
jcristau: Fair, I think I'll do that in a followup. Thanks for the review!

PR #681, opened 2023-03-02
[signingscript] Add apple notarization support

hneiva:
# ~~DO NOT MERGE~~

~~Secrets still need to be added to SOPS~~

It's ready to roll


signingscript/Dockerfile
```diff
@@ -10,10 +10,23 @@ RUN apt-get update \
  && apt-get clean \
  && ln -s /app/docker.d/healthcheck /bin/healthcheck
 
-COPY . /app
```
bhearsum, line 13: My gosh, does this mean we were copying the entire repo?! I had thought this was just everything in `signingscript`, but it looks like docker builds happen at the root of the repo?

signingscript/docker.d/build_msix_packaging.sh
```diff
@@ -1,10 +1,9 @@
 #!/bin/bash
 set +e
 
-git clone https://github.com/mozilla/msix-packaging msix-packaging
+git clone --depth=1 https://github.com/mozilla/msix-packaging --branch johnmcpms/signing --single-branch msix-packaging
```
bhearsum, line 4: Nice improvement!

signingscript/docker.d/apple_notarization_creds.yml
```diff
@@ -0,0 +1,21 @@
...
+    $match:
+      'COT_PRODUCT == "firefox"': 'project:releng:signing:'
+      'COT_PRODUCT == "thunderbird"': 'project:comm:thunderbird:releng:signing:'
+      'COT_PRODUCT == "mozillavpn"': 'project:mozillavpn:releng:signing:'
+      'COT_PRODUCT == "adhoc"': 'project:adhoc:releng:signing:'
+in:
+  $merge:
+    $match:
+      # TODO: Investigate one key per product
+      # TODO: Investigate one key per release type (nightly/release)
```
bhearsum, line 12: Honestly, I'm not sure there's any reason to do this. There's no differences in terms of who has access by product or release type (either directly to running pods, or by, eg: pushing to a repo). There's also no difference in what you can do with these creds. Even if we had separate ones for eg: Firefox Release and Thunderbird Nightly, you could notarize Firefox Release builds with Thunderbird Nightly notarization creds.

I'd check with hwine before going to any trouble like this.

signingscript/docker.d/apple_notarization_creds.yml
```diff
@@ -0,0 +1,21 @@
+$let:
```
bhearsum, line 1: I don't have any real objection, but out of curiosity, is there a reason this isn't in `passwords.yml`?

signingscript/src/signingscript/sign.py
```diff
@@ -1548,3 +1549,50 @@ async def sign_debian_pkg(context, path, fmt, *args, **kwargs):
...
+async def rcodesign_notarize(paths, notarization_workdir, creds_path):
+    NOTARIZATION_EXTENSIONS = ("app", "pkg")
+    for app in paths:
+        if app.split(".")[-1].lower() not in NOTARIZATION_EXTENSIONS:
+            raise SigningScriptError("Unable to notarize app format: %s" % app)
+        app_path = os.path.join(notarization_workdir, app)
+        command = [
+            "rcodesign",
+            "notary-submit",
+            "--staple",
```
bhearsum, line 1563: This causes rcodesign to block on notarization completing. This is a notable change from our current process, and will tie up available workers in whatever pool this runs in.

I think this is OK to do, _however_, this argues for creating a new pool that is exclusively responsible for notarization and stapling (to avoid other signing types getting blocked or delayed if there's a spike in notarization requests, or Apple is slow, etc, etc.). This will end up being a nice simplification and clean up of the current graph (where we have the confusing "sign1" -> "poll" -> "sign3" flow).

(I don't think we need to use the current model of separate workers for notarization and stapling, as long as we do use a separate worker type here, and make sure its `maxCapacity` is set fairly high.)

In an ideal world, each `signingscript` that is handling notarization would be able to claim multiple tasks (since most of its work is just sleeping and waiting on Apple) -- but that is well into the realm of future improvements, and absolutely not part of this initial work.

signingscript/src/signingscript/sign.py
```diff
@@ -1548,3 +1549,50 @@ async def sign_debian_pkg(context, path, fmt, *args, **kwargs):
...
+    
+    # Setup workdir
+    notarization_workdir = os.path.join(context.config["work_dir"], "apple_notarize")
+    _, compression = os.path.splitext(path)
+    all_file_names = await _extract_tarfile(context, path, compression, tmp_dir=notarization_workdir)
+    workdir_files = os.listdir(notarization_workdir)
+
+    # Notarize
+    with tempfile.NamedTemporaryFile() as temp_creds:
+        temp_creds.write(json.dumps(credential).encode("ascii"))
```
bhearsum: Is there a reason we can't simply deploy creds as this, rather than write them out with each task? We do this, for example, with Google credentials in pushapk workers: [creating files as part of init](https://github.com/mozilla-releng/scriptworker-scripts/blob/6f19b279ee606ebdb5c155f3a1afacb44be6d5e4/pushapkscript/docker.d/init_worker.sh#L69-L79), [setting the file path as part of the worker config](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/pushapkscript/docker.d/worker.yml#L190)

bhearsum (changes requested):
This looks quite good overall! I'd like to see a few things before this lands:
- Agreement on what pool(s) we're going to use for these new workers
- Reworking the creds to set-up the files at init (instead of with each task)
- Some testing through `dev` before this lands (perhaps through adhoc-signing as we discussed before?).

signingscript/docker.d/apple_notarization_creds.yml
hneiva: `passwords.yml` holds autograph passwords, adding anything else to it would entail refactoring some of the autograph secret loading

signingscript/src/signingscript/sign.py
hneiva: I'm not sure I love the idea of either:
a. Having the credentials in SOPS base64'd
b. Having to construct the json file in `init_worker.sh`

What if we moved the credentials creation to `async_main`? (`if !credential.exist, create it` sort of deal)

signingscript/docker.d/apple_notarization_creds.yml
bhearsum: Ah, OK. Sounds fine to keep it as you have it now.

signingscript/src/signingscript/sign.py
bhearsum: It will have to be in SOPS either way, no? (I don't think we have any other place to put secrets that these workers need?)

As far as init_worker vs. async_main - I don't feel terribly strongly either way. Although requiring an intermediary format to `async_main` makes local testing a bit more annoying? I'm not going to block on either approach - as long as it's not being done per-task I'm pretty happy.

signingscript/Dockerfile
hneiva: Heh unfortunately yes, and we don't have a `.dockerignore` file

hneiva:
Tested on dev pool: https://github.com/mozilla-releng/adhoc-signing/pull/152

signingscript/src/signingscript/sign.py
hneiva: Let's get some measurements on how quick this goes with adhoc, then we'll have a better idea of how bad it could be.

pyproject.toml
```diff
@@ -32,7 +32,7 @@ profile = "black"
 
 [tool.ruff]
 # Enable Pyflakes `E` and `F` codes by default.
-select = ["E", "F"]
+select = ["E", "F", "W"]
```
hneiva: This adds checks like requiring a blank line at end of files

signingscript/src/signingscript/script.py
```diff
@@ -75,6 +84,35 @@ def get_default_config(base_dir=None):
...
+        raise SigningScriptError("Credentials not found for scope: %s" % cert_type)
+    scope_credentials = apple_notarization_configs.get(cert_type)
+    if len(scope_credentials) != 1:
+        raise SigningScriptError("There should only be 1 scope credential, %s found." % len(scope_credentials))
+
+    context.apple_credentials_path = os.path.join(
+        os.path.dirname(context.config["apple_notarization_configs"]),
+        'apple_api_key.json',
+    )
+    if os.path.exists(context.apple_credentials_path):
```
bhearsum, line 107: This code path means that this function doesn't do what it claims ("write credentials"), and also doesn't throw an error. Why does it exist, and what can we do to improve this? Is it a remnant from when this was being called per task?

signingscript/src/signingscript/sign.py
```diff
@@ -1548,3 +1548,38 @@ async def sign_debian_pkg(context, path, fmt, *args, **kwargs):
...
+    notarization_workdir = os.path.join(context.config["work_dir"], "apple_notarize")
+    _, compression = os.path.splitext(path)
+    all_file_names = await _extract_tarfile(context, path, compression, tmp_dir=notarization_workdir)
+    workdir_files = os.listdir(notarization_workdir)
+
+    # Notarize
+    await rcodesign_notarize(workdir_files, notarization_workdir, context.apple_credentials_path)
+
+    # Compress files and return path to tarball
+    return await _create_tarfile(context, path, all_file_names, compression, notarization_workdir)
```
bhearsum, line 1586: Is there a particular reason we're sending these back as tarballs instead of pkg or apps (as we do with authenticode signing)?

pyproject.toml
jcristau: Let's move that to a separate PR please.

signingscript/src/signingscript/sign.py
hneiva: `.app`s are not compressed at all.
Since this is a "new" type of task, defaulting to a compressed output seems reasonable.
That being said, we might adjust it later if required 🙃

signingscript/src/signingscript/script.py
hneiva: Basically it will write the first time, and skip for any subsequent tasks.
The function still checks for scope issues, so there's a reason for it to exist.
Any specific suggestions on improving it?

bhearsum: In that case, it ought to have a different name, or move the scope checks out :).

signingscript/src/signingscript/sign.py
bhearsum: Ah right - I totally lost the plot on `.app` format :laughing: . This is fine.

bhearsum (approved):
I'm good with this with the `write_credentials` function renamed (or the scope check moved out).

PR #682, opened 2023-03-02
Bug 1806899 - [signingscript] update vendored mozbuild

gbrownmozilla:


gbrownmozilla:
I am surprised that so much has changed since #632 , but that seems to be the case.

I've copied python/mozbuild/mozbuild|mozpack and updated the README. Anything else to do/check?

jcristau:
Looks like you didn't pull in the setup.py update from https://hg.mozilla.org/mozilla-central/rev/b42c0644448c?
Other than that, LGTM.  I verified that `signingscript.sign` can still be imported when installing only the deps from `signingscript/requirements/base.txt`, to avoid a repeat of #630.

gbrownmozilla:
Thank you!

jcristau (approved):

PR #683, opened 2023-03-06
Bug 1806899 - Update vendored mozbuild (top level)

gbrownmozilla:


gbrownmozilla:
iscript tests test_omnija_same and test_omnija_sign fail with a faithful update of vendored/mozbuild. If I revert mozpack/mozjar.py and mozpack/copier.py to the old versions, the tests pass.
Local testing indicates https://hg.mozilla.org/mozilla-central/rev/3a6dd9a13ec9337f54f1a7f5448cb2d53a061ec8 is the cause of the test failures.
In particular, that rev sets creator_version/external_attr when mode is None, changing the archive contents.

jcristau:
I hit the same change when updating signingscript's copy, https://github.com/mozilla-releng/scriptworker-scripts/pull/587/commits/e889febcf16572093617f5115581a59e7f7bf0d1 updated the tests to match.

gbrownmozilla:
YES! Thank you!!

hneiva (approved):

PR #687, opened 2023-03-10
beetmoverscript: check destination path in push-to-maven

jcristau:
`check_maven_artifact_map` now verifies that the product name in the task payload matches the destination paths we're told to upload to.

This should prevent uploads to the wrong place because product names need to match entries in the product_buckets config, as a secondary protection to the permissions granted to cloud credentials.

### DIFF
--- a/beetmoverscript/src/beetmoverscript/constants.py
+++ b/beetmoverscript/src/beetmoverscript/constants.py
@@ -113,6 +113,14 @@ PRODUCT_TO_PATH = {
     "vpn": "pub/vpn/",
 }
 
+MAVEN_PRODUCT_TO_PATH = {
+    "appservices": "maven2/org/mozilla/appservices/",
+    "components": "maven2/org/mozilla/components/",
+    "nightly_components": "maven2/org/mozilla/components/",
+    "geckoview": "maven2/org/mozilla/geckoview/",
+    "telemetry": "maven2/org/mozilla/telemetry/",
+}
+
 PARTNER_REPACK_PUBLIC_PREFIX_TMPL = "pub/firefox/candidates/{version}-candidates/build{build_number}/"
 PARTNER_REPACK_PRIVATE_REGEXES = (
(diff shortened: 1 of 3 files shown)


beetmoverscript/src/beetmoverscript/constants.py
```diff
@@ -113,6 +113,14 @@
     "vpn": "pub/vpn/",
 }
 
+MAVEN_PRODUCT_TO_PATH = {
+    "appservices": "maven2/org/mozilla/appservices/",
+    "geckoview": "maven2/org/mozilla/geckoview",
```
JohanLorenzo: Nit: Missing trailing `/` which could hypothetically be harmful if one day we decided to publish a product maven library called something like `geckoview-foo` 

```suggestion
    "geckoview": "maven2/org/mozilla/geckoview/",
```

beetmoverscript/src/beetmoverscript/task.py
```diff
@@ -140,10 +140,13 @@ def get_maven_version(context):
 
 
 def check_maven_artifact_map(context, version):
-    """Check that versions in artifact map are consistent with a given version"""
+    """Check that paths in artifact map are consistent with a given product and version"""
+    product = utils.get_product_name(context.task, context.config).lower()
     for artifact_dict in context.task["payload"]["artifactMap"]:
         for dest_dict in artifact_dict["paths"].values():
             for dest in dest_dict["destinations"]:
+                if not dest.startswith(MAVEN_PRODUCT_TO_PATH[product]):
```
JohanLorenzo, line 148: 👍

JohanLorenzo (approved):
Looks good to me modulo a small nit! Thank you for making beetmover more secure!

beetmoverscript/src/beetmoverscript/constants.py
jcristau: Thanks for catching that; fixed.

gbrownmozilla (approved):
Nice.

PR #697, opened 2023-03-22
fix: signingscript notarization - fix credential handling

hneiva:


gabrielbusta (approved):

signingscript/src/signingscript/utils.py
```diff
@@ -92,11 +92,15 @@ def _load_scoped_configs(filename, cls, name):
     with open(filename) as f:
         raw_cfg = json.load(f)
 
-    cfg = {}
+    scope_configs = {}
+    # TODO: We should refactor how Autograph data is loaded to be key-pair instead of list
```
gabrielbusta, line 96: Yea

signingscript/src/signingscript/utils.py
```diff
@@ -92,11 +92,15 @@ def _load_scoped_configs(filename, cls, name):
...
     for scope, config in raw_cfg.items():
-        cfg[scope] = [cls(*s) for s in config]
+        if cls == Autograph:
+            scope_configs[scope] = [cls(*s) for s in config]
+        elif cls == AppleNotarization:
```
gbrownmozilla, line 100: What if it's something other than Autograph or AppleNotarization? (Could we make this an if/else?)

signingscript/src/signingscript/utils.py
gabrielbusta: If we refactored that you would not need the if else on the `cls` instantiation?

signingscript/src/signingscript/utils.py
hneiva: Good point, I'll raise an exception for now

PR #698, opened 2023-03-27
Bug 1824647 - treescript: update github's ssh host key

jcristau:
https://github.blog/2023-03-23-we-updated-our-rsa-ssh-host-key/


JohanLorenzo (approved):
LGTM! I manually checked the fingerprint against [1] too and we're good!

[1] https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/githubs-ssh-key-fingerprints

PR #699, opened 2023-03-28
fix: signingscript notarize - ignore dmg files

hneiva:
Adds `.background` file to `appletest.tar.gz` for tests


signingscript/src/signingscript/sign.py
```diff
@@ -1573,11 +1573,13 @@ async def apple_notarize(context, path, *args, **kwargs):
     """
     Notarizes given package(s) using rcodesign.
     """
+    # Files that are usually included in .dmg packages don't need notarizing
+    IGNORED_FILES = (".DS_Store", ".VolumeIcon.icns", ".background")
```
bhearsum: What do you think about doing this as an allowlist instead of a denylist? It feels like we'd need to update an allowlist of extensions much less often than a denylist of random filenames that might happen to be present.

bhearsum (commented):
I think the test tarball needs some more contents. Right now it only contains a pkg and `.background`. If we go with an allowlist (see below) it ought to contain any other file types that need notarization. If we stick with a denylist, we should make sure we test all files present in that list.

hneiva:
Sounds good.
We did have a "whitelist" on the `rcodesign_notarize` function. So I moved that to the controller function.
Also added a .app to the test tarball

signingscript/src/signingscript/sign.py
```diff
@@ -1567,20 +1564,29 @@ async def rcodesign_notarize(paths, notarization_workdir, creds_path):
...
+    NOTARIZATION_EXTENSIONS = ("app", "pkg")
     # Setup workdir
     notarization_workdir = os.path.join(context.config["work_dir"], "apple_notarize")
     _, compression = os.path.splitext(path)
     all_file_names = await _extract_tarfile(context, path, compression, tmp_dir=notarization_workdir)
     workdir_files = os.listdir(notarization_workdir)
 
+    # Filter supported file extensions
+    def is_supported(path):
+        return path.split(".")[-1].lower() in NOTARIZATION_EXTENSIONS
```
bhearsum: Somewhat of a nit, but we ought to use `os.path.splitext` for this instead, to avoid certain edge cases (like multiple dots in `path`). r=me with that adjusted - no need for re-review.

bhearsum (approved):

PR #702, opened 2023-04-05
Bug 1825582 - Force retries on GCS operations

gabrielbusta:
See [Bug 1825582](https://bugzilla.mozilla.org/show_bug.cgi?id=1825582) for more information.

### DIFF
--- a/beetmoverscript/src/beetmoverscript/gcloud.py
+++ b/beetmoverscript/src/beetmoverscript/gcloud.py
@@ -10,6 +10,7 @@ from google.api_core.exceptions import Forbidden
 from google.auth.exceptions import DefaultCredentialsError
 from google.cloud import artifactregistry_v1
 from google.cloud.storage import Bucket, Client
+from google.cloud.storage.retry import DEFAULT_RETRY
 from scriptworker.exceptions import ScriptWorkerTaskException
 
 from beetmoverscript.constants import CACHE_CONTROL_MAXAGE, RELEASE_EXCLUDE
@@ -115,8 +116,17 @@ async def upload_to_gcs(context, target_path, path):
     if blob.exists():
         log.warn("upload_to_gcs: Overriding file: %s", target_path)
     log.info("upload_to_gcs: %s -> Bucket: gs://%s/%s", path, bucket_name, target_path)
+    """
+    In certain cases, such as when handling *-latest directories, we need to overwrite existing file blobs.
+    Since we don't use `DELETE` requests in beetmover, the race condition mentioned in the GCS documentation [1] should not occur.
+    A race condition may arise if multiple tasks run simultaneously and attempt to upload to the same file key.
+    However, this issue exists independently of retries, so we proceed with retrying.
+    That's why we explicitly provide `retry=DEFAULT_RETRY` instead of utilizing `if_generation_match` [2], as recommended in the documentation.
 
-    return blob.upload_from_filename(path, content_type=mime_type)
+    [1] https://cloud.google.com/storage/docs/request-preconditions#multiple_request_retries
+    [2] https://cloud.google.com/storage/docs/xml-api/reference-headers#xgoogifgenerationmatch
+    """
+    return blob.upload_from_filename(path, content_type=mime_type, retry=DEFAULT_RETRY)
 
 
 async def import_from_gcs_to_artifact_registry(context):
@@ -225,4 +235,4 @@ def move_artifacts(client, bucket_name, blobs_to_copy, candidates_blobs, release
         else:
             log.info("Copying {} to {}".format(source, destination))
             source_blob = bucket.blob(source)
-            bucket.copy_blob(source_blob, bucket, destination)
+            bucket.copy_blob(source_blob, bucket, destination, retry=DEFAULT_RETRY)
--- a/beetmoverscript/tests/test_gcloud.py
+++ b/beetmoverscript/tests/test_gcloud.py
@@ -2,7 +2,9 @@ import os
 
 import pytest
 from google.api_core.exceptions import Forbidden
+from google.api_core.retry import Retry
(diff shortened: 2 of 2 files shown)


beetmoverscript/src/beetmoverscript/gcloud.py
```diff
@@ -116,7 +117,7 @@ async def upload_to_gcs(context, target_path, path):
         log.warn("upload_to_gcs: Overriding file: %s", target_path)
     log.info("upload_to_gcs: %s -> Bucket: gs://%s/%s", path, bucket_name, target_path)
 
-    return blob.upload_from_filename(path, content_type=mime_type)
+    return blob.upload_from_filename(path, content_type=mime_type, if_generation_match=0, retry=DEFAULT_RETRY)
```
gabrielbusta: `if_generation_match = 0` means Google Cloud Storage only performs the specified request if the object does not currently exist. This makes the operation idempotent and safe to retry.

`retry=DEFAULT_RETRY` ensures the operation will be retried.

gabrielbusta: The idempotency of operations is explained in the docs [here](https://cloud.google.com/storage/docs/retry-strategy#idempotency-operations). Here we can see that upload and copy requests are conditionally idempotent.

The behavior of the `if_generation_match` header is documented [here](https://cloud.google.com/storage/docs/xml-api/reference-headers#xgoogifgenerationmatch).

`retry=DEFAULT_RETRY` is not strictly necessary because the default behavior is to retry if the `if_generation_match` retry header is present. We can see that in the code [here](https://github.com/googleapis/python-storage/blob/7c8a178978d2022482afd301242ae79b2f9c737a/google/cloud/storage/retry.py#L141-L149), [here](https://github.com/googleapis/python-storage/blob/9ccdc5f2e8a9e28b2df47260d639b6af2708fe9a/google/cloud/storage/bucket.py#L1788), and [here](https://github.com/googleapis/python-storage/blob/9ccdc5f2e8a9e28b2df47260d639b6af2708fe9a/google/cloud/storage/blob.py#L2403).

That being said, I opted for passing the `DEFAULT_RETRY` policy to make it explicit that this request will be retried if we receive error codes from the GCS API. [Explicit is better than implicit](https://peps.python.org/pep-0020/) and `if_generation_match=0` doesn't scream "this request will be retried."

jcristau: Does this mean we'll now fail if the blob already exists?
Previously it looks like we would have warned but done it anyway.

gabrielbusta: Now that you mention it, I see the if statement 116 is similar to `if_generation_match=0`. We could also wrap the operations (e.g. `upload_to_gcs`) in a `redo.retry`

But, there's a [warning on the docs](https://cloud.google.com/storage/docs/request-preconditions#multiple_request_retries) about retries causing surprising behavior due to the distributed nature of GCS.

GCS responds with `HTTP 412 Precondition Failed ` if the operation's preconditions fail (which might happen on a retry) so we could catch that and print a warning.

jcristau: As far as I can tell there are cases where we do want to overwrite existing files, namely the *-latest directories, e.g. https://firefoxci.taskcluster-artifacts.net/EgbWxJAeQ-qv9FXBADaErQ/0/public/logs/live_backing.log shows warnings like:
`2023-04-07 11:51:44,543 - beetmoverscript.gcloud - WARNING - upload_to_gcs: Overriding file: pub/firefox/nightly/latest-mozilla-central/KEY`
That means we can't just set `if_generation_match=0` everywhere.

On the bright side, I don't think we ever issue a `DELETE` in beetmover, so the race condition described in the doc shouldn't happen.  There's a race condition if 2 tasks running concurrently try to upload to the same filename, but that exists regardless of retries.

Overall it seems to me that we should be able to use `retry=DEFAULT_RETRY` and not pass `if_generation_match`. That warrants a big fat comment in the code though.  What do you think?

jcristau: Another option might be to only allow overwriting files in the push-to-nightly action.  We could leave that to a follow up, after we've double checked there are no other cases of files getting overwritten.

gabrielBusta: Thanks for taking a closer look. What you are saying makes a lot of sense.

Revising...

gabrielbusta: Thanks for taking a closer look. What you are saying makes a lot of sense.

beetmoverscript/tests/test_gcloud.py
```diff
@@ -20,12 +22,14 @@ class FakeBlob:
...
-        def copy_blob(*args):
+        def copy_blob(*args, **kwargs):
             pass
 
-        def upload_from_filename(self, path, content_type):
+        def upload_from_filename(self, path, content_type, retry=DEFAULT_RETRY_IF_GENERATION_SPECIFIED, if_generation_match=None):
             assert path == self.PATH
             assert content_type == "application/zip"
+            assert isinstance(retry, (Retry, ConditionalRetryPolicy))
+            assert isinstance(if_generation_match, int) or if_generation_match is None
```
jcristau: `isinstance(if_generation_match, (int, type(None)))` maybe.

jcristau (approved):
lgtm

beetmoverscript/src/beetmoverscript/gcloud.py
bhearsum: For crosslinking, https://github.com/mozilla-releng/scriptworker-scripts/pull/1230 added support for rejecting overwrites to `upload_to_gcs` (but didn't go so far as to rejecting them for any Firefox uploads).

PR #703, opened 2023-04-12
Update release branch names in RELEASE_BRANCHES

gabrielbusta:


jcristau (commented):
IMO we should check with SRE whether there are currently any differences between archive.mozilla.org and download.cdn.mozilla.net, and/or whether we want to keep the option of splitting them out in the future and thus need to handle release branches specifically here, and if not then get rid of this entirely.

jcristau:
> IMO we should check with SRE whether there are currently any differences between archive.mozilla.org and download.cdn.mozilla.net, and/or whether we want to keep the option of splitting them out in the future and thus need to handle release branches specifically here, and if not then get rid of this entirely.

And actually, even if we do want to replace archive.m.o with download.cdn.m.n still, we might want to make it unconditional: before https://github.com/mozilla-releng/beetmoverscript/commit/d3d932ef3675134d288d7c654e2a9c8aa0a33873, only the "date" project branch was skipping it, and that's probably irrelevant nowadays.

gabrielbusta:
I checked with SRE

- It's the same service. No differences aside from hostname
- We should only publicize `archive.mozilla.org`
- `download.cdn.m.n` is a implementation detail that is subject to change

jcristau (changes requested):
Let's get rid of RELEASE_BRANCHES entirely then.

PR #705, opened 2023-04-12
[beetmoverscript] Remove "robocop" reference

gbrownmozilla:
Just tidying up: Robocop was removed in bug 1580832 (in 2019!).


gabrielbusta (approved):
Nice find.

PR #706, opened 2023-04-13
fix: Properly handle signingscript pkg notarization

hneiva:


signingscript/src/signingscript/sign.py
```diff
@@ -1565,31 +1565,50 @@ async def rcodesign_notarize(paths, notarization_workdir, creds_path):
...
 
     if not supported_files:
         raise SigningScriptError("No supported files found")
 
     # Notarize
     await rcodesign_notarize(supported_files, notarization_workdir, context.apple_credentials_path)
 
+    if extension == ".pkg":
+        # Copy pkg back - returns the destination path
+        return shutil.copy2(all_file_names[0], path)
```
bhearsum: This function is async, and this doesn't return a coroutine. I don't know if this will actually cause any issues -- but it's certainly unintuitive for the caller. We should make sure this returns a coroutine, even if it's already been resolved to this value.

bhearsum (changes requested):
It looks to me like this function now has two completely separate operating modes:
1) `path` is a pkg, where we have one file to notarize and copy somewhere
2) `path` is something else (I'm guessing this a directory containing a .app?), where we have a bunch of things to notarize and then repackage

This is fine...but I think a big if/else over most of the function would make it easier to read rather than multiple scattered blocks.

(For example, when I first this, it looked to me like it was possible for `supported_files` to contain a pkg _and_ something else, in which case the final `if extension` block would've skipped creating a tarball.)

signingscript/src/signingscript/sign.py
hneiva: 🤔 aiui `async` functions don't necessarily need to return coroutines.
In fact, looking thru all the other functions in this file, none of them do. They either return a flat value, or await another function, essentially returning the flat value.

bhearsum: Well, if we're doing it elsewhere...I'm not going to block on it here.

bhearsum (approved):

PR #707, opened 2023-04-14
feat: FFXP-1794 iscript mac hardened signing

hneiva:


iscript/src/iscript/hardened_sign.py
```diff
@@ -0,0 +1,182 @@
...
+)
+from iscript.util import get_sign_config
+from scriptworker_client.aio import download_file, raise_future_exceptions, retry_async
+from scriptworker_client.exceptions import DownloadError
+from scriptworker_client.utils import run_command
+
+log = logging.getLogger(__name__)
+
+
+async def download_signing_resources(hardened_sign_config, folder):
```
bhearsum, line 35: Did you consider adding the full entitlements to the payload (as opposed to downloading it at runtime?). The former has a couple of benefits I can see:
* No network dependence (more reliable)
* More reproducible (no chance of the remote URL changing over time, eg: if it refers to a branch or tag name

iscript/src/iscript/data/i_task_schema.json
```diff
@@ -99,6 +101,35 @@
             "entitlements-url": {
               "type": "string"
             },
+            "signing-config": {
```
bhearsum: Can we find another name for this, please? The entire schema is already arguably a "signing config" of sorts. If this is specifically only for hardened signing, perhaps something like `hardening-config`? (I'm open to other ideas - just wanted to throw something out there.)

iscript/src/iscript/constants.py
```diff
@@ -22,6 +22,9 @@
         "zipfile_cmd": "ditto",
         "create_pkg": False,
         "hardened_runtime_only_files": ["wg", "wireguard-go"],
+        "profiles": {
```
bhearsum: This doesn't appear to be used anywhere? (And if it is, it could use a better name. Perhaps `provisioning_profiles` if that's what they are.)

iscript/src/iscript/constants.py
```diff
@@ -22,6 +22,9 @@
...
+            "path": "local_profile_path"
+        }
```
bhearsum: Further up in this file I see we have `designated_requirements` entries for each product. I presume that under the new signing formats these are just going to be completely ignored, since they're coming in (directly or indirectly) via the payload instead? If that's the case, we ought to make that clear. A few suggestions (let's do at least one...but I'll leave it up to you how far to go):
* Rename to `non_hardened_designated_requirements`
* Note in the log with each hardened signing that the `designated_requirements` from the product config is being ignored
* Add new product configs that don't have this at all, eg: `mozillavpn-hardened`.

iscript/src/iscript/data/i_task_schema.json
```diff
@@ -99,6 +101,35 @@
...
+              "type": "array",
```
bhearsum, line 105: Can you also point to an example of what this might look like in practice? It would help to understand how the schema fits together.

iscript/src/iscript/hardened_sign.py
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env python
```
bhearsum, line 1: Hooray for segmenting this code off!

bhearsum (commented):
Overall this looks fine. I have some requests and suggestions to improve clarity and debugability in the future - but nothing overarching.

Has this been tested on try or anywhere yet that you can link to?

iscript/src/iscript/hardened_sign.py
hneiva: I think I'd still prefer having the exact file from the repo show up in the app entitlements without parsing and composing a new file. That way we "future proof" if the file ever changes to a different format (never doubt apple would do something like this lol)

iscript/src/iscript/constants.py
hneiva: I believe for now `designated_requirements` will still be hard-coded in constants due to its sensitivity.
The "profiles" object here was actually left in by mistake.. Oops :)

iscript/src/iscript/data/i_task_schema.json
hneiva: It will be a variation of this:
https://github.com/hafta/codesign-tree/blob/master/examples/01-firefox/codesign-map.json

hneiva:
> Has this been tested on try or anywhere yet that you can link to?

Tested manually on a quarantined mac. I plan on further testing this on adhoc before merging. PR coming up for it [hopefully] soon

hneiva:
Tested on this PR in dep: https://github.com/mozilla-releng/adhoc-signing/pull/166

Will test again as as prod signing once this PR is merged.

iscript/src/iscript/hardened_sign.py
bhearsum: Just to be clear: I'm not suggesting that we need to reconstitute the file, but rather we put its entire contents into one field of the payload (which can then be written out if needed, or simply passed along to whatever function requires it as a string).

I'll leave this your call - but if we start having issues with intermittents with this we should consider it.

bhearsum:
Holding off on the r+ until I see this tested in adhoc - but I think this is ready to go if that goes well.

jcristau:
Where can I find context on this work?  The commit messages and PR description are minimal...

hneiva:
@jcristau changed the title to include the main ticket
Ticket: https://mozilla-hub.atlassian.net/browse/FFXP-1794?search_id=216c4c7d-7f28-49c5-8eb6-1009ce6e0eea

Doc: https://docs.google.com/document/d/1eLg_jJvObPHirj1hykrNO91B7UOlxO8TTQ_knki4Irs/edit# 
(see Milestone 2)
Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1593072

jcristau:
@hneiva that's helpful, thanks

iscript/src/iscript/data/i_task_schema.json
```diff
@@ -99,6 +101,35 @@
...
+                  "runtime": {
+                    "type": "boolean"
+                  },
+                  "force": {
+                    "type": "boolean"
+                  },
+                  "entitlements": {
+                    "type": "string"
+                  },
+                  "globs": {
```
bhearsum, line 123: Now that I've seen what actually goes in here - can we find a better name for this? `paths` perhaps? Most notably, directory entries like `/Contents/MacOS/plugin-container.app` are not globs.

iscript/src/iscript/hardened_sign.py
```diff
@@ -0,0 +1,183 @@
...
+        )
+    await raise_future_exceptions(futures)
+    # Return map of url to file location
+    return url_map
+
+
+def check_globs(app_path, globs):
+    for path_glob in globs:
+        if not path_glob.startswith("/"):
+            raise IScriptError('ERROR: file pattern "{path_glob}" must start with "/"')
```
bhearsum: If it must start with it, should we be more graceful about this and just add it if it's missing?

iscript/src/iscript/data/i_task_schema.json
bhearsum: Is that link the authorative place going forward, or is this file authorative? If the former, please add a link.

bhearsum (commented):
What happened to all of the `contants.py` changes in the previous version of this? I know you said part of it was unnecessary, but https://github.com/mozilla-releng/scriptworker-scripts/pull/707#discussion_r1198565617 suggests that part of it was still needed.

iscript/src/iscript/hardened_sign.py
hneiva: Updated it so it can handle both cases

bhearsum (approved):

PR #714, opened 2023-04-21
feat: signinscript - recover from notarization connection error

hneiva:


signingscript/src/signingscript/sign.py
```diff
@@ -1512,7 +1536,38 @@ async def rcodesign_notarize(paths, notarization_workdir, creds_path):
...
         ]
-        await utils.execute_subprocess(command)
+
+        # Similar implementation to utils.execute_subprocess, but handling some errors:
+        message = 'Running "{}"'.format(" ".join(command))
+        log.info(message)
+        subprocess = await asyncio.create_subprocess_exec(
+            *command,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.STDOUT,
```
jcristau: nit: it'd be nice to avoid lumping stderr and stdout together, presumably we're only looking for the error message in one of them.

signingscript/src/signingscript/sign.py
```diff
@@ -1512,7 +1536,38 @@ async def rcodesign_notarize(paths, notarization_workdir, creds_path):
             creds_path,
             app_path,
         ]
-        await utils.execute_subprocess(command)
+
+        # Similar implementation to utils.execute_subprocess, but handling some errors:
+        message = 'Running "{}"'.format(" ".join(command))
```
jcristau: use a f-string?

signingscript/src/signingscript/sign.py
```diff
@@ -1512,7 +1536,38 @@ async def rcodesign_notarize(paths, notarization_workdir, creds_path):
...
+                raise SigningScriptError("Unknown error from notarization service.")
+            log.info("Connection reset by peer, attempting to recover")
+            for line in output_lines:
+                if "submission ID: " in line:
+                    submission_id = line.split(": ")[-1]
+                    break
+            else:
+                raise SigningScriptError("Unable to find notarization submission ID from logs.")
+            await rcodesign_notary_wait(submission_id, creds_path)
+            await rcodesign_staple(app_path)
```
bhearsum: If I understand how all of the `rcodesign` commands work correctly, it looks like we may end up either stapling in the original `notary-submit` command, or down here. I'd strongly suggest that you change this, and make the original command only submit, so that we're always stapling in exactly the same way. This will both make it easier to understand the flow of things now, and also avoid future pitfalls (such as stapling potentially being done with different arguments if we ever make changes there). It may also make it easier to parallelize in the future, since submission will no longer be a long lived subprocess.

signingscript/src/signingscript/sign.py
```diff
@@ -1512,7 +1536,38 @@ async def rcodesign_notarize(paths, notarization_workdir, creds_path):
...
+            if not output_lines or "Connection reset by peer" not in output_lines[-1]:
+                raise SigningScriptError("Unknown error from notarization service.")
+            log.info("Connection reset by peer, attempting to recover")
+            for line in output_lines:
+                if "submission ID: " in line:
+                    submission_id = line.split(": ")[-1]
+                    break
+            else:
+                raise SigningScriptError("Unable to find notarization submission ID from logs.")
+            await rcodesign_notary_wait(submission_id, creds_path)
```
bhearsum: What happens if this commands return an error? It looks like we'll bail immediately, without any retries? It seems like `notary_wait` ought to be called in a loop, or at the very least with multiple retries.

(In fact, maybe `notary-submit` should be retried as well.)

bhearsum (changes requested):

jcristau (commented):
Agree with Ben's comments about keeping the flow the same with/without retries.

signingscript/src/signingscript/sign.py
hneiva: f-string in this case makes it harder to copy-pasta into the terminal :P

hneiva:
Updated the code so it's more straight forward, and separates the `rcodesign` interface into it's own file.

Still missing tests, I'll add that next, but code is functional.

hneiva:
Fixed and added unit tests. Ready for full review 🚀

signingscript/src/signingscript/rcodesign.py
```diff
@@ -0,0 +1,126 @@
...
+        stdout = (await proc.stdout.readline()).decode("utf-8").rstrip()
+        if stdout:
+            log.info(stdout)
+            output_lines.append(stdout)
+        # Handle stderr
+        stderr = (await proc.stderr.readline()).decode("utf-8").rstrip()
+        if stderr:
+            # Unfortunately a lot of outputs from rcodesign come out to stderr
+            log.warn(stderr)
+            output_lines.append(stderr)
```
bhearsum, line 44: I agree with @jcristau's earlier comments about mixing stdout and stderr (this is just another form of that). Presumably you know whether it's stdout or stderr that has the string you're looking for, so you should be able to keep these separate?

signingscript/src/signingscript/sign.py
```diff
@@ -1523,11 +1510,36 @@ def _can_notarize(filename, supported_extensions):
...
+        attempts=ATTEMPTS,
+        retry_exceptions=RCodesignError,
+    )
+    # Staple
+    await retry_async(
+        func=rcodesign_staple,
+        args=[path],
+        attempts=ATTEMPTS,
+        retry_exceptions=RCodesignError,
+    )
```
bhearsum, line 1536: Are you sure we need to call all of these? The `rcodesign` docs seems to suggest that your first function should be waiting and stapling all by itself. It says: "Or to wait and automatically staple the file if notarization was successful:"
```
rcodesign notary-submit \
  --api-key-path ~/.appstoreconnect/key.json \
  --staple \
  path/to/file/to/notarize
```

It seems like the first call being retried ought to do everything we need (at the cost of resubmitting entirely on _any_ failure).

I think a better way to approach this may be:
1) Submit without `--staple`, so that `notarize` returns right away. This should be retried because the submission may fail
2) Run `notary_wait` with retries (obviously the polling may fail for various reasons).
3) Run `staple`. There shouldn't be any need to retry here.

What do you think? Am I overlooking something?

signingscript/src/signingscript/rcodesign.py
hneiva: @jcristau comment was regarding sending `stderr=asyncio.subprocess.STDOUT` to `asyncio.create_subprocess_exec` - which wouldn't allow us to collect them at all.

Regardless, as stated on the comment above the `log.warn`, `rcodesign`doesn't only output errors to stderr (there's good data in it). Therefore separating them has the opposite desired effect.
Let me know if the inline comment isn't enough.

signingscript/src/signingscript/sign.py
hneiva: `staple` does reach out to the Apple servers to collect the ticket, therefore I see the benefit of having a retry. But I can change it if you still think it's a good idea.

bhearsum: OK, good to know. I think the overarching comment still remains though: the first command here (`notary-submit --staple`) already does all of: submission, polling, stapling. If we're already retrying _that_ command, what is the purpose of running `notary_wait` and `staple` separately?

hneiva: The way I changed it "wait and staple" is optional. In this case, it just returns the submission id, then we have the wait, and then stapling.
```
# function:
async def rcodesign_notarize(app_path, creds_path, staple=False):
.......
# usage:
    submission_id = await retry_async(
        func=rcodesign_notarize,
        args=(path, creds_path), # <--- no staple
```
So, the first command only submits and returns the submission id.

I think you missed the change in the `notary-submit` code 🙃

bhearsum: Oh right, sorry. I somehow glossed over the fact that `--staple` was optional, heh. Alright, proceed!

bhearsum (approved):

PR #717, opened 2023-05-08
ci: Prebuild docker image before pinning packages

hneiva:
Speeds up pinning by around 2 minutes in total.

### DIFF
--- a/maintenance/pin-helper.sh
+++ b/maintenance/pin-helper.sh
@@ -13,15 +13,6 @@ else
     exit 1
 fi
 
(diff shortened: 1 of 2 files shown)


bhearsum (approved):
Just to be clear: this works because we're only running the apt & pip commands once (rather than once per container)?

hneiva:
Correct. Instead of running apt update/install + pip install each time we launch a container, we do it upfront on an image, and reuse that image for subsequent pinning payloads

PR #718, opened 2023-05-10
Update pkgbuild call to preserve ownership

krispyfries:
When pkgbuild runs its recommended ownership, it may set pkg ownership to `root:wheel`. This creates a problem in Firefox (https://bugzilla.mozilla.org/show_bug.cgi?id=1812480) where Firefox installs from .pkg require a password prompt to update. By preserving ownership, we assume that we have started with a clean ownership and will maintain it.


jcristau:
Do you know what sets the ownership in the first place?  (Trying to understand whether/why "assume we have started with a clean ownership" is reasonable)

krispyfries:
Closing this after conversation yesterday. In brief:

- problem further up the line creates incorrect ownership
- Future modernization efforts will help to correct the issue

PR #719, opened 2023-05-11
beetmover: disable s3 uploads to net-mozaws-{stage,prod}-delivery-{archive,firefox}

jcristau:
Per https://bugzilla.mozilla.org/show_bug.cgi?id=1832287#c1 we don't need to keep uploading these to s3 in addition to gcs.

Uploads to partner and maven buckets remain enabled at this point.

### DIFF
--- a/beetmoverscript/docker.d/worker.yml
+++ b/beetmoverscript/docker.d/worker.yml
@@ -230,7 +230,7 @@ clouds:
         'COT_PRODUCT == "firefox" && (ENV == "dev" || ENV == "fake-prod")':
           dep:
             fail_task_on_error: True
-            enabled: True
+            enabled: False
             credentials:
               id: { "$eval": "DEP_ID" }
               key: { "$eval": "DEP_KEY" }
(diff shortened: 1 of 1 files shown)


beetmoverscript/docker.d/worker.yml
```diff
@@ -433,7 +433,7 @@ clouds:
               nightly_components: 'maven-nightly-s3-upload-bucket-d4zm9oo354qe'
           nightly:
             fail_task_on_error: True
-            enabled: True
+            enabled: False
```
bhearsum, line 436: I see `nightly` is getting disabled for prod but not fake-prod and dev - is that an oversight?

jcristau: It was, thanks.  I think I got distracted wondering whether it was time to get rid of the TODO.

bhearsum (approved):

jcristau:
Reverted for breaking gecko releases.

PR #723, opened 2023-05-15
Bug 1832307 - [beetmover] Use new GCS credential env vars

ahal:
This uses the new format that separates env vars for prod vs dep.

PR #725, opened 2023-05-16
fix(beetmover): remove dep configs from release contexts

ahal:
We determined there's no reason to need to be able to run dep beetmover tasks from a level 3 context. To simplify things and prevent a possible source of confusion, let's remove the dep creds from prod environments.


ahal:
I decided to include this as part of https://github.com/mozilla-releng/scriptworker-scripts/pull/723

PR #742, opened 2023-05-24
Bug 1834836 - [beetmoverscript] Set app-services GCS credentials for …

ahal:
…stage and prod


ahal:
Initially I was going to do this at the same time as the general GCS credential switch, but that got backed out due to problems on mobile. So I'm landing this first to help unblock work with app-services (and if these creds also have problems, there are no pre-existing tasks to break, so it's fine).

bhearsum (approved):

PR #744, opened 2023-05-25
feat: signingscript notarization - add support for geckodriver

hneiva:


signingscript/src/signingscript/sign.py
```diff
@@ -1587,6 +1602,9 @@ async def apple_notarize(context, path, *args, **kwargs):
         shutil.rmtree(notarization_workdir, ignore_errors=True)
     os.mkdir(notarization_workdir)
 
+    if "geckodriver" in os.path.basename(path):
```
hneiva: TBH I don't love this. I think it should be a different behavior.
Thoughts?

bhearsum: Heh, I was wondering about this as well. My my first thought is: why do we have `supported_files` at all? Is it expensive to fail on unsupported files? Do we have cases where we pass both supported and unsupported files and must filter the latter out? (If the latter maybe we should stop passing unsupported files....)

I'm not certain I understand why a different behaviour would be beneficial (presumbly you'd still be sanity checking the file name in some way)?

bhearsum: Just to be clear - I'm not _actually_ suggesting that you expand the scope of this to changing or getting rid of `supported_files` - just food for thought for the future.

hneiva: In some cases, we receive some files (like `.background` for DMGs) and we don't notarize those.
So, as long as there's something we can notarize, we assume it's expected to be skipped.

bhearsum (approved):

PR #745, opened 2023-05-26
Create sig files for the new macOS GMP Executable

hafta:
On macOS, the GMP child process will transition from using the plugin-container executable to a new executable named "<App Name> Media Plugin Helper". Add the different executable names for each channel to the list of files which we need to generate .sig files for.


iscript/src/iscript/autograph.py
```diff
@@ -32,6 +32,10 @@
 # Blessed files call the other widevine files.
 _WIDEVINE_BLESSED_FILENAMES = (
     # plugin-container is the top of the calling stack
```
jcristau: Can you move the comment to stay next to plugin-container?  And/or update it to explain what media plugin helper is?

hafta: Added comment to explain these are the files that load the Widevine CDM and what the new entries are for.

jcristau (approved):
Thanks!

PR #746, opened 2023-05-31
Bug 1835922 - bump signingscript termination grace period again

jcristau:
https://bugzilla.mozilla.org/show_bug.cgi?id=1835922


hneiva (approved):
🚀

PR #760, opened 2023-06-05
beetmoverscript: remove unreachable code

jcristau:
Since the switch to declarative artifacts, move_beets is always called with an artifact_map, not a manifest.

### DIFF
--- a/beetmoverscript/tests/test_script.py
+++ b/beetmoverscript/tests/test_script.py
@@ -234,14 +234,13 @@ def restore_buildhub_file():
 
 # move_beets {{{1
 @pytest.mark.asyncio
-@pytest.mark.parametrize("task_filename", ("task.json", "task_artifact_map.json"))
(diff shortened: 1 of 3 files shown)


gabrielbusta (approved):
Good catch

beetmoverscript/tests/test_script.py
```diff
@@ -386,8 +381,8 @@ async def test_move_beets_raises(mocker):
     context.balrog_manifest = list()
     context.artifacts_to_beetmove = get_upstream_artifacts(context)
 
-    with pytest.raises(ScriptWorkerTaskException):
-        await move_beets(context, context.artifacts_to_beetmove, manifest=None, artifact_map=None)
+    with pytest.raises(ScriptWorkerTaskException) as excinfo:
```
jcristau:
```suggestion
    with pytest.raises(ScriptWorkerTaskException):
```

PR #762, opened 2023-06-07
Bug 1822689 - Modify pushflatpakscript to allow for …

dandarnell:
…pushing Thunderbird Flatpak to Flathub.

The official Thunderbird Flatpak will be ready for distribution to Flathub in the near future, and the pushing process makes use of the same scriptworker as Firefox's Flathub push.

### DIFF
--- a/pushflatpakscript/src/pushflatpakscript/flathub.py
+++ b/pushflatpakscript/src/pushflatpakscript/flathub.py
@@ -139,6 +139,23 @@ def check_and_extract_tar_archive(context, tar_file_path):
     return os.path.join(flatpak_tar_basedir, flatpak_deflated_dir)
 
 
+def check_app_id_matches_flatpak(context, flatpak_path):
+    # Extract all ostree refs from the supplied Flatpak repo
+    flatpak_refs = subprocess.check_output(["ostree", "refs"], cwd=flatpak_path).decode().splitlines()
+    
+    # Consolidate ostree refs into list of Flatpak IDs available in repo
+    flatpak_refs = [ref.split("/")[1] for ref in flatpak_refs if ref.startswith("app/")]
+    
+    # Create a list, if any, of all unexpected Flatpak IDs present in repo
+    invalid_refs = set(flatpak_refs) - {context.config["app_id"]}
+
+    if context.config["app_id"] not in flatpak_refs:
+        raise TaskVerificationError(f"Supplied app ID ({context.config['app_id']}) is not present in Flatpak!")
+    
+    if len(invalid_refs) > 0:
+        raise TaskVerificationError(f"One or more invalid app IDs are present in Flatpak!")
+
+
 def sanitize_buildid(bytes_input):
     """Flathub API returns bytes to we're decoding that to unicode string"""
     return bytes_input.decode().strip()
@@ -167,6 +184,9 @@ def push(context, flatpak_file_path, channel):
 
     log.info("Unpacking the tarball ...")
     deflated_dir = check_and_extract_tar_archive(context, flatpak_file_path)
+
+    log.info("Verifying supplied app ID matches flatpak app ID...")
+    check_app_id_matches_flatpak(context, deflated_dir)
 
     log.info(f"Pushing the flatpak to the associated {publish_build_output}")
     run_flat_manager_client_process(context, token_args + ["push", publish_build_output, deflated_dir])
--- a/pushflatpakscript/src/pushflatpakscript/task.py
+++ b/pushflatpakscript/src/pushflatpakscript/task.py
@@ -1,7 +1,5 @@
 from scriptworker.exceptions import TaskVerificationError
 
-FLATPAK_SCOPES_PREFIX = "project:releng:flathub:firefox:"
-
 _CHANNELS_AUTHORIZED_TO_REACH_FLATHUB = ("beta", "stable")
 ALLOWED_CHANNELS = ("mock", *_CHANNELS_AUTHORIZED_TO_REACH_FLATHUB)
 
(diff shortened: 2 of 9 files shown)


jcristau (changes requested):
Thanks for the patch :)

A couple of things:
- please fix the test failure
- also, the scope prefix (currently in pushflatpakscript/src/pushflatpakscript/task.py) needs to become part of the config, so it's different for firefox and thunderbird.

jfx2006:
This looks a lot better to me. Only comment I have is the tests should probably be expanded a bit to check that the TASKCLUSTER_SCOPE_PREFIX for Firefox only does Firefox pushes. I am missing the same in PR #775.

pushflatpakscript/src/pushflatpakscript/task.py
```diff
@@ -1,18 +1,16 @@
...
 ALLOWED_CHANNELS = ("mock", *_CHANNELS_AUTHORIZED_TO_REACH_FLATHUB)
 
 
 def get_flatpak_channel(config, task):
     payload = task["payload"]
     if "channel" not in payload:
-        raise TaskVerificationError(f"channel must be defined in the task payload. Given payload: {payload}")
-
+        raise TaskVerificationError(f"Channel must be defined in the task payload. Given payload: {payload}")
+
```
jcristau: stray whitespace

jcristau (commented):
Thanks, this looks almost ready.

One thing I'd like to see is protection against thunderbird uploading a firefox flatpak, and vice versa.  (Can we e.g. somehow extract the app id from the tarball, to check against an allowed prefix?)

dandarnell:
> Thanks, this looks almost ready.
> 
> One thing I'd like to see is protection against thunderbird uploading a firefox flatpak, and vice versa. (Can we e.g. somehow extract the app id from the tarball, to check against an allowed prefix?)

I've added a check that verifies the app ID being pushed matches the app ID in the decompressed Flatpak.

pushflatpakscript/src/pushflatpakscript/flathub.py
```diff
@@ -138,6 +138,13 @@ def files(tar, path):
     return os.path.join(flatpak_tar_basedir, flatpak_deflated_dir)
 
 
+def check_app_id_matches_flatpak(context, flatpak_path):
+    flatpak_app_ref_dir = flatpak_path + "/refs/heads/app/" + context.config["app_id"]
+
+    if not os.path.exists(flatpak_app_ref_dir):
```
jcristau: Is this right?  It seems as though you could craft a flatpak with both legit and bogus app refs and this check would be happy?
I'd be more defensive here and ensure that is the only app in the repo (which maybe means we need to install `ostree` in the pushflatpakscript image so we can run commands against it).

dandarnell: It turns out ostree was already available in the pushflatpakscript Docker image, so I updated this patch to use ostree to verify that the correct app is present in the Flatpak, and that no other apps are present.

dandarnell:
The patch now uses ostree to verify that a) the specified app ID is available in the supplied Flatpak, and b) no other app IDs are in the Flatpak.

pushflatpakscript/src/pushflatpakscript/flathub.py
```diff
@@ -138,6 +138,23 @@ def files(tar, path):
...
 
+def check_app_id_matches_flatpak(context, flatpak_path):
+    # Extract all ostree refs from the supplied Flatpak repo
+    flatpak_refs = subprocess.check_output(["ostree", "refs"], cwd=flatpak_path).decode().splitlines()
+    
+    # Consolidate ostree refs into list of Flatpak IDs available in repo
+    flatpak_refs = [ref.split("/")[1] for ref in flatpak_refs if ref.startswith("app/")]
+    
+    # Create a list, if any, of all unexpected Flatpak IDs present in repo
+    invalid_refs = [ref for ref in flatpak_refs if not context.config["app_id"]]
```
jcristau: That check looks broken, presumably what you want is something like `invalid_refs = set(flatpak_refs) - {context.config["app_id"]}`.

If possible we should have a test to cover that case.

jcristau (changes requested):
I think we're almost there.  Thanks :)

pushflatpakscript/src/pushflatpakscript/flathub.py
dandarnell: I've updated and tested the patch with your suggested change. Much more elegant than what I was trying to do!

dandarnell: If possible, I'd like to write the test in a separate PR so this can land and we can test our Flatpak pushing before Flathub implements their breaking changes in a few weeks.

jcristau (approved):

pushflatpakscript/src/pushflatpakscript/flathub.py
```diff
@@ -138,6 +138,23 @@ def files(tar, path):
     return os.path.join(flatpak_tar_basedir, flatpak_deflated_dir)
 
 
+def check_app_id_matches_flatpak(context, flatpak_path):
+    # Extract all ostree refs from the supplied Flatpak repo
+    flatpak_refs = subprocess.check_output(["ostree", "refs"], cwd=flatpak_path).decode().splitlines()
```
jcristau, line 144: Actually, let's install `ostree` explicitly in the Dockerfile, to avoid bad surprises later on.

jfx2006: > Actually, let's install `ostree` explicitly in the Dockerfile, to avoid bad surprises later on.

See PR #860.
There's also a PR in the scriptworker repo to go with this.

PR #767, opened 2023-06-21
Bug 1839649 - beetmover: add components and nightly_components maven …

jcristau:
…staging buckets for gecko


gbrownmozilla (changes requested):
I think you just need to update docker.d/init_worker.sh for the firefox case -- otherwise, looks good.

jcristau:
Good call.  Updated.

gbrownmozilla (approved):

PR #772, opened 2023-06-29
Bug 1840440 - beetmoverscript: remove private fxpartner buckets

jcristau:
The partner portal has been decommissioned and the buckets didn't get migrated to gcp.


hneiva (approved):
Oh nice. Thanks for putting this together!

PR #775, opened 2023-07-11
Bug 1817658 - Add support for Thunderbird to pushmsixscript.

jfx2006:


pushmsixscript/docker.d/init_worker.sh
```diff
@@ -36,7 +34,16 @@ case $ENV in
...
 
 case $COT_PRODUCT in
-  firefox) ;;
+  firefox)
+    export BETA_APPLICATION_ID=9NZW26FRNDLN
+    export RELEASE_APPLICATION_ID=9NZVDKPMR9RD
+    export TASKCLUSTER_SCOPE_PREFIX="project:releng:microsoftstore:"
+    ;;
+  thunderbird)
+    export BETA_APPLICATION_ID=ZZZZZTBDZZZZZ
```
jfx2006: Still working on getting the store side of things setup, will fill in these values when they are available.

bhearsum:
We should probably block landing this on having separate credentials for Firefox and Thunderbird msixscriptworkers, so that there's no possibility of one attempting to ship to the other.

@gbrownmozilla - you should probably have a look at this patch as well, since you're most familiar with this code.

JohanLorenzo:
> We should probably block landing this on having separate credentials for Firefox and Thunderbird msixscriptworkers

I chatted with @Sancus. He confirmed the Thunderbird team has a dedicated account. So, I agree, let's not reuse the Firefox creds here 🙂

bhearsum:
> > We should probably block landing this on having separate credentials for Firefox and Thunderbird msixscriptworkers
> 
> I chatted with @Sancus. He confirmed the Thunderbird team has a dedicated account. So, I agree, let's not reuse the Firefox creds here 🙂

For the record, this will require changes to secrets, and maybe additional changes to this repo as well.

jfx2006:
IIUC, The credentials are set init_worker.sh?
```sh
    test_var_set 'TENANT_ID'
    test_var_set 'CLIENT_ID'
    test_var_set 'CLIENT_SECRET'
    export TENANT_ID=`echo $TENANT_ID | base64 -d`
    export CLIENT_ID=`echo $CLIENT_ID | base64 -d`
    export CLIENT_SECRET=`echo $CLIENT_SECRET | base64 -d`
```

And those come from some Cloud-Ops repo that I can't see AIUI. Different workers for Thunderbird and Firefox to prevent overlap.

bhearsum:
> IIUC, The credentials are set init_worker.sh?
> 
> ```shell
>     test_var_set 'TENANT_ID'
>     test_var_set 'CLIENT_ID'
>     test_var_set 'CLIENT_SECRET'
>     export TENANT_ID=`echo $TENANT_ID | base64 -d`
>     export CLIENT_ID=`echo $CLIENT_ID | base64 -d`
>     export CLIENT_SECRET=`echo $CLIENT_SECRET | base64 -d`
> ```
> 
> And those come from some Cloud-Ops repo that I can't see AIUI. Different workers for Thunderbird and Firefox to prevent overlap.

Ah, indeed. I was thinking about https://github.com/mozilla-releng/scriptworker-scripts/blob/master/signingscript/docker.d/passwords.yml, where we have lots of per-product switches.

We'll certainly have to add some entries to sops, as well as the Jenkinsfile (https://github.com/mozilla-services/cloudops-infra/blob/master/projects/relengworker/Jenkinsfile).

gbrownmozilla (changes requested):
This looks great! I can't find anything wrong or concerning, and I am excited to see pushmsix being generalized to another product - thanks!

I think we should get the secrets and cloudops-infra changes in place before landing this PR, so I am "requesting changes" only to clear the review request until those changes are landed; I will gladly approve once those changes are sorted out (and I understand @bhearsum will help with that).

PR #777, opened 2023-07-13
fix: Add error checking on rcodesign wait

hneiva:


signingscript/src/signingscript/rcodesign.py
```diff
@@ -100,12 +101,33 @@ async def rcodesign_notary_wait(submission_id, creds_path):
...
 
+async def rcodesign_check_result(logs):
+    """Checks notarization results
+    rcodesign cli call can exit with 0 even though the notarization has failed
+    Args:
+        logs (list<str>): output from polling result
+    """
+    re_inprogress = re.compile("^poll state after.*InProgress$")
+    re_accepted = re.compile("^poll state after.*Accepted$")
+    re_invalid = re.compile("^poll state after.*Invalid$")
```
hneiva, line 120: For some reason `rcodesign` doesn't seem to error (exitcode > 0) when the package is invalid.

signingscript/src/signingscript/rcodesign.py
```diff
@@ -100,12 +101,33 @@ async def rcodesign_notary_wait(submission_id, creds_path):
...
+    for line in logs:
+        if re_inprogress.search(line):
```
gbrownmozilla, line 122: I think you could skip the check for in-progress entirely, but that's only an advantage if the log is very long and performance is a concern...and I suppose it is nice to be explicit about the common case.

gbrownmozilla (approved):

signingscript/src/signingscript/rcodesign.py
hneiva: It's not set to check the logs as they come in, only after the process is done.
So the log lines would look like this:
```
poll state after 50s: InProgress
poll state after 53s: InProgress
poll state after 57s: InProgress
poll state after 60s: InProgress
poll state after 64s: Accepted
.... <result logs>
```
So the `continue` on the first few lines, then return the function once it hits `Accepted`.

signingscript/src/signingscript/rcodesign.py
jcristau: Did you file an issue against rcodesign that could be referenced in a comment here?

hneiva: I was debating it.
The command used here was `notary-wait` which technically "waited" successfully, the fact that the logs say the package is invalid might not warrant an exit code from rcodesign itself.

jcristau: Well I think it should :)

PR #839, opened 2023-08-17
Use dev bounceradmin for dev bouncer scriptworkers

bhearsum:
As part of https://bugzilla.mozilla.org/show_bug.cgi?id=1819405 / https://mozilla-hub.atlassian.net/browse/SVCSE-1186 we are going to make it possible to use the new dev bounceradmin & bouncer instances in staging releases. To this end, we need to point dev bouncerscript instances at dev bounceradmin.

I'm also making the staging entry more explicit here, but that should be a no-op.

@jcristau - I think you had thoughts or concerns about this - so I'm flagging you explictly.

### DIFF
--- a/bouncerscript/docker.d/worker.yml
+++ b/bouncerscript/docker.d/worker.yml
@@ -5,9 +5,9 @@ taskcluster_scope_prefix: { "$eval": "TASKCLUSTER_SCOPE_PREFIX" }
 bouncer_config:
   $merge:
     $match:
-      'COT_PRODUCT == "firefox"':
+      'COT_PRODUCT == "firefox" && (ENV == "dev" || ENV == "fake-prod")':
         project:releng:bouncer:server:staging:
-          api_root: 'https://bounceradmin-default.stage.mozaws.net/api'
+          api_root: 'https://dev.bounceradmin.nonprod.webservices.mozgcp.net/api'
           timeout_in_seconds: 60
           username: { "$eval": "BOUNCER_USERNAME" }
           password: { "$eval": "BOUNCER_PASSWORD" }
@@ -19,9 +19,9 @@ bouncer_config:
           username: { "$eval": "BOUNCER_USERNAME" }
           password: { "$eval": "BOUNCER_PASSWORD" }
 
-      'COT_PRODUCT == "thunderbird"':
+      'COT_PRODUCT == "thunderbird" && (ENV == "dev" || ENV == "fake-prod")':
         project:comm:thunderbird:releng:bouncer:server:staging:
-          api_root: 'https://bounceradmin-default.stage.mozaws.net/api'
+          api_root: 'https://dev.bounceradmin.nonprod.webservices.mozgcp.net/api'
           timeout_in_seconds: 60
           username: { "$eval": "BOUNCER_USERNAME" }
           password: { "$eval": "BOUNCER_PASSWORD" }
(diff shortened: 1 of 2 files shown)


bouncerscript/docker.d/worker.yml
```diff
@@ -5,7 +5,14 @@ taskcluster_scope_prefix: { "$eval": "TASKCLUSTER_SCOPE_PREFIX" }
 bouncer_config:
   $merge:
     $match:
-      'COT_PRODUCT == "firefox"':
+      'COT_PRODUCT == "firefox" && ENV == "dev"':
```
jcristau: Don't we want to use this same bouncer instance for fake-prod workers?

bhearsum: I think it depends on what we want the default to be when pushing to try. I'm not opposed to it, but I seem to recall some concern around using the dev instance by default here.

jcristau: If possible I would like the default to be such that staging releases mostly "just work", i.e.:
- the bounceradmin instance behind release-bouncer-aliases in fake-prod and the bouncer instance behind release-bouncer-check use the same db
- the bouncer instance behind release-bouncer-check redirects to stage ftp, i.e. where fake-prod beetmover uploads artifacts.  AIUI, this means using the new "dev" bouncer.

Does that make sense?

bhearsum: Yes, this is exactly what I want too! (Also: it allows us to bouncer Balrog stage at bouncer dev, which means that final verify will be capable of passing.)

I'll update this to point fake-prod at dev bounceradmin.

There will also be a Gecko change needed to update the bouncer instance we use there which I'm currently working on.

bhearsum: I've updated this patch to reflect this. We'll also need a change to the relengworker secrets before this deploys.

https://phabricator.services.mozilla.com/D186634 is the Gecko patch we'll want.

bouncerscript/docker.d/worker.yml
```diff
@@ -5,9 +5,16 @@ taskcluster_scope_prefix: { "$eval": "TASKCLUSTER_SCOPE_PREFIX" }
...
```
jcristau: nit: looks like the common pattern is `'COT_PRODUCT == "firefox" && (ENV == "dev" || ENV == "fake-prod")'` instead of repeating

bouncerscript/docker.d/worker.yml
```diff
@@ -5,9 +5,16 @@ taskcluster_scope_prefix: { "$eval": "TASKCLUSTER_SCOPE_PREFIX" }
...
         project:releng:bouncer:server:staging:
-          api_root: 'https://bounceradmin-default.stage.mozaws.net/api'
```
jcristau, line 10: Do we want to keep the old stage bounceradmin as an option with its own scope or remove it?

jcristau (approved):
I guess one thing worth noting is that with this change, staging is no longer an option for prod workers. I think that's fine.

bouncerscript/docker.d/worker.yml
bhearsum: Good call - will do.

bouncerscript/docker.d/worker.yml
bhearsum: I don't see any value in being able to use it, personally. For the purposes of level 1 workers it is misconfigured IMO (being it doesn't link to the same place level 1 workers publish to).

bhearsum:
@jcristau - I had to make an additional change here after discovering we hardcode (ugh) the URL to the public bouncer endpoint in `constants.py`. I've just updated that in place for now, but if pushed I'll move it into `worker.yml`.

jcristau (approved):

PR #843, opened 2023-09-14
Bug 1567912 - [pushapkscript] Support pushing AAB files to Google Play

gbrownmozilla:
* re-pin pushapkscript dependencies to pick up mozapkpublisher 6.2.0 with AAB support
 * update pushapkscript to use push_aab for "*.aab" upstream artifacts
 * add AAB unit tests


pushapkscript/Dockerfile.test
```diff
@@ -4,10 +4,15 @@ FROM python:${PYTHON_VERSION}
...
+RUN apt-get update && \
+    apt-get install -y default-jdk && \
+    apt-get clean && \
+    curl -L https://github.com/google/bundletool/releases/download/1.15.4/bundletool-all-1.15.4.jar -o /app/bundletool.jar
+
+ENV BUNDLETOOL_PATH=/app/bundletool.jar
+
 COPY MANIFEST.in setup.py tox.ini /app/
 COPY requirements/ /app/requirements/
-COPY docker.d/image_setup.sh /app/
```
gbrownmozilla, line 9: Not related to the AAB efforts...I just noticed that image_setup.sh is obsolete.

pushapkscript/Dockerfile
```diff
@@ -27,6 +27,9 @@ RUN python -m venv /app \
  && cd /app/configloader \
  && /app/configloader_venv/bin/pip install -r requirements/base.txt \
  && /app/configloader_venv/bin/pip install . \
- && cd /app
+ && cd /app \
+ && curl -L https://github.com/google/bundletool/releases/download/1.15.4/bundletool-all-1.15.4.jar -o /app/bundletool.jar
```
gbrownmozilla: pushflatpakscript does a similar curl download in its Dockerfile

pushapkscript/Dockerfile
```diff
@@ -27,6 +27,9 @@ RUN python -m venv /app \
...
+
+ENV BUNDLETOOL_PATH=/app/bundletool.jar
```
gbrownmozilla, line 37: pushflatpakscript sets a similar environment variable in its init_worker.sh, but this seemed more straight-forward

pushapkscript/Dockerfile
JohanLorenzo: Sounds good! As an improvement, let's enforce the checksum here. This will remove this supply-chain attack vector. We have an example there: https://github.com/mozilla-releng/staging-mozilla-vpn-client/blob/391f3c05a766e299ee68a3ca50d645e358fdd1ad/taskcluster/docker/conda-base/Dockerfile#L8-L15

pushapkscript/Dockerfile.test
```diff
@@ -4,10 +4,15 @@ FROM python:${PYTHON_VERSION}
 
 WORKDIR /app
 
+RUN apt-get update && \
+    apt-get install -y default-jdk && \
+    apt-get clean && \
+    curl -L https://github.com/google/bundletool/releases/download/1.15.4/bundletool-all-1.15.4.jar -o /app/bundletool.jar
```
JohanLorenzo: Improvement: let's enforce the checksum here too.

pushapkscript/src/pushapkscript/script.py
```diff
@@ -29,18 +29,26 @@ async def async_main(context):
...
     else:
-        log.info('This product is configured with "skip_check_signature", so the signing of the ' "APK will not be verified.")
+        log.info('This product is configured with "skip_check_signature", so the signing of the ' "files will not be verified.")
 
     log.info("Delegating publication to mozapkpublisher...")
     with contextlib.ExitStack() as stack:
         files = [stack.enter_context(open(apk_file_name)) for apk_file_name in all_apks_paths]
         publish.publish(product_config, publish_config, files, contact_server)
+        files = [stack.enter_context(open(aab_file_name)) for aab_file_name in all_aabs_paths]
+        publish.publish_aab(product_config, publish_config, files, contact_server)
```
JohanLorenzo, line 55: Question: What's the reasoning for having a single pushapk task that handles both APKs and the AAB? If I understand correctly, switching to AABs is permanent so we won't have both at the same time. I wonder if this could cause confusion for people who have little context on APKs vs AABs. What do you think?

JohanLorenzo (approved):
Looks good to me overall! I have one question about allowing AABs in the same task as APKs but I won't block on that - this could be addressed in a follow-up, if needed.

pushapkscript/src/pushapkscript/script.py
gbrownmozilla: Doing it this way delegates the issue to the task configuration: We can have one task with both APK and AAB upstream artifacts, or separate tasks with only APK (or AAB) upstream artifacts; pushapkscript will just push whatever artifacts are configured. I suppose it is odd to have "pushapkscript" pushing AABs but I don't think it is worth renaming everything now.

pushapkscript/Dockerfile
gbrownmozilla: Love it! Thank you!

pushapkscript/src/pushapkscript/script.py
JohanLorenzo: Thanks for the context!

> We can have one task with both APK and AAB upstream artifacts, [...] pushapkscript will just push whatever artifacts are configured.

If I'm not mistaken, this case won't work, right? If a product turned AABs on, then Google Play will refuse any APKs submitted, won't it? (Just making sure I didn't misunderstand anything obvious 🙂)

gbrownmozilla: Right! We should configure tasks to push APKs or AABs, but not both.

...so let's add a check that no one tries to do both...

pushapkscript/tests/test_script.py
```diff
@@ -15,12 +15,20 @@
...
 @pytest.mark.asyncio
-@pytest.mark.parametrize("android_product", ("aurora", "beta", "release", "dep", "focus"))
-async def test_async_main(monkeypatch, android_product):
+@pytest.mark.parametrize("android_product, upstream_artifacts, is_aab, raises", 
+    (
+        ("beta", { "someTaskId": ["/some/path/to/one.apk", "/some/path/to/another.apk"], "someOtherTaskId": ["/some/path/to/yet_another.apk"]}, False, False ),
+        ("release", { "someTaskId": ["/some/path/to/one.apk", "/some/path/to/another.apk"], "someOtherTaskId": ["/some/path/to/yet_another.apk"]}, False, False ),
+        ("dep", { "someTaskId": ["/some/path/to/one.aab", "/some/path/to/another.aab"], "someOtherTaskId": ["/some/path/to/yet_another.aab"]}, True, False ),
+        ("focus", { "someTaskId": ["/some/path/to/one.aab", "/some/path/to/another.aab"], "someOtherTaskId": ["/some/path/to/yet_another.aab"]}, True, False ),
+        ("aurora", { "someTaskId": ["/some/path/to/one.aab", "/some/path/to/another.aab"], "someOtherTaskId": ["/some/path/to/yet_another.apk"]}, True, True ),
```
JohanLorenzo, line 24: Thanks for testing this edge case 👍

PR #845, opened 2023-09-18
Bug 1567912 - [pushapkscript] fixup dockerfile sha check

gbrownmozilla:


gbrownmozilla:
Changes made in my earlier PR fail on deployment, apparently because of the use of ARG:
https://firefox-ci-tc.services.mozilla.com/tasks/FpvI8zAOSwCCe6sa_VffVg/runs/0/logs/public/logs/live.log

hneiva (approved):

PR #846, opened 2023-09-19
Re-enable s3 uploads

bhearsum:
Disabling them broke buildhub2: https://github.com/mozilla-releng/buildhub2/issues/1042


gbrownmozilla (approved):
Thanks!

PR #848, opened 2023-09-26
re-pin all requirements

gbrownmozilla:


gbrownmozilla:
Changes generated by `maintenance/pin.sh`.

bhearsum (approved):
This all looks pretty safe. The only major bump is `build`, which comes in indirectly via `check-manifest` and `pip-tools`. Its changelog doesn't suggest anything that will bust.

PR #850, opened 2023-09-27
Bug 1855470 - treescript: allow version bumps of mobile/android/version.txt

jcristau:


gbrownmozilla (approved):
I am not very familiar with treescript, but this seems like a great idea and lgtm.

PR #851, opened 2023-09-28
Bug 1567912 - Add AAB signatures on staging for reference-browser

gbrownmozilla:


gbrownmozilla:
As a next step in the AAB project, we want to try signing the reference-browser AAB with staging autograph, and then pushing that signed AAB to Google Play.

signingscript/docker.d/init_worker.sh
```diff
@@ -206,6 +206,8 @@ case $ENV in
         test_var_set 'GPG_PUBKEY_PATH'
         test_var_set 'AUTOGRAPH_REFERENCE_BROWSER_PASSWORD'
         test_var_set 'AUTOGRAPH_REFERENCE_BROWSER_USERNAME'
+        test_var_set 'AUTOGRAPH_STAGE_REFERENCE_BROWSER_PASSWORD'
```
gbrownmozilla, line 209: I'm adding new AUTOGRAPH_STAGE vars here, in mobile / PROD. We would normally access autograph stage from a staging environment...but I don't think we have a staging-reference-browser. Will this be OK?

signingscript/src/signingscript/utils.py
```diff
@@ -210,7 +210,7 @@ def is_apk_autograph_signing_format(format_):
 
     """
     # TODO Remove autograph_focus once format is migrated
-    return format_ and format_.startswith("autograph_apk_") or format_ == "autograph_focus"
+    return format_ and format_.startswith("autograph_apk_") or format_ == "autograph_focus" or format_ == "autograph_stage_aab"
```
gbrownmozilla, line 213: Why didn't we need to handle the `autograph_stage_apk_v3` case here?

signingscript/docker.d/passwords.yml
```diff
@@ -294,6 +294,11 @@ in:
              {"$eval": "AUTOGRAPH_REFERENCE_BROWSER_PASSWORD"},
              ["autograph_apk"]
             ]
+          - ["https://autograph-external.stage.autograph.services.mozaws.net",
```
gbrownmozilla, line 297: Johan, in our chat you said to use new scopes. Is that necessary, or can I ride along with the existing `release-signing`?

signingscript/docker.d/init_worker.sh
JohanLorenzo: [staging-reference-browser](https://github.com/mozilla-releng/staging-reference-browser) exists 😃 I don't remember when it was used the last time, though.

signingscript/docker.d/passwords.yml
JohanLorenzo: Mmmmm, that's a good point! Yeah, new scopes are unnecessary. We're just adding a new format. I agree, let's use `release-signing`. That'll be consistent with Desktop and all the formats we have there.

signingscript/src/signingscript/utils.py
JohanLorenzo: Good question! I think we've never handled v3 signing in signingscript. Hal Wine and I had a conversation about it before the all hands. I'll try to add support if I find some time in the next week and if it doesn't conflict the work on AAB. The latter remains our priority.

JohanLorenzo (approved):
Looks good to me! Looking forward to seeing AABs signed with the new key!

PR #852, opened 2023-10-12
Update dependencies

jcristau:


jcristau:
Major bumps for taskcluster (55.3.0 to 56.0.1) and executing (1.2.0 to 2.0.0).
The former shouldn't be an issue (the backwards-compat changes are on the queue / service side).
The latter looks like it basically just dropped python2 support.

bhearsum (approved):

PR #855, opened 2023-10-13
[signingscript] Remove old scopes from SCOPES.md

ahal:
The `format:` scopes are not validated anywhere and are not necessary in
task definitions. For Gecko, they were removed in:
https://hg.mozilla.org/mozilla-central/rev/4b3b6bb262a03314eda0bfd0fcc00d8e4407f32b

Let's remove them from `SCOPES.md` to avoid confusion.


ahal:
Note, we should also remove grants for these in `ci-config`, but if projects are still setting them somewhere it would cause Decision task failures.

gbrownmozilla (approved):

PR #859, opened 2023-10-26
Bug 1859937 - Add comm-release support to bouncerscript.

jfx2006:


bouncerscript/src/bouncerscript/constants.py
```diff
@@ -7,6 +7,14 @@
     "thunderbird-latest-ssl": r"^Thunderbird-\d+\.\d+(\.\d+)?-SSL$",
     "thunderbird-msi-latest-ssl": r"^Thunderbird-\d+\.\d+(\.\d+)?-msi-SSL$",
     "thunderbird-msix-latest-ssl": r"^Thunderbird-\d+\.\d+(\.\d+)?-msix-SSL$",
+    "thunderbird-esr-latest": r"^Thunderbird-\d+\.\d+(\.\d+)?esr$",
```
jcristau, line 10: Are non-ssl aliases actually needed?

jcristau (approved):

bouncerscript/src/bouncerscript/constants.py
jfx2006: non-SSL releases refer to the ancient days when the US couldn't export encryption tech, correct?

Probably not technically needed... thunderbird.net does not refer to them.

jcristau: I'm not sure TBH, it predates my involvement by some years, I assume it's something like that, a relic from the days where https wasn't ubiquitous, which feels obsolete.  I'm happy to keep this here for now if you want though, it doesn't really hurt anything, and the client code can decide which aliases to actually populate/update, let me know.

PR #861, opened 2023-10-30
pushflatpak: pass flatpak-repackage log url to flat-manager-client (bug 1859510)

jcristau:


ahal (approved):
Do we know what the expectations are around this file? E.g, are they going to link to it in which case will it be a problem when it expires? Hopefully they download + host it themselves

jcristau:
> Do we know what the expectations are around this file? E.g, are they going to link to it in which case will it be a problem when it expires? Hopefully they download + host it themselves

@barthalion can you speak to that?

barthalion:
> Do we know what the expectations are around this file? E.g, are they going to link to it in which case will it be a problem when it expires? Hopefully they download + host it themselves

Apologies for the late answer. It doesn't need to be permanent -- the idea is to give users the possibility to establish a "chain of trust", and probably won't be looked beyond the latest release. Even if someone complains about that, we have no formal policy about how long the build log URL is supposed to be up anyway so it's really your call.

PR #864, opened 2023-11-07
fix: Refactor iscript provisioning profile config

hneiva:
Renames the name and path properties to profile_name and target_path


ahal (approved):
Lgtm!

PR #866, opened 2023-11-10
Bug 1864101 - Remove 'target_store' parameter from push_apk calls

gbrownmozilla:
The target_store parameter was removed from push_apk() in mozapkpublisher 6.3.0; this updates pushapkscript to match those expectations.

There are other uses of target_store in pushapkscript which could probably also be cleaned up now...let's decide that in a follow-up.


JohanLorenzo (approved):
LGTM!

PR #867, opened 2023-11-14
Restrict platformdirs to <4 due to constraints in virtualenv package

bhearsum:
This is due to https://github.com/pypa/virtualenv/issues/2666.


ahal (approved):

jcristau (commented):
The platformdirs pin could live in test.in, as far as I can tell, looks like it's pulled in by black in "test" environments, and by virtualenv, via tox, in "local" environments, and "base" isn't affected.

bhearsum:
May as well just let this get addressed in #869.

PR #871, opened 2023-11-15
bouncerscript: don't require successive nightly versions outside of prod

bhearsum:
This is part of the fix for https://bugzilla.mozilla.org/show_bug.cgi?id=1861749, where we want to make sure we're able to run `bouncer-locations` on Try, and we won't always be guaranteed to run it every cycle.

### DIFF
--- a/bouncerscript/docker.d/worker.yml
+++ b/bouncerscript/docker.d/worker.yml
@@ -2,6 +2,7 @@ work_dir: { "$eval": "WORK_DIR" }
 artifact_dir: { "$eval": "ARTIFACTS_DIR" }
 verbose: { "$eval": "VERBOSE == 'true'" }
 taskcluster_scope_prefix: { "$eval": "TASKCLUSTER_SCOPE_PREFIX" }
+require_successive_versions: { "$eval": "ENV == 'prod'" }
 bouncer_config:
   $merge:
     $match:
--- a/bouncerscript/src/bouncerscript/constants.py
+++ b/bouncerscript/src/bouncerscript/constants.py
@@ -78,19 +78,19 @@ PRODUCT_TO_DESTINATIONS_REGEXES = {
 
 _BOUNCER_PATH_REGEXES_PER_PRODUCT_DEFAULT = {
     "firefox-nightly-latest": (
-        r"^(/firefox/nightly/latest-mozilla-central-l10n/firefox-\d+\.0a1\.:lang\."
+        r"^(/firefox/nightly/latest-(mozilla-central|try)-l10n/firefox-\d+\.0a1\.:lang\."
         r"(?:linux-i686\.tar\.bz2|linux-x86_64\.tar\.bz2|mac\.dmg|win32\.installer\.exe|win64\.installer\.exe|win64-aarch64\.installer\.exe))$"
     ),
     "firefox-nightly-latest-ssl": (
-        r"^(/firefox/nightly/latest-mozilla-central/firefox-\d+\.0a1\.en-US\."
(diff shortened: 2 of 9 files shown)


bouncerscript/tests/test_script.py
```diff
@@ -153,13 +158,26 @@ async def test_bouncer_aliases(aliases_context, mocker):
...
+            False,
+        ),
+        (
+            [
+                {"os": "win", "id": "47767", "path": "/firefox/nightly/latest-try/firefox-62.0a1.en-US.win32.installer.exe"},
+                {"os": "linux", "id": "47764", "path": "/firefox/nightly/latest-try/firefox-62.0a1.en-US.win64.installer.exe"},
+            ],
+            [
+                {"os": "win", "id": "47767", "path": "/firefox/nightly/latest-try/firefox-63.0a1.en-US.win32.installer.exe"},
+                {"os": "linux", "id": "47764", "path": "/firefox/nightly/latest-try/firefox-63.0a1.en-US.win64.installer.exe"},
```
jcristau, line 171: If I'm reading this right we're testing with successive versions here, but we should probably have a test that non-successive versions don't raise?

bouncerscript/tests/test_script.py
```diff
@@ -194,6 +217,13 @@ async def toggled_get_locations_info(*args, **kwargs):
             mocker.patch.object(bscript, "get_locations_info", new=mock)
         await bouncer_locations(locations_context)
 
+        # We don't run this check unless info is being updated.
+        if updated_info:
+            if require_successive_versions and updated_info:
```
jcristau:
```suggestion
            if require_successive_versions:
```
We already know `updated_info` is true-ish here.

bhearsum: Whoops - thanks. I meant to undo that before pushing, heh.

bouncerscript/tests/test_script.py
bhearsum: Right yes, thank you!

bhearsum:
Holding this as a draft until I test it with dev scriptworkers.

bouncerscript/src/bouncerscript/constants.py
```diff
@@ -70,39 +70,39 @@
 
 _BOUNCER_PATH_REGEXES_PER_PRODUCT_DEFAULT = {
     "firefox-nightly-latest": (
-        r"^(/firefox/nightly/latest-mozilla-central-l10n/firefox-\d+\.0a1\.:lang\."
+        r"^(/firefox/nightly/latest-(mozilla-central|try)-l10n/firefox-\d+\.0a1\.:lang\."
```
bhearsum, line 81: This is not ideal, but given that all we do is version bump these, it's probably OK? I could be convinced to move it to the config...

bhearsum:
Try run for this is at https://treeherder.mozilla.org/jobs?repo=try&revision=8df90d3b340d9ec18b1723847f6d89bcd0badf20

bouncerscript/docker.d/worker.yml
```diff
@@ -2,6 +2,10 @@ work_dir: { "$eval": "WORK_DIR" }
 artifact_dir: { "$eval": "ARTIFACTS_DIR" }
 verbose: { "$eval": "VERBOSE == 'true'" }
 taskcluster_scope_prefix: { "$eval": "TASKCLUSTER_SCOPE_PREFIX" }
+require_successive_versions:
+  $if: 'ENV == "prod"'
+  then: true
+  else: false
```
jcristau:
```suggestion
require_successive_versions: { "$eval": "ENV == 'prod'" }
```

jcristau (approved):

PR #872, opened 2023-11-16
feat: Mac hardened signing on signingscript

hneiva:


signingscript/Dockerfile
```diff
@@ -40,6 +40,7 @@ WORKDIR /app
 # Install signingscript + configloader + widevine
 RUN python -m venv /app \
  && cd signingscript \
+ && /app/bin/pip install /app/scriptworker_client \
```
bhearsum, line 38: question: why are we installing scriptworker_client twice? (and its related requirements). If this is not necessary, please remove it from the above block.

signingscript/docker.d/apple_signing_creds.yml
```diff
@@ -0,0 +1,24 @@
...
+      'COT_PRODUCT == "firefox"': 'project:releng:signing:'
+      'COT_PRODUCT == "thunderbird"': 'project:comm:thunderbird:releng:signing:'
+      'COT_PRODUCT == "mozillavpn"': 'project:mozillavpn:releng:signing:'
+      'COT_PRODUCT == "adhoc"': 'project:adhoc:releng:signing:'
+in:
+  $merge:
+    $match:
+      'ENV == "prod" && scope_prefix':
+        '${scope_prefix[0]}cert:release-signing':
+          - "app_credentials": {"$eval": "APPLE_APP_SIGNING_CREDENTIALS"}
```
bhearsum: question: Are these usernames, a username+password, or something else? If they're usernames or other singular values, `app_username` or similar would be better here.

(After reading further down, it looks like these might be base64 encoded pkcs12 bundles? If that's the case, I suggest names like `app_pkcs12_bundle`, and `pkcs12_passphrase`)

signingscript/docker.d/apple_signing_creds.yml
```diff
@@ -0,0 +1,24 @@
...
+            "installer_credentials": {"$eval": "APPLE_INSTALLER_SIGNING_CREDENTIALS"}
+            "password": {"$eval": "APPLE_SIGNING_CREDS_PASSWORD"}
```
bhearsum: suggestion: Similarly, what is this password associated with? (Can it named in a way that makes this clear?)

signingscript/docker.d/build_msix_packaging.sh
```diff
@@ -8,3 +8,8 @@ cd msix-packaging
 ./makelinux.sh --pack
 
 cd ..
+
+cp msix-packaging/.vs/bin/makemsix /usr/bin
+cp msix-packaging/.vs/lib/libmsix.so /usr/lib
+
+rm -rf msix-packaging
```
bhearsum, line 15: praise: thank you for cleaning up the left I mess here :)

signingscript/docker.d/build_rcodesign.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
bhearsum: todo: rename this script to `install_rcodesign.sh` (we are not building it, so let's be clear about that in the name)

signingscript/setup.py
```diff
@@ -5,7 +5,8 @@
 with open(os.path.join(os.path.abspath(os.path.dirname(__file__)), "version.txt")) as f:
     version = f.read().rstrip()
 
-install_requires = ["arrow", "mar", "scriptworker", "taskcluster", "mohawk", "winsign", "macholib"]
+with open(os.path.join(os.path.abspath(os.path.dirname(__file__)), "requirements", "base.in")) as f:
+    install_requires = ["scriptworker_client"] + f.readlines()
```
bhearsum, line 9: praise: thank you for updating/fixing this!

signingscript/src/signingscript/apple.py
```diff
@@ -0,0 +1,42 @@
...
+import os
+from shutil import copy2
+
+from signingscript.exceptions import SigningScriptError
+
+log = logging.getLogger(__name__)
+
+
+PROVISIONING_PROFILE_FILENAMES = {
+    "firefox": "orgmozillafirefox.provisionprofile",
```
bhearsum, line 11: suggestion (non-blocking): this might be better off living in a config somewhere

signingscript/src/signingscript/apple.py
```diff
@@ -0,0 +1,42 @@
...
+    "devedition": "orgmozillafirefoxdeveloperedition.provisionprofile",
+    "nightly": "orgmozillanightly.provisionprofile",
```
bhearsum, line 13: todo: Only one of these files exists. The others should either by added, or commented out for now. (We'll get error messages about `profile_name not allowed` instead of `provisioning profile not found` this way, which I think makes things more clear.)

signingscript/src/signingscript/apple.py
```diff
@@ -0,0 +1,42 @@
...
+            raise SigningScriptError(f"profile_name not allowed: {profile_name}")
+
+        profile_path = os.path.join(os.path.dirname(__file__), "data", profile_name)
+        if not os.path.exists(profile_path):
+            raise SigningScriptError(f"Provisioning profile not found: {profile_name}")
+
+        # Resolve absolute destination path
+        target_abs_path = os.path.join(bundlepath, target_path if target_path[0] != "/" else target_path[1:])
+        if os.path.exists(target_abs_path):
+            log.warning("Provisioning profile at {target_path} already exists, overriding.")
```
bhearsum, line 39: question: Do we know that overriding an existing provisioning profile is always the right thing to do? Are there even any cases where we expect one in a bundle already, or would that be indicative of a problem further up the chain?

Either way, thank you for logging this explicitly.

signingscript/src/signingscript/sign.py
```diff
@@ -1633,3 +1625,71 @@ async def apple_notarize_geckodriver(context, path, *args, **kwargs):
...
+
+    # Get configs from task payload
+    payload = context.task.get("payload", {})
+    hardened_sign_config = payload.get("hardened-sign-config")
+    assert hardened_sign_config and isinstance(hardened_sign_config, list)
+    provisioning_profile_config = payload.get("provisioning-profile-config")
+
+    signed = False
+    for file in os.scandir(signing_dir):
+        if file.is_dir() and file.name.endswith(".app"):
```
bhearsum, line 1658: question: How confident are you about which creds to use based on filenames, or how often might these change? I'm wondering if we should have the payload be selecting which cert to use instead of hardcoding it in the code.

Either way, I think ipa files should be handled now, or soon?

signingscript/src/signingscript/sign.py
```diff
@@ -1633,3 +1625,71 @@ async def apple_notarize_geckodriver(context, path, *args, **kwargs):
...
+            # Developer ID Application certificate
+            creds = context.apple_app_signing_creds_path
+        elif file.is_file() and file.name.endswith(".pkg"):
+            # Use installer credentials
+            creds = context.apple_installer_signing_creds_path
+        else:
+            # If not pkg AND not a directory (.app) - then skip file
```
bhearsum, line 1665: todo: log any files that are skipped to make debugging easier.

signingscript/src/signingscript/sign.py
```diff
@@ -1633,3 +1625,71 @@ async def apple_notarize_geckodriver(context, path, *args, **kwargs):
...
+            continue
+
+        bundle_path = os.path.join(signing_dir, file.path)
+        if provisioning_profile_config:
+            copy_provisioning_profiles(bundle_path, provisioning_profile_config)
+
+        # TODO: widevine and omnija signing should run from formats?
```
bhearsum, line 1673: todo: address this todo :). (I assume this means we should only be doing this if widevine & omnija signing are explicitly requested in the payload? But please correct me if that's wrong...)

signingscript/src/signingscript/utils.py
```diff
@@ -28,11 +28,21 @@ class Autograph:
 @dataclass
 class AppleNotarization:
     """Apple notarization configuration object."""
+
     issuer_id: str
     key_id: str
     private_key: str
 
 
+@dataclass
```
bhearsum, line 37: praise: Nice usage of dataclasses in this patch!

signingscript/src/signingscript/utils.py
```diff
@@ -238,3 +275,22 @@ def split_autograph_format(format_):
...
+    Args:
+        context: The running task context object
+        source (str): the DMG sorce path
+        dest (str): the destination directory to extract files
+    """
+    dmg = context.config["dmg"]
+    hfsplus = context.config["hfsplus"]
+    undmg_cmd = [dmg, "extract", source, "tmp.hfs"]
+    log.info(undmg_cmd)
+    await execute_subprocess(undmg_cmd, cwd=context.config["work_dir"], log_level=logging.DEBUG)
```
bhearsum, line 292: question: Did we ever decide what to do about the symlinks? If we're not handling them here, have you tested to make sure the outputs from `repackage-signing` still contain working symlinks?

bhearsum (changes requested):
(I'm going to try out [Coventional Comments](https://conventionalcomments.org/#labels) for this review.)

I'm marking this as request changes for the few blocking items, questions, and because I'd like a link to one or more test runs of this.

Overall it looks very good though, nice work!

signingscript/Dockerfile
hneiva: I meant to remove it from the block above since it has no effect whatsoever (root user)

signingscript/docker.d/apple_signing_creds.yml
hneiva: Yeah that sounds better.

signingscript/docker.d/apple_signing_creds.yml
hneiva: Renamed.

signingscript/docker.d/build_rcodesign.sh
hneiva: Done

signingscript/src/signingscript/apple.py
hneiva: Added the missing files

signingscript/src/signingscript/apple.py
hneiva: Under normal scenarios it shouldn't happen. But when testing things, this makes sure the profile is always "what it should be".

signingscript/src/signingscript/sign.py
hneiva: If we are not naming the bundles `.app` or not naming installers `.pkg`, I think we have bigger problems.

So far I haven't found a way of signing `ipa` bundles with rcodesign 😢

signingscript/src/signingscript/sign.py
hneiva: Done

signingscript/src/signingscript/sign.py
hneiva: I think this should stay a TODO, and we find a better way of signing them in the future without failing due to the DMG symlink. Currently they are always signed together, so it's safe to assume here in the meantime.

signingscript/src/signingscript/utils.py
hneiva: I was testing this in adhoc signer, which doesn't have repackaging setup.
I guess we could try this on Try?

hneiva (commented):
Updating PR changes on ~~next commit~~ the commit above....?

signingscript/src/signingscript/sign.py
bhearsum: OK, let's file an issue or bug and point at that for the follow-up, in that case.

signingscript/src/signingscript/utils.py
bhearsum: That sounds like a good next step!

signingscript/src/signingscript/sign.py
bhearsum: Did this get filed somewhere?

signingscript/src/signingscript/utils.py
bhearsum: Did this testing happen?

bhearsum (approved):
Approving, but please make sure that the symlink situation is verified before this is used in prod. (You can deploy this is to prod safely, I think, this it's a new format, but it shouldn't be used outside of try or a project branch until we know the symlinks are fine.)

Nice work!

PR #875, opened 2023-11-20
beetmoverscript: push-to-release in parallel to gcs and s3 (bug 1865643)

jcristau:
Attempting to see if that makes things faster.


jcristau:
It seems both the gcs and s3 implementations are actually sync under the hood so this might not help at all...

jcristau:
Closing this. #983 disables push-to-release to s3.

PR #881, opened 2023-11-28
chore: Revert platformdirs pinning + repin all packages

hneiva:
Virtualenv issue has been resolved.


hneiva:
@jcristau I don't know where platformdirs was causing an issue, can you help me verify this patch fixes it?

jcristau:
Oops, I'd pushed this change yesterday but forgot to open the PR.  Now #883, which should be pretty much the same as this.

jcristau (approved):

PR #884, opened 2023-11-28
[treescript] add android_nimbus_update action

gbrownmozilla:
Here's a treescript action to replace the Nimbus workflows on firefox-android:
https://github.com/mozilla-mobile/firefox-android/blob/main/.github/workflows/fenix-update-nimbus-experiments.yml
https://github.com/mozilla-mobile/firefox-android/blob/main/.github/workflows/focus-update-nimbus-experiments.yml

Basically, these periodically update fenix/app/src/main/res/raw/initial_experiments.json and focus-android/app/src/main/res/raw/initial_experiments.json to keep them in sync with https://experimenter.services.mozilla.com/api/v6/experiments-first-run/ . `jq` is used to filter the downloaded json content by app-name before updating the appropriate file.

### DIFF
--- a/treescript/Dockerfile
+++ b/treescript/Dockerfile
@@ -4,7 +4,7 @@ RUN groupadd --gid 10001 app && \
     useradd -g app --uid 10001 --shell /usr/sbin/nologin --create-home --home-dir /app app
 
 RUN apt-get update \
- && apt-get install -y mercurial \
+ && apt-get install -y jq mercurial \
  && apt-get clean \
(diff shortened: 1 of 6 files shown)


jcristau:
Hmm, if memory serves the plan for this was to create a phab patch and have a human review/land it, much like the periodic-updates patches.  Which means it wouldn't have to live in treescript.

gbrownmozilla:
Oops!!

I checked with Charlie Humphreys and they agree: Better to put these updates in phabricator and get human eyes on them before landing.

PR #887, opened 2023-12-01
Bug 1864901 - Add comm-unified repository to treescript

dandarnell:
This is to allow for Thunderbird's new comm-unified repository.

### DIFF
--- a/treescript/docker.d/init_worker.sh
+++ b/treescript/docker.d/init_worker.sh
@@ -40,7 +40,7 @@ case $COT_PRODUCT in
     test_var_set 'SSH_KEY'
     export REPO_TYPE="hg"
     export TRUST_DOMAIN="comm"
-    export UPSTREAM_REPO=""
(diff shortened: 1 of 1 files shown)


dandarnell:
@jcristau Could you review this when you get a chance?

ahal (approved):

ahal (approved):

ahal:
I gave Rob write access to the repo so he can push to `dev-treescript`. I think we should prove this out on dev first as I'm guessing there will be many changes needed for this.

(And btw the `master` and `production*` branches are still protected and only releng can push there)

jfx2006:
I have tested this change locally and it works quite nicely. I will test again on a comm-treescript-dev worker when they are available.

I'm confident with this change: it's use with robustcheckout as an empty value for comm merges originated in #309. Now that comm-unified exists, I believe the "firefoxtree" extension could be dropped from `extensions.rc` and most of that PR reverted. 

While the comm-unified repository is rather new, it's been in use on comm-central for the past few week without any issues.

PR #108, opened 2019-12-06
No bug: remove zip logic from beetmoverscript

MihaiTabara:
Cleanup in beetmoverscript:
* removes all the maven/zip related things with the archive
* remove templates since we now have EOL'ed esr60
* cleanup in README to reflect new GCP documentation

PR #889, opened 2023-12-06
beetmoverscript: remove zip_max_file_in_mb config option

jcristau:
It's been unused since #108.


JohanLorenzo (approved):
I'm the one who added these params and I confirm #108 made them irrelevant anymore. The original intent was to prevent zip bombs. Back then, beetmover was in charge of unzipping an archive before submitting individual files to maven buckets. We wanted to reduce the attack surface by rejecting some zip archives that looked out of shape. 

 Thanks for cleaning this up!

PR #890, opened 2023-12-06
beetmoverscript: remove code to handle private partner uploads (bug 1868354)

jcristau:
I removed the buckets from the config in #772 but left this dead code sitting around.


bhearsum (approved):
Mostly a rubberstamp...thank you!

PR #891, opened 2023-12-06
Bug 1868558 - Add new CoT keys as part of the Dec 2023 rotation

JohanLorenzo:
Depends on https://github.com/mozilla-releng/scriptworker/pull/629


jcristau:
Why are we upgrading anything but scriptworker here?

JohanLorenzo:
Sorry for the confusion. I will upgrade scriptworker in a separate commit after 49.1.0 is published.

jcristau:
I understand that. I'm saying scriptworker is the only thing that needs to change in this PR.

JohanLorenzo:
I totally agree. That's why I'm thinking of doing it in 2 steps. This way, we can revert everything but scriptworker.

JohanLorenzo:
Discussed privately with @jcristau. I had forgotten we upgrade dependencies on Thursdays. Thus, I'll keep this PR to jump bumping scriptworker.

ahal (approved):

PR #892, opened 2023-12-06
ci: upgrade to Taskgraph 7.0.0

ahal:


ahal:
The task failures are due to us adding a call to `git config` in `run-task`. But what I don't understand is why we're getting permission denied here, but the Python image in Balrog works. It uses an identical Dockerfile (other than Python version).

I guess it could have been fixed in the upstream `python:3.11` image?

ahal:
I still don't understand what that permission denied was about, but I noticed that the `pushmsixscript` tox tasks were specifying a custom Docker image that wrapped the base Python one. Only this custom image didn't add anything that the base one didn't already specify. So I deleted it and now the error went away

hneiva (approved):
nice

PR #896, opened 2023-12-11
Bump beetmoverscript pinned dependencies

jcristau:
For some reason `pip` is confused and refusing to install our previous set of dependencies, so roll forward to unblock a deploy.


gbrownmozilla (approved):

jcristau:
Fixed with #898 instead.

PR #897, opened 2023-12-12
ci: fix caching of k8s-image tasks on PRs

jcristau:
Fix the copy of the "resources" attribute from tox tasks to k8s-image, so the latter's cache digest properly depends on the image contents.


jcristau:
The k8s-image-beetmoverscript-python39 failure here matches what we're seeing in production, and this PR's aim is to expose that breakage.  There's a temporary fix in #896, and an attempt at a "real" fix in #898.

PR #898, opened 2023-12-12
beetmover: use newer pip to fix installing our dependencies

jcristau:


jcristau:
This is based on #897 to make sure the k8s-image task runs on the PR.

ahal (approved):

PR #901, opened 2024-01-12
Bug 1871138 - Add 'mozillavpn' to shipitscript

ahal:
This is blocked on updating sops + cloudops-infra


ahal:
sops / cloudops-infra have been updated. Though there aren't any new environment variables being defined, so in retrospect landing this ahead of time probably would have been fine.

jcristau (approved):

PR #906, opened 2024-01-23
Bug 1876041 - [treescript] fix destination path for android-10n-sync

gbrownmozilla:
As seen in https://firefox-ci-tc.services.mozilla.com/tasks/Nr18HjI8TYyH1pnoA6dGog, currently the destination paths for `android-l10n-sync` are incorrect, duplicating part of the destination path by including the path to the .toml file. What was I thinking!?

`android-l10n-import` is working fine - no changes required there.


jcristau (approved):
Can this be covered by a test?

PR #907, opened 2024-01-25
feat(ci)!: Refactor script docker images and upload with skopeo

### LINKED ISSUE #372: switch k8s-image tasks to kaniko

### LINKED ISSUE #837: update k8s-image docker image to use skopeo
We currently use docker-in-docker to build/upload the image, which means we use a super old docker, and a deprecated image manifest schema.

hneiva:
Closes #837 
Closes #372


taskcluster/kinds/push-image/kind.yml
```diff
@@ -0,0 +1,157 @@
...
+        cache-dotcache: false
+        command:
+            - /usr/local/bin/push_image.sh
+    fetches:
+        image:
+            - artifact: image.tar.zst
+              extract: false
+    # TODO: Uncomment to enable pushing
+    # scopes:
+    #     - secrets:get:project/releng/scriptworker-scripts/deploy
```
hneiva: self-reminder - there's more than just this.

taskcluster/kinds/docker-image/kind.yml
```diff
@@ -22,21 +22,57 @@ tasks:
...
+    signingscript:
+        parent: base-image
+    treescript:
+        parent: base-image
+
+    # Testing images
+    pushapkscript-test-py39:
+        definition: base-image-test
+        args:
+            DOCKER_IMAGE_PARENT: python:3.9.7
```
jcristau: Might want to call this BASE_IMAGE to match the above one, and avoid confusion with DOCKER_IMAGE_PARENT that comes from having `parent` in the task definition.

taskcluster/kinds/docker-image/kind.yml
```diff
@@ -22,21 +22,57 @@ tasks:
...
-        definition: signingscript
-        parent: python38
-    signingscript-python39:
-        definition: signingscript
-        parent: python39
+
+    # Used by push-image kind
+    skopeo: {}
+
+    base-image:
```
jcristau: Maybe just `base`?

taskcluster/kinds/push-image/kind.yml
```diff
@@ -0,0 +1,157 @@
...
+    name: "{name}-python{python_version}"
+    worker:
+        taskcluster-proxy: true
+        docker-image: {in-tree: skopeo}
+        max-run-time: 1800
+        env:
+            # TODO: This should probably be set in a transform? either that or remove it because we don't really use it
+            VCS_HEAD_REPOSITORY: scriptworker-scripts
+            VCS_HEAD_REV: main
+            APP_VERSION: 1.0.0
```
jcristau: We should figure out the actual requirements wrt version.json

taskcluster/kinds/push-image/kind.yml
```diff
@@ -0,0 +1,157 @@
...
+    run-on-tasks-for: ["action", "github-pull-request", "github-push"]
+    # run-on-tasks-for: [github-push]
+    # run-on-git-branches:
+    #     - dev
+    #     - production
```
jcristau: Update this before landing.

taskcluster/kinds/push-image/kind.yml
jcristau: push-image also needs to know whether to push to dev or production, which doesn't seem to be handled yet?

hneiva: > We should figure out the actual requirements wrt version.json

I recall having an issue with Jenkins when the file was missing. I could be wrong though

taskcluster/docker/pushapkscript/Dockerfile
```diff
@@ -1,9 +1,30 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
 FROM $DOCKER_IMAGE_PARENT
 
-VOLUME /builds/worker/checkouts
-VOLUME /builds/worker/.cache
+# %include pushapkscript
+ADD --chown=app:app topsrcdir/pushapkscript /app/pushapkscript
```
jcristau, line 8: We need to copy files from `topsrcdir/pushapkscript/docker.d` to `/app/docker.d` on the image (same for the other scripts)

taskcluster/docker/base/Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN python -m venv /app \
+ && cd /app/scriptworker_client \
+ && /app/bin/pip install -r requirements/base.txt \
+ && /app/bin/pip install . \
+ && python -m venv /app/configloader_venv \
+ && cd /app/configloader \
+ && /app/configloader_venv/bin/pip install -r requirements/base.txt \
+ && /app/configloader_venv/bin/pip install .
+
+CMD ["/app/docker.d/init.sh"]
```
jcristau, line 34: Might need to move/copy .dockerignore from the root to each of the script's docker dirs?  Not sure if that works with image_builder.

taskcluster/docker/pushapkscript/Dockerfile
hneiva: Nice find!

hneiva: Should be copied everywhere now.

taskcluster/docker/base/Dockerfile
hneiva: Tested it out, having the .dockerignore file in /app seems to work.

taskcluster/kinds/docker-image/kind.yml
```diff
@@ -22,21 +22,75 @@ tasks:
...
+    # TODO: Figure out PYTHON_REQ_SUFFIX when we decide to have py3.10
+    # base-3.10:
+    #     definition: base
+    #     args:
+    #         BASE_IMAGE: python:3.10
+    addonscript:
+        definition: script
+        parent: base
+        args:
+            SCRIPT_NAME: addonscript
```
ahal, line 41: optional suggestion: These could be set in a transform file that runs before the docker-image transforms.

taskcluster/kinds/docker-image/kind.yml
```diff
@@ -22,21 +22,75 @@ tasks:
...
+    signingscript:
+        parent: base
+    treescript:
+        parent: base
+
+    # Testing images
+    pushapkscript-test-py39:
+        definition: base-test
+        args:
+            BASE_IMAGE: python:3.9.7
```
ahal, line 87: optional suggestion: maybe worth pulling this out into a base image too?

taskcluster/kinds/push-image/kind.yml
ahal: I don't think it's required, but I also think it can be handy to see what sources the image was built from, so doesn't hurt to leave around imo.

taskcluster/kinds/tox/kind.yml
```diff
@@ -95,13 +95,13 @@ tasks:
     pushapkscript:
         python-versions: [39]
         worker:
-            docker-image: {in-tree: 'pushapkscript-python{python_version}'}
+            docker-image: {in-tree: 'pushapkscript-test-py39'}
```
ahal: Why remove this interpolation? I think we should leave it as long as we support the `python-versions` key.

taskcluster/kinds/push-image/kind.yml
hneiva: It should at least be set to the right properties now, matching what we already have (had?)

taskcluster/kinds/push-image/kind.yml
```diff
@@ -0,0 +1,59 @@
...
+    balrogscript: {}
+    beetmoverscript: {}
+    bouncerscript: {}
+    githubscript: {}
+    pushapkscript: {}
+    pushflatpakscript: {}
+    pushmsixscript: {}
+    shipitscript: {}
+    signingscript: {}
+    treescript: {}
```
ahal, line 60: Might be nice to use `from_deps` transforms here, if for no other reason than to have one less place we need to list out all the *scripts. Not a big deal though.

taskcluster/scriptworker_taskgraph/transforms/push_image.py
```diff
@@ -0,0 +1,52 @@
...
+                "VCS_HEAD_REV": head_rev,
+                "DOCKER_REPO": docker_repo,
+                "DOCKER_TAG": docker_tag,
+                "DOCKER_ARCHIVE_TAG": docker_archive_tag,
+            }
+        )
+        push_docker_image = config.params.get("push_docker_image")
+        if push_docker_image:
+            env.update({"PUSH_DOCKER_IMAGE": "1"})
+            scopes.append("secrets:get:project/releng/scriptworker-scripts/deploy")
```
ahal, line 49: I guess this is the existing scope path right? I'd like to change this to use the `trust-domain-scopes` feature, but we should leave that for later.

ahal (approved):
Thanks, this is great! Definitely a much more comprehensive refactor than I was expecting.

taskcluster/kinds/docker-image/kind.yml
```diff
@@ -22,21 +22,75 @@ tasks:
...
+    pushapkscript:
+        parent: base
+    pushflatpakscript:
+        parent: base
+    pushmsixscript:
+        parent: base
+    signingscript:
+        parent: base
+    treescript:
+        parent: base
```
ahal, line 81: Would it be possible to merge the `base` and `script` Dockerfiles? It looks like `script` is just installing the `base.txt` requirements, isn't that needed for these ones with a custom image anyway?

taskcluster/kinds/push-image/kind.yml
hneiva: IMO it's worth keeping it explicit in this case, since it's a very limited list size-wise, with high impact when modified.

taskcluster/scriptworker_taskgraph/transforms/push_image.py
hneiva: I'll file an issue for it.

taskcluster/kinds/docker-image/kind.yml
hneiva: The dockerfile for some of the scripts need a lot of changes+testing before it can use the generic + script installed. I feel like that should come from separate PRs

PR #911, opened 2024-02-07
bug 1879021: Adjust ship it scope prefix for mobile

bhearsum:
I'm looking to simplify and make consistent ship it scope grants in ci-configuration. To support this, I want to put all ship it scopes for GitHub projects behind their `trust_domain`. Currently Firefox Android is the only one not behind this.

### DIFF
--- a/shipitscript/docker.d/init_worker.sh
+++ b/shipitscript/docker.d/init_worker.sh
@@ -11,8 +11,11 @@ test_var_set() {
 }
 
 case $COT_PRODUCT in
-  firefox|mobile)
+  firefox)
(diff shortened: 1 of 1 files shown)


JohanLorenzo (approved):
Discussed with @jcristau. To him, we should do it even though the android repo migration is in progress. Let's do it then! 

cc @gbrownmozilla

bhearsum:
We discussed this and decided against making this adjustment due to the fact that we would also need to adjust https://github.com/mozilla-mobile/firefox-android/blob/main/taskcluster/ci/mark-as-shipped/kind.yml, which is not worth doing prior to the migration into Gecko.

PR #914, opened 2024-02-14
ci: Upgrade Taskgraph to version 7.3.0

hneiva:
- Removes pinned requirements files, since they are now baked into the decision image


taskcluster/kinds/tox/kind.yml
```diff
@@ -9,7 +9,6 @@ kind-dependencies:
 
 transforms:
     - scriptworker_taskgraph.transforms.python_version:transforms
-    - scriptworker_taskgraph.transforms.tox:transforms
```
bhearsum, line 12: Can you explain why this is getting dropped? afaict the code in this transform still runs.

hneiva: I also removed the transform file. At this point it's not doing anything, since taskgraph is automatically adding docker-image as dependency.

bhearsum (approved):

PR #919, opened 2024-02-20
ci: Prevent level 3 decision from running on PRs

hneiva:


hneiva:
Tested with:
`taskgraph full -J --diff -p taskcluster/test/params/production-signingscript-push.yml`
`taskgraph full -J --diff -p taskcluster/test/params/pull-request.yml`

Both had identical task sets

hneiva:
taskcluster_yml_validator seems fine with it as well.
(FWIW So does the decision task that ran on this PR _/shrug_)

gbrownmozilla (approved):

PR #922, opened 2024-02-22
Bug 1876573 - Implement bitrisescript

ahal:


bitrisescript/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
...
+  cli:
+    build: '.'
+    tty: true
+    command: '/bin/bash'
+    volumes:
+      - './:/src'
+    working_dir: '/src'
+    environment:
+      - PATH=/app/bin:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
+
```
hneiva, line 12: Do we really need docker compose files? Is it used anywhere?

bitrisescript/src/bitrisescript/bitrise.py
```diff
@@ -0,0 +1,218 @@
...
+        async with self.request(method, url, **kwargs) as r:
+            log.debug(f"{method_and_url} returned HTTP code {r.status}")
+            response = await r.json()
+
+        log.debug(f"{method_and_url} returned JSON {pformat(response)}")
+
+        return response
+
+
+async def wait_for_build_finish(build_slug: str, poll_interval: int = 60) -> None:
```
hneiva, line 112: +1 Type annotations!

bitrisescript/src/bitrisescript/bitrise.py
```diff
@@ -0,0 +1,218 @@
...
+            break
+        else:
+            log.info("Log is still running. Waiting another minute...")
+            await asyncio.sleep(poll_interval)
+
+    download_url = response["expiring_raw_log_url"]
+    if download_url:
+        await download_file(download_url, os.path.join(artifacts_dir, "bitrise.log"))
+    else:
+        log.error("Bitrise has no log to offer for job {0}. Please check https://app.bitrise.io/build/{0}".format(build_slug))
```
hneiva: Shouldn't it raise an exception at this point if there's no logs? (or are we intentionally ignoring this case)

bitrisescript/src/bitrisescript/bitrise.py
```diff
@@ -0,0 +1,218 @@
...
+
+async def download_file(download_url: str, file_destination: str, chunk_size: int = 512) -> None:
+    """Download a file.
+
+    Args:
+        download_url (str): Url to download.
+        file_destination (str): Path to save the file.
+        chunk_size (int): Size in bytes to stream to disk at a time (default: 512).
+    """
+    async with RetryClient() as s3_client:
```
hneiva: I'm not sure why this variable is called `s3`

bitrisescript/docker-compose.yml
```diff
@@ -0,0 +1,12 @@
+version: '3'
```
bhearsum, line 1: I realize this isn't unique to bitrisescript, but what's the purpose of having a docker-compose file?

bitrisescript/src/bitrisescript/bitrise.py
```diff
@@ -0,0 +1,218 @@
...
+    client = BitriseClient()
+    endpoint = f"/builds/{build_slug}/log"
+
+    while True:
+        response = await client.do_request(endpoint)
+        if response["is_archived"] is True:
+            log.info("Log is now ready")
+            break
+        else:
+            log.info("Log is still running. Waiting another minute...")
```
hneiva: [more info below (above?)] this is an example of log missing some context when dealing with multiple jobs.

bitrisescript/src/bitrisescript/bitrise.py
```diff
@@ -0,0 +1,218 @@
...
+from typing import Any
+
+from aiohttp_retry import RetryClient
+
+from scriptworker_client.exceptions import TaskVerificationError
+from scriptworker_client.utils import get_single_item_from_sequence
+
+BITRISE_API_URL = "https://api.bitrise.io/v0.1"
+
+log = logging.getLogger(__name__)
```
hneiva, line 15: suggestion: should we make the `log` object more context-aware? In case multiple jobs are running at the same time, it would provide us a way of differentiating each log line.

Not required for this initial PR, maybe something to just keep in mind?

bitrisescript/docker.d/init_worker.sh
```diff
@@ -0,0 +1,30 @@
...
+}
+
+case $COT_PRODUCT in
+  mobile)
+    case $ENV in
+      dev|fake-prod)
+        test_var_set 'BITRISE_ACCESS_TOKEN_STAGING'
+        ;;
+      prod)
+        test_var_set 'BITRISE_ACCESS_TOKEN_PROD'
```
bhearsum: I know we name these environment variables differently for stage/prod elsewhere, but seeing this in the context of a new script makes me wonder if we should just use `BITRISE_ACCESS_TOKEN`, and leave it to k8s to deal with setting it appropriately. I'm not sure what value the different names provide (maybe a very small amount of footgun protection?), and they certainly add complexity to this and `worker.yml`.

bitrisescript/src/bitrisescript/bitrise.py
```diff
@@ -0,0 +1,218 @@
...
+import os
+from pprint import pformat
+from typing import Any
+
+from aiohttp_retry import RetryClient
+
+from scriptworker_client.exceptions import TaskVerificationError
+from scriptworker_client.utils import get_single_item_from_sequence
+
+BITRISE_API_URL = "https://api.bitrise.io/v0.1"
```
bhearsum, line 13: As you've probably seen, [Bitrise says](https://devcenter.bitrise.io/en/api.html):
> There’s no long term compatibility promise for v0.1, although we try to do our best to not to break anything unless we have to. Once we’re happy with v0.1 we’ll “freeze” it as v1.0, for which we’ll provide long term support.

With this in mind, we may want to be cautious about building any release workflows on it for the time being.

bitrisescript/tox.ini
```diff
@@ -0,0 +1,50 @@
...
+envlist = docker
+
+[testenv:docker]
+whitelist_externals=docker
+deps =
+usedevelop = false
+depends =
+skip_install = true
+commands =
+    docker build --build-arg PYTHON_VERSION=3.8 --build-arg PYTHON_REQ_SUFFIX=.py38 -t bitrisescript-{envname}-py38-test -f Dockerfile.test .
```
hneiva: I don't think we need 3.8 here

bitrisescript/src/bitrisescript/bitrise.py
```diff
@@ -0,0 +1,218 @@
...
+    async def set_app_prefix(self, app):
+        """Sets the client's prefix to the specified Bitrise app.
+
+        This scopes the client to only the specified application.
+
+        Args:
+            app (str): Name of the Bitrise application to scope to.
+        """
+        apps = (await self.do_request("/apps"))["data"]
+        app = get_single_item_from_sequence(
```
bhearsum: suggestion: use a distinct name for the function argument and this return value. If I understand correctly, `app` starts off as a string when it comes into the function, and then turns into a dict after this returns.

bitrisescript/src/bitrisescript/bitrise.py
```diff
@@ -0,0 +1,218 @@
...
+            sequence=apps,
+            condition=lambda obj: obj["repo_slug"] == app,
+            ErrorClass=TaskVerificationError,
+            no_item_error_message=f"No Bitrise app matching '{app}' found!",
+            too_many_item_error_message=f"More than one Bitrise app matching '{app}' found!",
+        )
+        self.prefix = f"/apps/{app['slug']}"
+
+    async def do_request(self, endpoint: str, method: str = "get", **kwargs: Any) -> dict[str, Any]:
+        """Perform a request against the Bitrise API.
```
bhearsum, line 64: It looks like you need to implement pagination here: https://devcenter.bitrise.io/en/api/pagination-of-api-calls.html

hneiva (approved):
Looks great! 🚀 
Only minor adjustments, so I'll pre-approve.

bitrisescript/src/bitrisescript/bitrise.py
```diff
@@ -0,0 +1,218 @@
...
+                    chunk = await resp.content.read(chunk_size)
+                    if not chunk:
+                        break
+                    fd.write(chunk)
+
+    log.info(f"'{file_destination}' downloaded")
+
+
+async def run_build(artifacts_dir: str, workflow_id: str, **build_params: Any) -> None:
+    """Run the bitrise build corresponding to the specifid build params.
```
bhearsum: typo: specifid

bitrisescript/src/bitrisescript/bitrise.py
ahal: This was copy/pasted from the code in `firefox-ios`. I'm guessing it was because Bitrise is putting its artifacts in S3 storage.. But I agree, that's not a detail we should care about to any degree. I'll update it.

bhearsum (changes requested):
I think this largely looks fine. I'm marking it as Changes Requested due to the need to handle pagination. (If you think that's not needed immediately, I could be convinced.)

I also have a general question/concern: what do we want to do if a bitrisescript worker is terminated after a bitrise build is fired, but before it's completed? Do we want to try to cancel it as part of shutdown? Or do nothing and start a new build? Do we want the retried bitrisescript task to try to find it?

It seems like ideally we'd want to try to find the old bitrise build - but maybe that's a future optimization.

bitrisescript/src/bitrisescript/bitrise.py
ahal: To be honest, I'm not sure. This logic was copy/pasted from existing logic in `firefox-ios`. I'm not sure if the log being missing means there was some fatal error or not. Maybe there's cases where the log doesn't exist but other artifacts do? But that seems unlikely.

I think you're right and it would be safer to just raise an exception here, and if we ever run into a scenario where it's expected to not have a log, we can adjust this at that time.

bitrisescript/docker-compose.yml
ahal: I don't know, but they exist for all the other scripts so I added one here too. It's a reasonable question to ask, but let's leave this in this PR and if we decide to stop providing them, we can remove them all in a future PR.

bitrisescript/src/bitrisescript/bitrise.py
ahal: Hm, I don't know that we have much choice in that matter :/. Building release workflows is why we're triggering these from Taskcluster in the first place.

The good news is that we'll be using this for non-release workflows as well (e.g triggering builds + tests), so if something breaks, odds are it'll surface on-push or pull-request instead of a release graph.

bitrisescript/docker.d/init_worker.sh
ahal: Yeah, I debated this myself. I think it protects against us accidentally copying the wrong token in SOPS. If I mistakenly add `BITRISE_ACCESS_TOKEN_PROD` under the nonprod block, it'll be ignored. Whereas if the name is the same and I do that, we'll potentially have dev workers affecting production.

That said, your comment made me realize something. In Bitrise, API tokens are scoped globally to the org. So in this particular case we do want both level 1 and level 3 workers to use the same token (and sensitive workflows need to be protected by scopes).

We could theoretically still use a "STAGING" token for dev workers, but this would require us to make a new org in Bitrise just for this purpose. So I think you're right and there's no point in having two separate variables.

bitrisescript/src/bitrisescript/bitrise.py
ahal: Actually, on second look the log url expires in 10 minutes, so maybe that's the motivation behind making it an error rather than an exception. Would we really want to hold up a whole graph just because we missed our chance to download the bitrise log? The error message even links to the build where you can presumably still see the bitrise log, so now I'm thinking this shouldn't be fatal.

bitrisescript/docker-compose.yml
ahal: Heitor asked the same thing and.. I don't know! I'm going to add it for now because all the other scripts have it, but yeah.. maybe we can remove them all in a follow-up PR.

bitrisescript/src/bitrisescript/bitrise.py
ahal: Good idea, I attempted to get this working via a logging adapter.. but the global state is shared across async contexts, so I think we'd need to pass the logging object around everywhere. I think I'll just leave it for now..

bitrisescript/src/bitrisescript/bitrise.py
ahal: Thanks, I did add some more context to the log messages themselves. Would be better to use an adapter like you suggest, but this was easier for now.

bitrisescript/README.md
JohanLorenzo, line 1: Improvement for the future: It'd be great if we factorize all the boilerplate files somewhere. It would make the review and the addition of a new `*script` easier. No action needed today.

bitrisescript/src/bitrisescript/bitrise.py
```diff
@@ -0,0 +1,218 @@
...
+        if data["finished_at"]:
+            log.info(f"Build {build_slug} is finished")
+            break
+
+        log.debug(f"Build {build_slug} is still running, waiting {poll_interval}s...")
+        await asyncio.sleep(poll_interval)
+
+    if data["status_text"] != "success":
+        if data["status_text"] == "error":
+            raise Exception(f"Job {build_slug} errored! Got: {response}")
```
JohanLorenzo: Nit: Should we raise a more precise exception here? E.g.: `BitriseError`?

bitrisescript/tests/test_bitrise.py
JohanLorenzo, line 1: Big thank you for adding tests! We had none for https://github.com/mozilla-mobile/firefox-ios/blob/a6349a66ead67a5e09b2d5b804b837a48400ee6b/taskcluster/scripts/bitrise-schedule.py

JohanLorenzo (approved):
Thanks for converting the existing script into a `*script` one! This is much better 👍  This looks good to me. I agree with the points raised by @bhearsum, I leave you two agree on the best approach 🙂

ahal:
I implemented pagination, added a more specific Exception class, switched to a single TOKEN variable and fixed a bug around unclosed aiohttp sessions.

ahal:
> I also have a general question/concern: what do we want to do if a bitrisescript worker is terminated after a bitrise build is fired, but before it's completed? Do we want to try to cancel it as part of shutdown? Or do nothing and start a new build? Do we want the retried bitrisescript task to try to find it?
>
> It seems like ideally we'd want to try to find the old bitrise build - but maybe that's a future optimization.

I missed this question! Good call out.. I'm not sure how we could find an already in-progress build, and I'm assuming we'd only want to find existing builds if there was a spot termination or some other infra error (e.g, not user cancel). So given those two facts, I'd lean towards just always attempting to cancel the build in Bitrise.

Remind me.. do we just need to handle SIGTERM? If it's alright with you, I think I'd like to tackle this in a separate PR as there will be a few layers to this (due to needing to handle SIGTERM and then cancel each future and then have each future cancel the build via Bitrise API). I'll file an issue and tackle it soon.

ahal:
See https://github.com/mozilla-releng/scriptworker-scripts/issues/924 for cancel feature

bhearsum:
> > I also have a general question/concern: what do we want to do if a bitrisescript worker is terminated after a bitrise build is fired, but before it's completed? Do we want to try to cancel it as part of shutdown? Or do nothing and start a new build? Do we want the retried bitrisescript task to try to find it?
> > It seems like ideally we'd want to try to find the old bitrise build - but maybe that's a future optimization.
> 
> I missed this question! Good call out.. I'm not sure how we could find an already in-progress build, and I'm assuming we'd only want to find existing builds if there was a spot termination or some other infra error (e.g, not user cancel). So given those two facts, I'd lean towards just always attempting to cancel the build in Bitrise.

That sounds fine to me. If we get to a point where we're wasting a bunch of credits because of this we can always reconsider.

> Remind me.. do we just need to handle SIGTERM?

Honestly, I don't know...but it seems likely?

> If it's alright with you, I think I'd like to tackle this in a separate PR as there will be a few layers to this (due to needing to handle SIGTERM and then cancel each future and then have each future cancel the build via Bitrise API). I'll file an issue and tackle it soon.

Yup, this is perfectly fine. Just wanted to make sure it was thought about.

bhearsum (approved):

PR #927, opened 2024-02-28
feat(ci)!: Refactor script docker images and upload with skopeo

hneiva:


ahal (approved):
Thanks, lgtm!

taskcluster/scriptworker_taskgraph/transforms/push_image.py
```diff
@@ -0,0 +1,52 @@
...
+                "VCS_HEAD_REV": head_rev,
+                "DOCKER_REPO": docker_repo,
+                "DOCKER_TAG": docker_tag,
+                "DOCKER_ARCHIVE_TAG": docker_archive_tag,
+            }
+        )
+        push_docker_image = config.params.get("push_docker_image")
+        if push_docker_image:
+            env.update({"PUSH_DOCKER_IMAGE": "1"})
+            scopes.append("secrets:get:project/releng/scriptworker-scripts/deploy")
```
ahal: We should put this in the `task-defaults` section in kind.yml instead, so the task always requires this scope. Previously this was used when we wanted to build but not push, but now that these are two separate tasks there's no point

ahal (changes requested):

ahal (approved):

PR #933, opened 2024-03-01
Revert "feat(ci)!: Refactor script docker images and upload with skopeo" for breaking push_image

hneiva:
Reverts mozilla-releng/scriptworker-scripts#932

Reverting since there seems to be a problem with the secret.


hneiva:
Nevermind... it was failing because the bitrise repo doesn't exist 🤦

PR #936, opened 2024-03-05
Bug 1833259 - signing: enable fenix and focus signing on gecko prod workers

jcristau:


gbrownmozilla:
LGTM but needs test updates.

gbrownmozilla (approved):

PR #937, opened 2024-03-05
tests: don't set GPG_PUBKEY in signingscript context

jcristau:
In bug 1832830 it was replaced by GPG_PUBKEY_PATH which is set directly in init_worker.sh.


hneiva (approved):
LGTM

PR #943, opened 2024-03-12
Call pip install with --no-deps

jcristau:
The dependency resolution happens when generating the requirements files, so no need to do that again (and risk breaking things) at install time.

### DIFF
--- a/addonscript/Dockerfile
+++ b/addonscript/Dockerfile
@@ -13,11 +13,11 @@ COPY . /app
 
 RUN python -m venv /app \
  && cd addonscript \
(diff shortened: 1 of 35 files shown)


hneiva (approved):
Have you seen this break things?
Approving, as this seems more like a safeguard/"shouldn't happen, but you never know" _kinda-thing._

jcristau:
It definitely broke stuff in balrog (https://github.com/mozilla-releng/balrog/issues/2647).  I seem to remember something here where we ended up rebuilding requirements files for a similar reason, but I can't recall the details.  You're right it's more of a safeguard than fixing an actual current issue.

PR #946, opened 2024-03-15
Bug 1885493 - signingscript: add autograph_xpi formats, enable them in adhoc-t

jcristau:


signingscript/docker.d/passwords.yml
```diff
@@ -200,6 +200,11 @@ in:
              {"$eval": "AUTOGRAPH_GPG_PASSWORD"},
              ["autograph_gpg"]
             ]
+          - ["https://autograph-external.prod.autograph.services.mozaws.net",
```
JohanLorenzo: I might be wrong. My understanding was that we would talk to the staging instance of autograph. If that's actually the case, we should replace this URL by: 

```suggestion
          - ["https://autograph-external.stage.autograph.services.mozaws.net",
```

jcristau: Quite right!

JohanLorenzo (approved):
This looks great to me. Thanks for adding all the tests. I'm confident this is the right approach

signingscript/src/signingscript/task.py
```diff
@@ -160,6 +160,8 @@ async def sign(context, path, signing_formats, **kwargs):
 
 
 def _get_signing_function_from_format(fmt):
+    if fmt.startswith("autograph_xpi"):
```
willdurand, line 163: for consistency:

```suggestion
    if fmt.startswith("autograph_xpi_"):
```

jcristau: That's on purpose, I wanted the bare `autograph_xpi` to just use the default options (SHA256 / ES256).

willdurand: ha, sorry about that.

PR #948, opened 2024-03-15
fix(ci): Description of push-image kinds

hneiva:


taskcluster/scriptworker_taskgraph/transforms/push_image.py
```diff
@@ -12,6 +12,14 @@
 transforms = TransformSequence()
 
 
+@transforms.add
+def set_task_description(config, tasks):
+    """Set task description"""
+    for task in tasks:
+        task["description"] = f"Push {task['name']} docker image to docker hub"
```
jcristau: Do we maybe want to differentiate dev vs prod here?

jcristau (approved):

PR #951, opened 2024-03-19
bug 1886181: switch dev and -t- xpi signing to cas_cur keyids + autograph stage

bhearsum:
I think this is going to be a temporary measure while we work on content signing root CA succession. After that, I imagine we'll want all of our signing workers (outside of adhoc) to be pointing at prod?

The dev switch is not strictly necessary, but I don't think it's worth the effort of separating the two out for the period of time this will be live - but I could be convinced otherwise.


jcristau (approved):
makes sense to me.

PR #952, opened 2024-03-19
Bug 1886262 - Add autograph_xpi signing to adhoc-signing

ahal:


signingscript/docker.d/passwords.yml
```diff
@@ -200,6 +200,11 @@ in:
              {"$eval": "AUTOGRAPH_GPG_PASSWORD"},
              ["autograph_gpg"]
             ]
+          - ["https://autograph-external.prod.autograph.services.mozaws.net",
+             {"$eval": "AUTOGRAPH_XPI_USERNAME"},
+             {"$eval": "AUTOGRAPH_XPI_PASSWORD"},
+             ["autograph_xpi", "autograph_xpi_sha1_es256_es384", "autograph_xpi_sha1_es256_ps256", "autograph_xpi_sha1_es256", "autograph_xpi_sha1_ps256"],
```
jcristau: If we use the same formats here as in the stage case below, the latter will not be reachable (see `signingscript.sign.get_autograph_config`).  To have autograph prod and stage xpi signing in adhoc-t, they should get distinct sets of formats.

jcristau (changes requested):

signingscript/docker.d/passwords.yml
ahal: Good catch!

It seems like this function should probably factor in the `keyid` right? Otherwise any formats that the `prod` config has will be unreachable in the `stage` config. If the formats are different, then at least there's *a* format we can test on stage, but we might want to test all of them.

jcristau: I'm not sure.  One thing we could do is have `autograph_stage_xpi*` for stage autograph, and `autograph_xpi*` format using prod.  That'd mirror the existing `autograph_stage_mar384` vs `autograph_mar384` formats.

ahal: Yeah, I like that. Nice and simple

ahal:
This ended up needing a fair bit of pre-req work. I can split things off into separate PRs if certain commits end up being more controversial than others.

jcristau:
As far as I can tell "Add 'stage' to xpi formats for configs using autograph stage" + "Add autograph_xpi formats to adhoc trust domain" is all that's needed here?  If so, I'd prefer if the other two commits (around how we treat key ids) were in a separate PR.

ahal:
Sounds good, was thinking of doing this anyway.

It's not quite true though because I'll need to invent another format for the `webextensions_rsa`  keyid, but that's a simple change.

ahal:
Note that I'm not 100% sure whether the autograph credentials we have can access the `webextensions_rsa` keyid or not, so this is a bit of trial and error.

signingscript/docker.d/passwords.yml
```diff
@@ -446,6 +452,18 @@ in:
...
+             {"$eval": "AUTOGRAPH_XPI_PASSWORD"},
+             ["webextension"],
+             "webextensions_rsa"  # unprivileged xpi signing
+            ]
+          - ["https://autograph-external.prod.autograph.services.mozaws.net",
+             {"$eval": "AUTOGRAPH_XPI_USERNAME"},
+             {"$eval": "AUTOGRAPH_XPI_PASSWORD"},
+             ["privileged_webextension"],
+             "extension_rsa"  # privileged xpi signing
+            ]
```
jcristau: Let's be careful not to abuse this, the mozilla-extensions pipeline should be used instead generally.

signingscript/src/signingscript/sign.py
```diff
@@ -966,6 +966,9 @@ def _xpi_signing_options(fmt):
         cose_algorithms = algos.split("_")
         if not cose_algorithms or set(cose_algorithms) - {"PS256", "ES256", "ES384", "ES512"}:
             raise SigningScriptError(f"Unsupported format {fmt}")
+    elif fmt == "webextension":
+        cose_algorithms = ["ES256"]
+        digest = "SHA1"
```
jcristau, line 971: This should presumably switch to sha2 eventually?

jcristau (approved):
Thanks!

signingscript/src/signingscript/sign.py
ahal: I assume so, but this matches what AMO currently does, so willdurand and lgreco wanted to be consistent with that.

ahal:
I [found a way](https://lite.datasette.io/?csv=https://gist.githubusercontent.com/hwine/9949f9e58dd514161cbecf7e6263a01a/raw/prod-signers.csv&csv=https://gist.githubusercontent.com/hwine/9949f9e58dd514161cbecf7e6263a01a/raw/prod-authorizations_edge.csv&csv=https://gist.githubusercontent.com/hwine/9949f9e58dd514161cbecf7e6263a01a/raw/prod-authorizations_app.csv#/data?sql=select%0A++%5Bprod-signers%5D.signer%2C%0A++%5Bprod-authorizations_app%5D.user_app%2C%0A++%5Bprod-signers%5D.type_%2C%0A++%5Bprod-signers%5D.mode%2C%0A++%5Bprod-signers%5D.date_start%2C%0A++%5Bprod-signers%5D.date_end%0Afrom+%5Bprod-signers%5D+join+%5Bprod-authorizations_app%5D+on+%5Bprod-signers%5D.signer+%3D+%5Bprod-authorizations_app%5D.signer%0Aorder+by+%5Bprod-signers%5D.signer+asc%0Alimit+1000&_hide_sql=1) to map our autograph credentials to signers. Good news is we do have creds that can access the `webextensions_rsa` signer (which we use for langpacks). But since it's different than the `extension_rsa` credentials, we'll need to use separate env vars and need to update this PR, SOPS and cloudops-infra.

signingscript/src/signingscript/sign.py
willdurand: Yeah, it'd be better to keep `sha1` for now.

ahal:
Re-requesting review, enough has changed here that it should get a second look.

ahal:
Removed access to the privileged signer for now. We may end up moving that extension into the addons-pipeline.

ahal:
We ultimately decided not to do this through adhoc-signing. I'll make a new PR for the name changes.

PR #953, opened 2024-03-20
Support multiple scope prefixes in bitrisescript

ahal:


bitrisescript/src/bitrisescript/task.py
```diff
@@ -1,33 +1,33 @@
...
 
 
-def validate_scope_prefixes(config, task):
-    """Validates scope prefixes.
+def _get_allowed_scope_prefixes(config):
+    prefixes = config["taskcluster_scope_prefixes"]
+    return [prefix if prefix.endswith(":") else "{}:".format(prefix) for prefix in prefixes]
+
+
+def extract_common_scope_prefix(config: dict[str, Any], task: dict[str, Any]) -> str:
```
hneiva, line 12: are these scope resolution functions generic enough to move to something like scriptworker_client.task?

hneiva (approved):
LGTM

bitrisescript/src/bitrisescript/task.py
ahal: I think so, good call! I copy/pasted it from another script and it looks like a bunch of them have something very similar.. Should be a follow-up though.

PR #955, opened 2024-03-29
Bug 1882852 - remove vendored createprecomplete in iscript and signin…

jcristau:
…gscript

This needed to be kept in sync with the code in mozilla-central, or bad things happen, which bit us in
https://bugzilla.mozilla.org/show_bug.cgi?id=1882322

Instead, directly add remove instructions for the extra signature files we're adding, leaving the rest of the file unchanged.


jcristau:
Tested on -dev: https://firefoxci.taskcluster-artifacts.net/XS13mwJISxS5Wizvr-zMKQ/5/public/logs/precomplete.diff

bhearsum (approved):
I think this is probably OK, but I think there's some risk if a backwards incompatible change is made in the future? Two possible alternatives:
- Pull `createprecomplete` from the correct revision, and run _that_ instead.
- Update/recreate the precomplete file as part of a downstream task (either `repackage-signing` or a new task). Arguably, we shouldn't be creating the precomplete file until after signing anyways. (This is probably more work than its worth at this time, though.)

jcristau:
> I think this is probably OK, but I think there's some risk if a backwards incompatible change is made in the future? Two possible alternatives:
> 
>     * Pull `createprecomplete` from the correct revision, and run _that_ instead.
> 
This doesn't feel right.  The premise of scriptworkers is not to run arbitrary code specified in the payload.

>     * Update/recreate the precomplete file as part of a downstream task (either `repackage-signing` or a new task). Arguably, we shouldn't be creating the precomplete file until after signing anyways. (This is probably more work than its worth at this time, though.)

Ack, maybe.

bhearsum:
> > I think this is probably OK, but I think there's some risk if a backwards incompatible change is made in the future? Two possible alternatives:
> > ```
> > * Pull `createprecomplete` from the correct revision, and run _that_ instead.
> > ```
> 
> This doesn't feel right. The premise of scriptworkers is not to run arbitrary code specified in the payload.
>

Uh...good point! Indeed, this is a bad idea.

> > ```
> > * Update/recreate the precomplete file as part of a downstream task (either `repackage-signing` or a new task). Arguably, we shouldn't be creating the precomplete file until after signing anyways. (This is probably more work than its worth at this time, though.)
> > ```
> 
> Ack, maybe.

Just to be clear, I don't think this needs to happen now if it happens at all. Landing this patch is a good improvement at the very least.

hneiva (approved):
I don't know much about precomplete. Syntax LGTM

signingscript/src/signingscript/sign.py
```diff
@@ -558,16 +557,26 @@ def _get_omnija_signing_files(file_list):
...
+            fh.write(line)
+            instr, path = line.strip().split(None, 1)
+            if instr != "remove":
+                continue
+            if not path or path[0] != '"' or path[-1] != '"':
+                continue
+            file = path[1:-1]
+            if _get_widevine_signing_files([file]):
+                sigfile = _get_mac_sigpath(file)
+                fh.write('remove "{}"\n'.format(sigfile))
```
jcristau, line 578: I'm now worrying that this could add a duplicate remove instruction if the file was already signed in the input archive.

[EDIT: s/not/now/ *sigh*]

jcristau: Updated to deduplicate, with a test.

PR #957, opened 2024-04-02
Bug 1889223 - Submit all files to notarization before polling

hneiva:
This is the first behavior that consumes all files at once, so it's a bit odd how it's being handled in script.py

### DIFF
--- a/signingscript/src/signingscript/script.py
+++ b/signingscript/src/signingscript/script.py
@@ -10,7 +10,7 @@ import aiohttp
 import scriptworker.client
 
 from signingscript.exceptions import SigningScriptError
-from signingscript.task import build_filelist_dict, sign, task_cert_type, task_signing_formats
+from signingscript.task import apple_notarize_stacked, build_filelist_dict, sign, task_cert_type, task_signing_formats
 from signingscript.utils import copy_to_dir, load_apple_notarization_configs, load_autograph_configs
 
 log = logging.getLogger(__name__)
@@ -24,6 +24,7 @@ async def async_main(context):
         context (Context): the signing context.
 
     """
+    work_dir = context.config["work_dir"]
     async with aiohttp.ClientSession() as session:
         all_signing_formats = task_signing_formats(context)
         if "gpg" in all_signing_formats or "autograph_gpg" in all_signing_formats:
@@ -36,7 +37,7 @@ async def async_main(context):
             if not context.config.get("widevine_cert"):
                 raise Exception("Widevine format is enabled, but widevine_cert is not defined")
 
-        if "apple_notarization" in all_signing_formats or "apple_notarization_geckodriver" in all_signing_formats:
+        if {"apple_notarization", "apple_notarization_geckodriver", "apple_notarization_stacked"}.intersection(all_signing_formats):
             if not context.config.get("apple_notarization_configs", False):
                 raise Exception("Apple notarization is enabled but apple_notarization_configs is not defined")
             setup_apple_notarization_credentials(context)
@@ -44,9 +45,11 @@ async def async_main(context):
         context.session = session
         context.autograph_configs = load_autograph_configs(context.config["autograph_configs"])
 
-        work_dir = context.config["work_dir"]
         filelist_dict = build_filelist_dict(context)
         for path, path_dict in filelist_dict.items():
+            if path_dict["formats"] == ["apple_notarization_stacked"]:
+                # Skip if only format is notarization_stacked - handled below
+                continue
             copy_to_dir(path_dict["full_path"], context.config["work_dir"], target=path)
             log.info("signing %s", path)
             output_files = await sign(context, os.path.join(work_dir, path), path_dict["formats"], authenticode_comment=path_dict.get("comment"))
@@ -55,6 +58,15 @@ async def async_main(context):
                 copy_to_dir(os.path.join(work_dir, source), context.config["artifact_dir"], target=source)
             if "gpg" in path_dict["formats"] or "autograph_gpg" in path_dict["formats"]:
                 copy_to_dir(context.config["gpg_pubkey"], context.config["artifact_dir"], target="public/build/KEY")
+
+        # notarization_stacked is a special format that takes in all files at once instead of sequentially like other formats
+        notarization_dict = {path: path_dict for path, path_dict in filelist_dict.items() if "apple_notarization_stacked" in path_dict["formats"]}
+        if notarization_dict:
+            output_files = await apple_notarize_stacked(context, notarization_dict)
+            for source in output_files:
+                source = os.path.relpath(source, work_dir)
+                copy_to_dir(os.path.join(work_dir, source), context.config["artifact_dir"], target=source)
+
     log.info("Done!")
 
 
--- a/signingscript/src/signingscript/script.py
+++ b/signingscript/src/signingscript/script.py
@@ -45,11 +45,16 @@ async def async_main(context):
         context.session = session
         context.autograph_configs = load_autograph_configs(context.config["autograph_configs"])
 
+        # TODO: Make task.sign take in the whole filelist_dict and return a dict of output files.
+        #       That would likely mean changing all behaviors to accept and deal with multiple files at once.
+
         filelist_dict = build_filelist_dict(context)
         for path, path_dict in filelist_dict.items():
             if path_dict["formats"] == ["apple_notarization_stacked"]:
                 # Skip if only format is notarization_stacked - handled below
                 continue
+            if "apple_notarization_stacked" in path_dict["formats"]:
+                raise SigningScriptError("apple_notarization_stacked cannot be mixed with other signing types")
             copy_to_dir(path_dict["full_path"], context.config["work_dir"], target=path)
             log.info("signing %s", path)
             output_files = await sign(context, os.path.join(work_dir, path), path_dict["formats"], authenticode_comment=path_dict.get("comment"))
--- a/signingscript/src/signingscript/script.py
+++ b/signingscript/src/signingscript/script.py
@@ -65,6 +65,7 @@ async def async_main(context):
                 copy_to_dir(context.config["gpg_pubkey"], context.config["artifact_dir"], target="public/build/KEY")
 
         # notarization_stacked is a special format that takes in all files at once instead of sequentially like other formats
+        # Should be fixed in https://github.com/mozilla-releng/scriptworker-scripts/issues/980
         notarization_dict = {path: path_dict for path, path_dict in filelist_dict.items() if "apple_notarization_stacked" in path_dict["formats"]}
         if notarization_dict:
             output_files = await apple_notarize_stacked(context, notarization_dict)
--- a/signingscript/src/signingscript/sign.py
+++ b/signingscript/src/signingscript/sign.py
@@ -1673,11 +1673,11 @@ async def apple_notarize_stacked(context, filelist_dict):
     """
     ATTEMPTS = 5
 
-    # notarization submissions map (path -> submission_id)
-    submissions_map = {}
     relpath_index_map = {}
+    paths_to_notarize = []
     task_index = 0
-    # Submit to notarization one by one
+
+    # Create list of files to be notarized + check for potential problems
     for relpath, path_dict in filelist_dict.items():
         task_index += 1
         relpath_index_map[relpath] = task_index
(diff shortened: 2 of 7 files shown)


signingscript/src/signingscript/script.py
```diff
@@ -35,17 +36,19 @@ async def async_main(context):
...
                 raise Exception("Apple notarization is enabled but apple_notarization_configs is not defined")
             setup_apple_notarization_credentials(context)
 
         context.session = session
         context.autograph_configs = load_autograph_configs(context.config["autograph_configs"])
 
-        work_dir = context.config["work_dir"]
         filelist_dict = build_filelist_dict(context)
         for path, path_dict in filelist_dict.items():
+            if path_dict["formats"] == ["apple_notarization_stacked"]:
```
bhearsum, line 53: Will this actually function correctly if some files are not `apple_notarization_stacked`? Eg: will we sign (or notarize) them correctly, and then move on and do the stacked ones correctly? If not, we should disallow `apple_notarization_stacked` for only some files, and bail at a higher level here.

signingscript/src/signingscript/script.py
```diff
@@ -54,6 +57,15 @@ async def async_main(context):
                 copy_to_dir(os.path.join(work_dir, source), context.config["artifact_dir"], target=source)
             if "gpg" in path_dict["formats"] or "autograph_gpg" in path_dict["formats"]:
                 copy_to_dir(context.config["gpg_pubkey"], context.config["artifact_dir"], target="public/build/KEY")
+
+        # notarization_stacked is a special format that takes in all files at once instead of sequentially like other formats
+        notarization_dict = {path: path_dict for path, path_dict in filelist_dict.items() if "apple_notarization_stacked" in path_dict["formats"]}
```
bhearsum, line 69: Is using a new format for this just to make it safer to roll out? If so, what are the plans for `apple_notarization` and `apple_notarization_geckodriver` after we switch to stacked? (Ideally, it seems like we ought to move everything to stacked.)

signingscript/src/signingscript/sign.py
```diff
@@ -1650,7 +1649,85 @@ async def apple_notarize_geckodriver(context, path, *args, **kwargs):
...
+    for relpath, path_dict in filelist_dict.items():
+        task_index += 1
+        relpath_index_map[relpath] = task_index
+        notarization_workdir = os.path.join(context.config["work_dir"], f"apple_notarize-{task_index}")
+        utils.mkdir(notarization_workdir, delete_before_create=True)
+        _, extension = os.path.splitext(relpath)
+        if extension == ".pkg":
+            path = os.path.join(notarization_workdir, relpath)
+            utils.copy_to_dir(path_dict["full_path"], notarization_workdir, target=relpath)
+            submissions_map[path] = await retry_async(
```
bhearsum: How many files will we be notarization in parallel? We may want to limit how many we do at once. Especially spread over all of the iscript machines, this could end up doing a lot at once, and potentially hitting a rate limit or other problems.

(Note: if we end up making a single request with a big zip file, this is not an issue.)

signingscript/src/signingscript/sign.py
```diff
@@ -1650,7 +1649,85 @@ async def apple_notarize_geckodriver(context, path, *args, **kwargs):
...
+    Submits everything before polling for status.
+    """
+    ATTEMPTS = 5
+
+    # notarization submissions map (path -> submission_id)
+    submissions_map = {}
+    relpath_index_map = {}
+    task_index = 0
+    # Submit to notarization one by one
+    for relpath, path_dict in filelist_dict.items():
```
bhearsum, line 1681: Correct me if I'm wrong, but I seem to recall that we can upload a zipfile with multiple files we want notarized inside. Assuming that's true, I suggest we construct a new zipfile and make a single request rather than making a bunch of smaller ones. (This would avoid potential issues with rate limiting or similar problems.)

signingscript/src/signingscript/sign.py
```diff
@@ -1650,7 +1649,85 @@ async def apple_notarize_geckodriver(context, path, *args, **kwargs):
...
     notarization_workdir = os.path.join(context.config["work_dir"], "apple_notarize")
-    shutil.rmtree(notarization_workdir, ignore_errors=True)
-    utils.mkdir(notarization_workdir)
+    utils.mkdir(notarization_workdir, delete_before_create=True)
 
     return await _notarize_geckodriver(context, path, notarization_workdir)
+
+
+@time_async_function
+async def apple_notarize_stacked(context, filelist_dict):
```
bhearsum, line 1669: There's a lot of overlap between this function and `_notarize_all`. If we're planning to keep the latter, some refactoring to avoid all of the duplication would be nice.

signingscript/src/signingscript/sign.py
```diff
@@ -1650,7 +1649,85 @@ async def apple_notarize_geckodriver(context, path, *args, **kwargs):
...
+                args=(path, context.apple_credentials_path),
+                attempts=ATTEMPTS,
+                retry_exceptions=RCodesignError,
+            )
+        else:
+            await _extract_tarfile(context, path_dict["full_path"], extension, notarization_workdir)
+            workdir_files = os.listdir(notarization_workdir)
+            supported_files = [filename for filename in workdir_files if _can_notarize(filename, (".app", ".pkg"))]
+            if not supported_files:
+                raise SigningScriptError("No supported files found")
```
bhearsum: It would be preferable to do this sort of sanity checking prior to submitting _anything_ - otherwise we're making notarization requests that will never be used.

signingscript/src/signingscript/sign.py
```diff
@@ -1650,7 +1649,85 @@ async def apple_notarize_geckodriver(context, path, *args, **kwargs):
...
+
+    for path in submissions_map.keys():
+        await retry_async(
+            func=rcodesign_staple,
+            args=[path],
+            attempts=ATTEMPTS,
+            retry_exceptions=RCodesignError,
+        )
+
+    # Staple + create tarball where necessary
```
bhearsum: Stapling happens in the block just above this, right? If so, this comment could use an update.

signingscript/src/signingscript/task.py
```diff
@@ -55,6 +61,9 @@
         "autograph_rsa": sign_file_detached,
         "apple_notarization": apple_notarize,
         "apple_notarization_geckodriver": apple_notarize_geckodriver,
+        # This format is handled in script.py
+        # "apple_notarization_stacked": apple_notarize_stacked,
+        "apple_notarization_stacked": noop_sign,
```
bhearsum: This is pretty hacky. My suggestion for an alternative would be to add a layer of indirection: have all of the files passed to _all_ of these handlers, and then have them deal with iterating over the files. This would clean-up the `script.py` changes a lot as well.

signingscript/src/signingscript/sign.py
```diff
@@ -1633,8 +1633,7 @@ async def apple_notarize(context, path, *args, **kwargs):
     """
     # Setup workdir
     notarization_workdir = os.path.join(context.config["work_dir"], "apple_notarize")
-    shutil.rmtree(notarization_workdir, ignore_errors=True)
-    utils.mkdir(notarization_workdir)
+    utils.mkdir(notarization_workdir, delete_before_create=True)
```
bhearsum: This is `delete_before_create` necessary now? Especially for the existing notarization functions?

bhearsum (changes requested):
This overall logic here is fine, but I'd _really_ prefer to see some cleanup happen that avoids the need for having completely separate paths in `script.py` for the stacked notarization and other formats. Doing that would make some of the other comments irrelevant.

I'd also like to see if we can upload a single zip instead of a bunch of smaller requests (see comment further down).

signingscript/src/signingscript/script.py
hneiva: Good point. I'll disallow mixed formats when stacked is present

signingscript/src/signingscript/sign.py
hneiva: We're mostly handling multiple l10n versions of the same app, this should only only happen if something is broken upstream, and likely would be missing for all packages. It's a rare enough case that there's no point in optimizing for.

signingscript/src/signingscript/script.py
hneiva: Yes, I initially replaced `apple_notarization` but decided against it to allow for the feature to ride the train.
I'll add deprecation notices on the old functions.

hneiva: I'll leave geckodriver alone for now, and refactor that as we move things around later.

signingscript/src/signingscript/sign.py
hneiva: That's why I called it `stacked` and not `parallel`.
It's still submitting each file sequentially, but it doesn't start polling for results until all files have been submitted.
That way the max parallel submissions are controlled by our chunking still.
Hopefully this won't hit a limit on apple's end.

signingscript/src/signingscript/sign.py
hneiva: For what I've gathered it's supported - but on the apple side it will check the apps sequentially, making things slower.
With the current setup we're making the exact same amount of parallel requests, so not much of a risk to be rate-limited - unless they could by requests per ~minute(?), which would be weird.

signingscript/src/signingscript/sign.py
hneiva: it's more of a safeguard. I don't see why we wouldn't want that..?

signingscript/src/signingscript/sign.py
bhearsum: Correct me if I'm wrong, but we're still going to have more files in the process of being notarized at the same time, no?

signingscript/src/signingscript/sign.py
bhearsum: I guess what I'm wondering is: why is it getting added? Did you run into something? (I don't really have a problem with it...)

hneiva: To be fair, this is more relevant for local testing to ensure we don't have anything in the workdir. AIUI In CI scriptworker takes care of deleting everything after each task is complete.

signingscript/src/signingscript/script.py
```diff
@@ -36,17 +37,24 @@ async def async_main(context):
...
+        if {"apple_notarization", "apple_notarization_geckodriver", "apple_notarization_stacked"}.intersection(all_signing_formats):
             if not context.config.get("apple_notarization_configs", False):
                 raise Exception("Apple notarization is enabled but apple_notarization_configs is not defined")
             setup_apple_notarization_credentials(context)
 
         context.session = session
         context.autograph_configs = load_autograph_configs(context.config["autograph_configs"])
 
-        work_dir = context.config["work_dir"]
+        # TODO: Make task.sign take in the whole filelist_dict and return a dict of output files.
```
bhearsum, line 48: Let's get an issue on file for this, please.

signingscript/src/signingscript/script.py
bhearsum: Did disallowing mixed formats get implemented?

signingscript/src/signingscript/script.py
bhearsum: Let's get an issue on file to remove this later, too.

signingscript/src/signingscript/sign.py
bhearsum: Just want to highlight this one last time. We're submitted the files synchronously, but we're not waiting on one to finish notarizating before we submit the next one - which means more will be in Apple's system at any given time than before. I have no issue with this, but please keep an eye on notarization tasks when this is activated to see if we start to hit more timeouts or intermittents. If we do, we should add some limits.

signingscript/src/signingscript/task.py
```diff
@@ -56,6 +58,8 @@
         "autograph_rsa": sign_file_detached,
         "apple_notarization": apple_notarize,
         "apple_notarization_geckodriver": apple_notarize_geckodriver,
+        # This format is handled in script.py
+        # "apple_notarization_stacked": apple_notarize_stacked,
```
bhearsum, line 65: We should be able to map this format to a function once the first todo above (about `sign`) is addressed. When you file that issue, please reference it here as well.

bhearsum (commented):
Leaving unapproved for the moment for the question about disallowing mixed formats - I'm OK with what's here though.

bhearsum (approved):
Thanks for filing https://github.com/mozilla-releng/scriptworker-scripts/issues/980!

Please add a comment near the relevant code before merging this.

signingscript/src/signingscript/sign.py
jcristau: I don't think something called mkdir should be doing rmtree.  Since this seems to be a no-op (we were calling rmtree + mkdir, now we're calling mkdir which itself calls rmtree before re-creating the dir), I'd rather this was reverted.

signingscript/src/signingscript/sign.py
jcristau: Ultimately I think this gets us closer to the previous behaviour where one task was submitting files to apple, and then a separate task was polling the service to get the notarized artifact back and staple it?

signingscript/src/signingscript/sign.py
jcristau: I don't think this is about optimization, it's about being defensive, not trusting your inputs, and failing early.

jcristau (changes requested):
Requesting changes for the mkdir bits, I'm fine with the rest of this, thanks.

hneiva:
Removed `delete_before_create`

> Ultimately I think this gets us closer to the previous behaviour where one task was submitting files to apple, and then a separate task was polling the service to get the notarized artifact back and staple it?

At a high level, yes!

> I don't think this is about optimization, it's about being defensive, not trusting your inputs, and failing early.

This feels like we're over-engineering for a red herring case, but I've updated the code to check everything before submitting.

jcristau (approved):

PR #960, opened 2024-04-03
Bug 1889340 - signingscript: remove authenticode sha1 formats

jcristau:


jcristau:
I'll confirm with the vpn team before actually landing this, but might as well get review already.

(vpn was the last project using sha1, and that was fixed in https://bugzilla.mozilla.org/show_bug.cgi?id=1889336 which is in the 2.22.0 release candidate right now)

bhearsum (approved):
Thank you for taking the time to rip this out!

PR #961, opened 2024-04-04
signingscript: update autograph integration test

jcristau:
This updates signingscript/tests/integration/test_autograph.py to pass locally, when run against `docker run --name autograph -d -p 5500:5500 -v $(pwd)/signingscript/tests/data/:/data/ mozilla/autograph:7.1.17 /go/bin/autograph -c /data/autograph_server_test_config.yaml`.

Not adding any plumbing to actually run this automatically or make it easier to run locally though, that is left as a later exercise.


bhearsum (approved):
Thank you!

PR #966, opened 2024-04-10
ci: Add formatting for python and yaml

### LINKED ISSUE #962: formatting checks lost in #666
We added ruff, but AFAICT `ruff check` (the default command) doesn't do any formatting checks - just linting.

cc @hneiva

hneiva:
Closes #962

### DIFF
--- a/taskcluster/kinds/check/kind.yml
+++ b/taskcluster/kinds/check/kind.yml
@@ -21,8 +21,6 @@ task-defaults:
     worker:
         docker-image: {in-tree: 'python39'}
         max-run-time: 1800
-    dependencies:
-        build-docker-image: build-docker-image-python39
 
 tasks:
     ruff:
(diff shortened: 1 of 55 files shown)


taskcluster/kinds/check/kind.yml
```diff
@@ -11,20 +11,22 @@ transforms:
...
+        code-review: true
+    worker-type: b-linux
+    worker:
+        docker-image: {in-tree: 'python39'}
+        max-run-time: 1800
+    dependencies:
+        docker-image: build-docker-image-python39
+
 jobs:
     ruff:
```
bhearsum: nit: it would be nice to give this a suffix (maybe `-lint`?) to distinguish this from `ruff-format`.

bhearsum (approved):
This looks very good to me -- I appreciate you going the extra mile and adding yaml linting!

PR #969, opened 2024-04-15
fix(beetmover) - Bug 1891591 add android-components bucket config

hneiva:
Added same config as mobile prod environment

https://bugzilla.mozilla.org/show_bug.cgi?id=1891591

### DIFF
--- a/beetmoverscript/docker.d/worker.yml
+++ b/beetmoverscript/docker.d/worker.yml
@@ -288,6 +288,7 @@ clouds:
               key: { "$eval": "MAVEN_KEY" }
             product_buckets:
               geckoview: 'maven-default-s3-upload-bucket-1705kkmo2aj3c'
+              components: 'maven-default-s3-upload-bucket-1705kkmo2aj3c'
           maven-nightly-production:
             fail_task_on_error: True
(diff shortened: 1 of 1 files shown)


ahal (commented):
@JohanLorenzo we were hoping you could confirm whether this is the right bucket for these tasks

beetmoverscript/docker.d/worker.yml
```diff
@@ -288,6 +288,7 @@ clouds:
               key: { "$eval": "MAVEN_KEY" }
             product_buckets:
               geckoview: 'maven-default-s3-upload-bucket-1705kkmo2aj3c'
+              components: 'maven-default-s3-upload-bucket-1705kkmo2aj3c'
```
JohanLorenzo, line 291: I confirm this is the bucket we should use. The mobile prod workers pointed to it: 

https://github.com/mozilla-releng/scriptworker-scripts/blob/92fd6ce0d72a32ee2428dfc2781c357e351015a2/beetmoverscript/docker.d/worker.yml#L403

JohanLorenzo (approved):
Thanks guys for taking care of this fix!

PR #974, opened 2024-04-17
Bug 1891815 - Increase bitrise MAX_TASK_TIMEOUT to 2 hours

ahal:
I don't think the comment about json-e not supporting integers is valid anymore. I tested in the playground and integer context values get passed on as integers by $eval.


bhearsum (approved):

ahal:
I ended up needing to revert this.. turns out it does get interpreted as a string and causes a failure in scriptworker.

PR #975, opened 2024-04-17
Bug 1891815 - Increase global task timeout to two hours

ahal:
We have bitrisescript tasks that take longer than an hour and it looks like we currently have no way to set a per script timeout (see the comment about json-e in scriptworker.yml).

So the quick fix here is to bump the global timeout. I believe it's extremely rare (if not impossible) for us to hit this timeout anyway, because Kubernetes has its own timeout (`terminationGracePeriodSeconds`) that is much shorter than this one. The only reason I even knew this timeout exists is because I had set the above to `7200` for bitrisescript. All other scripts have this value set to `3600` or less, which means they should get force killed by k8s long before we ever reach the scriptworker timeout.

### DIFF
--- a/docker.d/init.sh
+++ b/docker.d/init.sh
@@ -113,7 +113,7 @@ if [ "$ENV" == "prod" ]; then
 fi
 export TASK_CONFIG=$CONFIG_DIR/worker.json
 export TASK_LOGS_DIR=$ARTIFACTS_DIR/public/logs
-export TASK_MAX_TIMEOUT=3600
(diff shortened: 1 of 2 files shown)


bhearsum (approved):
This seems fine to try. Maybe it's worth writing a query to track 95th percentile or max runtime by pool? (This is something the old taskclusteretl data is still good for :P)

ahal:
We should fix scriptworker to accept a string value for this and cast it to an int over there. Then we can have per script timeouts.

PR #976, opened 2024-04-18
feat(bitrisescript): scan bitrise.log for PERFHERDER_DATA and dump it…

ahal:
… to stdout

Perfherder is hardcoded to scan `public/live.log` for PERFHERDER_DATA. Since this shows up in `<workflow>/bitrise.log`, we need logic to forward these lines over to the main log.

### DIFF
--- a/bitrisescript/src/bitrisescript/bitrise.py
+++ b/bitrisescript/src/bitrisescript/bitrise.py
@@ -1,6 +1,7 @@
 import asyncio
 import logging
 import os
+from pathlib import Path
 from pprint import pformat
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


bitrisescript/src/bitrisescript/bitrise.py
```diff
@@ -183,6 +184,31 @@ async def download_log(build_slug: str, artifacts_dir: str, poll_interval: int =
...
+    Args:
+        artifacts_dir (str): Path to the artifact directory.
+    """
+    path = Path(artifacts_dir) / "bitrise.log"
+    if not path.is_file():
+        log.warning(f"Not scanning for Perfherder data, {path} does not exist!")
+        return
+
+    perfherder_data_lines = [line for line in path.read_text().splitlines() if line.startswith("PERFHERDER_DATA")]
+    print(perfherder_data_lines)
```
hneiva: Should we use `log.info` instead of `print`, or is this intentional?

hneiva (approved):
LGTM

bitrisescript/src/bitrisescript/bitrise.py
ahal: Uhh.. that is a leftover debug statement I forgot to remove ><

PR #977, opened 2024-04-19
feat(bitrisescript): Add build_variations key to support more variation of workflows in bitrise

### LINKED ISSUE #963: bitrisescript: Support workflow specific environment variables
Bitrisescript has the ability to run multiple workflows from a single task (so we only have one Taskcluster task uselessly polling for completion rather than N). But this feature is severely limited by the fact that the build_params apply globally to all workflows. Notably, environment variables are part of the build_params, so we can only make use of this capability if none of the workflows require any environment variables.

Instead we should provide a mechanism for bitrisescript tasks to apply specific environment variables to specific workflows.

hneiva:
Closes #963

### DIFF
--- a/bitrisescript/src/bitrisescript/task.py
+++ b/bitrisescript/src/bitrisescript/task.py
@@ -1,8 +1,40 @@
 import os
+from copy import deepcopy
 from typing import Any
 
 from scriptworker_client.exceptions import TaskVerificationError
 from scriptworker_client.utils import get_single_item_from_sequence
+
+
+def _deep_merge_dict(source: dict, dest: dict) -> dict:
+    """Deep merge two dictionaries.
+
+    Copied from taskgraph.utils.templates.merge_to
+    https://github.com/taskcluster/taskgraph/blob/main/src/taskgraph/util/templates.py
+
+    Args:
+        source (dict): The base dictionary to be copied from.
+        dest (dict): The destinatioon dictionary modified.
+    """
+    for key, value in source.items():
+        # Override mismatching or empty types
+        if type(value) != type(dest.get(key)):  # noqa: E721
+            dest[key] = source[key]
+            continue
+
+        # Merge dict
+        if isinstance(value, dict):
+            _deep_merge_dict(value, dest[key])
+            continue
+
+        # Merge list
+        if isinstance(value, list):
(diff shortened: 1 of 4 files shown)


bitrisescript/src/bitrisescript/task.py
```diff
@@ -86,16 +102,27 @@ def get_bitrise_workflows(config: dict[str, Any], task: dict[str, Any]) -> list[
...
         task (dict): The task definition.
+        workflow (str): Optional workflow reference used to load workflow_params
 
     Returns:
-        dict: The bitrise build_params to specify. Empty dict if unspecified.
+        dict: The bitrise build_params to specify. Always adds workflow_id to the returned dict.
     """
-    return task["payload"].get("build_params", {})
+    global_params = task["payload"].get("global_params", {})
+    workflow_params = task["payload"].get("workflow_params", {}).get(workflow)
```
ahal, line 132: Idea: What if instead of an object keyed by workflow id, this was just a flat list that included `workflow_id` in the params?

E.g, instead of:
```
workflow_params:
    foo: [{environments: {VAL: 1}, {environments: {VAL: 2}}]
    bar: []
```

We had:
```
workflow_params:
    - {workflow_id: foo, environment: {VAL: 1}}
    - {workflow_id: foo, environment: {VAL: 2}}
    - {workflow_id: bar}
```
?

I feel like this would make the logic a little bit cleaner, though if we do this we'd need to remember to validate the specified workflows against our scopes and error out if we're missing any scopes.

bitrisescript/src/bitrisescript/task.py
```diff
@@ -1,10 +1,26 @@
 import os
+from copy import deepcopy
 from typing import Any
 
 from scriptworker_client.exceptions import TaskVerificationError
 from scriptworker_client.utils import get_single_item_from_sequence
 
 
+def _deep_merge_dict(base: dict, ref: dict) -> dict:
```
ahal: You might just want to copy/paste this function:
https://github.com/taskcluster/taskgraph/blob/main/src/taskgraph/util/templates.py#L40

It's pretty battle tested.

ahal:
Thanks for writing this up, looks great!

ahal (approved):

PR #978, opened 2024-04-22
feat(bitrisescript): Resume job if possible

### LINKED ISSUE #972: bitrisescript: support re-connecting to a Bitrise build
In certain circumstances (e.g unexpected worker shutdown), it would be nice if a task could "re-connect" to a running Bitrise build. Bitrise has an API to list builds, so it should be sufficient to query this and check if a build has already been submitted for the current revision / workflow / build params.

This way we don't trigger a new build in Bitrise every time the Taskcluster task retries.

See also #971 . There are cases where we'll want to abort the build in Bitrise and there are cases we'll want to leave it running so we can re-connect later.

hneiva:
Depends on #977 

Closes #972

### DIFF
--- a/bitrisescript/src/bitrisescript/bitrise.py
+++ b/bitrisescript/src/bitrisescript/bitrise.py
@@ -3,7 +3,7 @@ import logging
 import os
 from pathlib import Path
 from pprint import pformat
-from typing import Any
+from typing import Any, Optional
 
 from aiohttp_retry import RetryClient
 
@@ -42,6 +42,7 @@ class BitriseClient:
             token (str): Bitrise API token.
         """
         self.headers["Authorization"] = token
+        log.info("Authorization header set")
 
     async def set_app_prefix(self, app: str):
         """Sets the client's prefix to the specified Bitrise app.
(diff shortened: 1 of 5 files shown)


bitrisescript/src/bitrisescript/bitrise.py
```diff
@@ -258,11 +276,39 @@ async def run_build(artifacts_dir: str, workflow_id: str, **build_params: Any) -
...
+        workflow_id (str): Identifier of workflow to run.
+        kwargs (Any): Additional parameter to forward to bitrise.
+
+    Returns:
+        The list of running builds
+    """
+    client = BitriseClient()
+    data = {
+        "workflow": workflow_id,  # Note: I don't know why this API uses workflow instead of workflow_id
+        "status": 0,  # not finished (0), successful (1), failed (2), aborted with failure (3), aborted with success (4)
```
ahal, line 296: I think we should also re-use successful builds if possible. E.g
1. Taskcluster task gets killed
2. Bitrise build finishes successfully
3. Taskcluster task is re-run
4. Taskcluster task finds successful build, downloads its artifacts and returns

hneiva: Mmmmm

> Taskcluster task gets killed

How would we differentiate from when we actually want a re-build?

ie: We have a green run on bitrise, but still want another build to be triggered.

ahal: Good point.. I'm not sure

hneiva: For now I think we should only resume if it's currently running on bitrise. That way we can cancel a stuck workflow on the bitrise side if we really need to, and a rerun would fire up a new workflow.

bitrisescript/src/bitrisescript/bitrise.py
```diff
@@ -258,11 +276,40 @@ async def run_build(artifacts_dir: str, workflow_id: str, **build_params: Any) -
...
+
+def find_running_build(running_builds: list[dict], build_params: Any):
+    """Given a list of running builds, find a matching build given build_params
+
+    Args:
+        running_builds (list[dict]): List of running builds
+        build_params (Any): Build parameters to compare
+    """
+    for build in running_builds:
+        if build.get("original_build_params") == build_params:
```
ahal, line 311: Heh, that's convenient.

ahal (approved):
Thanks, lgtm!

PR #979, opened 2024-04-23
Bug 1889299 - add new authenticode signing certificate

jcristau:
This PR:
- adds the renewed authenticode cert and corresponding CA chain
- adds new signing formats using this cert, and enables them for adhoc-signing

Once we're happy with adhoc testing we can either decide to roll out the new formats for the other trust domains (less risk but takes longer to roll out), or simply point the regular `autograph_authenticode_sha2{,_stub}` formats at the new files.


jcristau:
To quote from the commit message:

authenticode_prod_202404.crt contains the signing cert and a couple of
intermediates that chain up to "DigiCert Assured ID Root CA".

```
$ awk '{ print | "openssl x509 -subject -issuer -noout" } /END/ { close("openssl x509 -subject -issuer -noout") }' < signingscript/src/signingscript/data/authenticode_prod_202404.crt
subject=C = US, ST = California, L = San Francisco, O = Mozilla Corporation, OU = Firefox Engineering Operations, CN = Mozilla Corporation
issuer=C = US, O = "DigiCert, Inc.", CN = DigiCert Trusted G4 Code Signing RSA4096 SHA384 2021 CA1
subject=C = US, O = "DigiCert, Inc.", CN = DigiCert Trusted G4 Code Signing RSA4096 SHA384 2021 CA1
issuer=C = US, O = DigiCert Inc, OU = [www.digicert.com](http://www.digicert.com/), CN = DigiCert Trusted Root G4
subject=C = US, O = DigiCert Inc, OU = [www.digicert.com](http://www.digicert.com/), CN = DigiCert Trusted Root G4
issuer=C = US, O = DigiCert Inc, OU = [www.digicert.com](http://www.digicert.com/), CN = DigiCert Assured ID Root CA
```

authenticode_prod_ca_202404.crt contains the "DigiCert Assured ID Root
CA" root (for verification purposes).  Specifically, it was downloaded
from https://cacerts.digicert.com/DigiCertAssuredIDRootCA.crt.pem

In the future we might want to check that all Windows versions we
support trust the newer "DigiCert Trusted Root G4" root, and stop the
chain there.

jcristau:
> In the future we might want to check that all Windows versions we support trust the newer "DigiCert Trusted Root G4" root, and stop the chain there.

[Bug 1769081](https://bugzilla.mozilla.org/show_bug.cgi?id=1769081) is evidence that at least some versions of windows 7 do not trust the G4 root, so for now (and at least until ESR115 is EOL), it's safer to chain to an older root, which is the same we're currently chaining to.

bhearsum (approved):
> Once we're happy with adhoc testing we can either decide to roll out the new formats for the other trust domains (less risk but takes longer to roll out), or simply point the regular autograph_authenticode_sha2{,_stub} formats at the new files.

As annoying as it is, I'd recommend using the new formats everywhere. On the off chance we see any issues from this it should make the cause very, very clear (and in the case of Gecko, sheriffs would be able to revert it).

If you're rolling it out feel free to ignore though :).

jcristau:
> > Once we're happy with adhoc testing we can either decide to roll out the new formats for the other trust domains (less risk but takes longer to roll out), or simply point the regular autograph_authenticode_sha2{,_stub} formats at the new files.
> 
> As annoying as it is, I'd recommend using the new formats everywhere. On the off chance we see any issues from this it should make the cause very, very clear (and in the case of Gecko, sheriffs would be able to revert it).
> 
> If you're rolling it out feel free to ignore though :).

I'll probably start this way (i.e. with the new formats) at least for nightly and maybe beta, but ultimately we'll need to switch away from the current cert and reverting is not an option after June 19th anyway, so when we get closer to that it might be simpler to have everything use the new cert?

jcristau:
Rebased to not be on top of the sha1 removal.

PR #984, opened 2024-04-29
signingscript: add list of signing formats to the README

jcristau:


signingscript/README.md
```diff
@@ -80,6 +80,43 @@ key_ids:
...
+- `autograph_authenticode_sha2_rfc3161_stub`: sign windows binary (PE, MSI, MSIX) using autograph and sha2 hash, adding a dummy certificate in the chain for attribution purposes, and using the rfc3161 protocol for timestamping
+- `autograph_authenticode_202404`: sign windows binary (PE, MSI, MSIX) using autograph and sha2 hash, using the certificate issued 2024-04-02
+- `autograph_authenticode_202404_stub`: sign windows binary (PE, MSI, MSIX) using autograph and sha2 hash, using the certificate issued 2024-04-02, and adding a dummy certificate in the chain for attribution purposes
+- `autograph_authenticode_ev`: sign windows binary using autograph, using the EV (extended validation) code signing certificate, necessary for windows kernel modules
+- `autograph_debsign`: gpg-sign a debian changes file and associated dsc and/or buildinfo, using autograph
+- `autograph_gpg`: get a detached PGP signature for a file, using autograph's data signing endpoint
+- `gpg`: get a detached PGP signature for a file, using autograph's file signing endpoint
```
bhearsum: Is this one still used? (It seems like we should get a bug on file for migrating anything using it to `autograph_gpg` and mark it deprecated...)

bhearsum (approved):
Thank for documenting all of this.

Seeing it laid out like this makes me want to rework the signingscript payload to have things like timestamp protocol, crosssign cert, etc. as actual task metadata rather than it being encoded as part of an opaque "format"...but that is obviously far out of scope for this!

signingscript/README.md
jcristau: The `gpg` format is indeed unused as far as I can tell.

PR #993, opened 2024-05-14
bug 1896651: switch langpack signing to new production AMO certificate

bhearsum:
This is waiting on:
1) Confirmation that this is the right keyid to use
2) The hawk id for these can use this keyid
3) For AMO to switch to this keyid as well (planned for Thursday May 16th @ 9am PT)

### DIFF
--- a/signingscript/docker.d/passwords.yml
+++ b/signingscript/docker.d/passwords.yml
@@ -52,7 +52,8 @@ in:
             - ["https://autograph-external.prod.autograph.services.mozaws.net",
                {"$eval": "AUTOGRAPH_LANGPACK_USERNAME"},
                {"$eval": "AUTOGRAPH_LANGPACK_PASSWORD"},
-               ["autograph_langpack"]
+               ["autograph_langpack"],
+               "webextensions_rsa_dep_202402"
             ]
             - ["https://autograph-external.prod.autograph.services.mozaws.net",
                {"$eval": "AUTOGRAPH_FOCUS_USERNAME"},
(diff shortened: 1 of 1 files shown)


bhearsum:
> Confirmation that this is the right keyid to use

@willdurand - I think we covered this in Slack? Do you want to stamp for this?

> The hawk id for these can use this keyid

As far as we know, it can. We've already seen that, for example, AMO can use its new keyid with its existing hawk creds

> For AMO to switch to this keyid as well (planned for Thursday May 16th @ 9am PT)

We decided on Slack that we may go a bit earlier for langpacks; as long as these go around the same time we're happy with either order.

signingscript/docker.d/passwords.yml
```diff
@@ -52,7 +52,8 @@ in:
             - ["https://autograph-external.prod.autograph.services.mozaws.net",
                {"$eval": "AUTOGRAPH_LANGPACK_USERNAME"},
                {"$eval": "AUTOGRAPH_LANGPACK_PASSWORD"},
-               ["autograph_langpack"]
+               ["autograph_langpack"],
+               "cas_cur_webextensions_rsa"
```
willdurand: I think this should be more like a prod signer:

```suggestion
               "webextensions_rsa_dep_202402"
```

willdurand (approved):

signingscript/docker.d/passwords.yml
willdurand: Ha wait, no.  That won't work.

willdurand (changes requested):
Ha wait, it's more complicated than that.

hneiva (approved):
Preemptive approval assuming keyids are correct

hneiva:
❗ Remember to update the bug title before merging please

willdurand (approved):
Looks good, thanks!

PR #997, opened 2024-05-16
bug 1896651: Reland langpack dep signing change to 202402 keyid

bhearsum:
Requires an autograph deploy before we can merge and deploy this


bhearsum:
Autograph deploy happened yesterday.

jcristau (approved):

PR #998, opened 2024-05-17
signingscript: remove support for deprecated/unused `gpg` format

jcristau:
This was replaced by `autograph_gpg` a long time ago (sending a hash for autograph to sign, instead of the entire payload).


bhearsum (approved):
Thank you for the clean up!

PR #999, opened 2024-05-17
Re-backout langpack dep signing changes

bhearsum:
hawk creds still don't have access to the new keyid


jcristau (approved):

bhearsum:
Autograph deployed a fix instead; no need to back out.

PR #1000, opened 2024-05-17
Bug 1896964 - Cleanup firefox-android repo references

gbrownmozilla:
Remove all references to the firefox-android and staging-firefox-android repos.


githubscript/examples/config.example.json
```diff
@@ -1,15 +1,15 @@
 {
     "work_dir": "/tmp/githubscript/work_dir",
     "github_projects": {
-        "fenix": {
+        "reference-browser": {
           "allowed_actions": ["release"],
           "github_token": "SOME_TOKEN",
           "github_owner": "mozilla-mobile",
-          "github_repo_name": "firefox-android",
+          "github_repo_name": "reference-browser",
```
gbrownmozilla: reference-browser does not actually use githubscript. I've replaced firefox-android with reference-browser in examples and tests just to provide the name of an active github repo; open to suggestions for a better approach!

beetmoverscript/docker.d/worker.yml
```diff
@@ -11,8 +11,6 @@ taskcluster_scope_prefixes:
         - 'project:releng:beetmover:'
       'COT_PRODUCT == "thunderbird"':
         - 'project:comm:thunderbird:releng:beetmover:'
-      'COT_PRODUCT == "mobile"':
-        - 'project:mobile:firefox-android:releng:beetmover:'
```
jcristau, line 15: We should also remove all the other mobile cot-product bits in this file and init_worker.sh.

githubscript/examples/config.example.json
jcristau: Should we remove the mobile bits entirely and use an example from xpi instead?

signingscript/docker.d/passwords.yml
```diff
@@ -78,42 +78,6 @@ in:
...
           ]
-        project:mobile:firefox-android:releng:signing:cert:dep-signing:
-          - ["https://autograph-external.prod.autograph.services.mozaws.net",
-             {"$eval": "AUTOGRAPH_GPG_USERNAME"},
-             {"$eval": "AUTOGRAPH_GPG_PASSWORD"},
-             ["autograph_gpg"]
-          ]
-          - ["https://autograph-external.prod.autograph.services.mozaws.net",
-             {"$eval": "AUTOGRAPH_FOCUS_USERNAME"},
-             {"$eval": "AUTOGRAPH_FOCUS_PASSWORD"},
```
jcristau, line 89: init_worker.sh should be updated to not require variables we're no longer using.

treescript/src/treescript/gecko/versionmanip.py
```diff
@@ -31,9 +31,7 @@
 }
 
 _VERSION_CLASS_PER_END_OF_SOURCE_REPO = {
-    "fenix": MobileVersion,
-    "firefox-android": MobileVersion,
-    "focus-android": MobileVersion,
+    "placeholder-repo": MobileVersion,  # not currently used
```
jcristau: Any reason to not just remove the unused code?

jcristau (changes requested):
Thanks for cleaning this up :)

beetmoverscript/docker.d/worker.yml
gbrownmozilla: Thanks. I think I have those cleaned up now.

signingscript/docker.d/passwords.yml
gbrownmozilla: Done. Also noticed/remembered pushapkscript.

githubscript/examples/config.example.json
gbrownmozilla: Yes!

treescript/src/treescript/gecko/versionmanip.py
gbrownmozilla: Sorry, I'm not sure I understand what you had in mind. I removed that `placeholder-repo` line - it didn't add value. Were you thinking of removing `_VERSION_CLASS_PER_END_OF_SOURCE_REPO`? I've left it thinking that it's generic and might be useful again.
Of course `treescript/github` is also entirely unused now, but I think that's useful in the foreseeable future. Removing that, or cleaning up the github tests further, seems to get a little complicated; I wouldn't mind leaving that for a follow-up.

tests/test_init.py
```diff
@@ -256,6 +256,7 @@ def generate_params():
     xfail = {
         "pushapk-firefox-dev",
         "pushapk-firefox-fake-prod",
+        "pushapk-firefox-prod",
```
jcristau, line 260: Curious what the problem with these is :)

jcristau (approved):
Thanks.

I think we'll want to remove the beetmover mobile pools and the githubscript mobile prod pool from https://github.com/mozilla-services/cloudops-infra/blob/master/projects/relengworker/Jenkinsfile and https://github.com/mozilla-services/cloudops-infra/blob/master/projects/relengworker/Jenkinsfile.dev before deploying this, to avoid failed deploys.

tests/test_init.py
gbrownmozilla: Me too! It wasn't just one problem, of course. Fixed now.

gbrownmozilla: Actually, it seemed fixed locally, but wasn't reliable in all environments. I found *another* problem - pushapk's init_worker relies on keytool - so I restored the xfails and added a comment.

githubscript/examples/task.example.json
```diff
@@ -1,99 +1,73 @@
...
+    "bpEeoIECRSqfjA09GXgUIw",
+    "XP1dJHLKT6ONw7wy7eGqPQ"
   ],
   "requires": "all-completed",
   "routes": [
-    "tc-treeherder.v2.fenix.b94cfdf06be2be4b5a3c83ab4095eb2ecde7ba71.0",
     "checks"
   ],
-  "priority": "highest",
+  "priority": "lowest",
```
hneiva, line 16: heheh this is a good idea :)

hneiva (approved):
LGTM

PR #1001, opened 2024-05-23
Bump dependencies

bhearsum:
Non-test major bumps are mozilla-version, which dropped Python 3.7, and taskcluster, which is traditionally no-risk to take.


bhearsum:
sigh, done against the wrong thing, redoing this

PR #1010, opened 2024-05-30
scriptworker_client: also run tests on python 3.11

jcristau:


jcristau:
This is on top of #1009.

bhearsum (approved):

PR #1013, opened 2024-05-30
fix(treescript): use separate Github apps for prod vs dev

ahal:
Initially I didn't bother splitting these out because I figured they both require the exact same permissions on the repo anyway. So if the dev app secret were compromised, you'd be just as hosed as if the prod app secret were compromised.

However, it didn't occur to me that we might only install the `dev` app on staging repos and that we don't want a bug on treescript-dev to accidentally expose the `prod` app's private key.

To this end, I created a `releng-treescript-dev` app. We should probably avoid installing this app on production repos, and instead only add it to `staging-*` repos. To help with this, I restricted the app to the `mozilla-releng` org.

### DIFF
--- a/treescript/docker.d/worker.yml
+++ b/treescript/docker.d/worker.yml
@@ -27,5 +27,9 @@ hg_ssh_config:
 github_config:
   $if: 'REPO_TYPE == "git"'
   then:
+    app_id:
+      $if: 'ENV == "prod"'
+      then: 390757  # https://github.com/apps/releng-treescript
+      else: 909547  # https://github.com/apps/releng-treescript-dev
     privkey_file: {"$eval": "GITHUB_PRIVKEY_FILE"}
   else: {}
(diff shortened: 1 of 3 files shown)


bhearsum:
I think it's good to separate out the dev and prod credentials for the reasons you outline, but I'm less sure about not using the dev app on production repositories. I guess it seems like it _should_ be avoidable, but I can imagine a case may come up where we want to debug something on a production repo with a dev scriptworker. I guess we could decide that we won't allow that though?

ahal:
Well we can do it if necessary (and maybe remove the app afterwards)..

The difference between `treescript` and every other `*script` here is that the `dev` workers have the same privileges on the repo as the `prod` workers. So e.g, `signingscript-dev` would use a dev cert, but `treescript-dev` would still have write access to the repo, same as `prod`.

So if someone pushes a change to `treescript-dev` that accidentally leaks the private key, all repos with `releng-treescript-dev` installed would be compromised.

bhearsum:
ah, right, thanks for spelling that out. I agree that we probably shouldn't install the dev app on non-dev repos in that case.

bhearsum (approved):

PR #1015, opened 2024-05-31
treescript: add support for fetching l10n revisions in a github repo

bhearsum:
This looks like a bulky patch, but most of it is copy/paste to avoid touching any of the existing l10n bump code paths.

The most notable change in the l10n bumper code is `build_revision_dict_github`, which is vastly simplified compared to its hg compared part. In part this is because we only have one string repo to worry about, but also because @eemeli declared that we don't need to support pinning revisions anymore - so I didn't implement that here. After we cut over to the Github repo and know we aren't going back, we can rip out the hg support.

Most of the rest of this is just wiring in calls to `l10n_bump_github` when appropriate.

This is probably best reviewed commit-by-commit, as the first couple of commits are somewhat unrelated, but necessary, precursors to the core work here (most notably, allowing hg and git to be configured for the same treescript instance).

### DIFF
--- a/treescript/src/treescript/gecko/l10n.py
+++ b/treescript/src/treescript/gecko/l10n.py
@@ -15,8 +15,11 @@ from copy import deepcopy
 
 from scriptworker_client.aio import download_file, retry_async, semaphore_wrapper
 from scriptworker_client.exceptions import DownloadError
+from scriptworker_client.github import extract_github_repo_owner_and_name
 from scriptworker_client.utils import load_json_or_yaml
+from treescript.exceptions import TaskVerificationError, TreeScriptError
 from treescript.gecko import mercurial as vcs
+from treescript.github.client import GithubClient, UnknownBranchError
 from treescript.util.task import CLOSED_TREE_MSG, DONTBUILD_MSG, get_dontbuild, get_ignore_closed_tree, get_l10n_bump_info
 from treescript.util.treestatus import check_treestatus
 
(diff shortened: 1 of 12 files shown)


treescript/src/treescript/gecko/l10n.py
```diff
@@ -232,3 +275,58 @@ async def l10n_bump(config, task, repo_path):
...
+            old_contents = load_json_or_yaml(path, is_path=True)
+            # find latest revision in l10n_repo_url
+            new_contents = await build_revision_dict_github(client, bump_config, repo_path)
+            if old_contents == new_contents:
+                continue
+            with open(path, "w") as fh:
+                fh.write(json.dumps(new_contents, sort_keys=True, indent=4, separators=(",", ": ")))
+            locale_map = build_locale_map(old_contents, new_contents)
+            message = build_commit_message(bump_config["name"], locale_map, dontbuild=dontbuild, ignore_closed_tree=ignore_closed_tree)
+            await vcs.commit(config, repo_path, message)
```
ahal, line 325: Maybe I'm misunderstanding something, but doesn't this only work for Mercurial? For Github I think you need to use the client + GraphQL API to commit. There's no checkout to work with afaict.

There's an example of this here:
https://github.com/mozilla-releng/scriptworker-scripts/blob/da0149e6de61cbe5e6a00723b5b9d4867a61afe2/treescript/src/treescript/github/versionmanip.py#L99

At a high level the above is:
1. Calling `client.get_files(...)` on all files that need to be modified.
2. Computing the diff between the old and new versions of the files.
3. Calling `client.commit` with the diff info.

It's a bit more complicated to implement, but it's kind of nice to not need to deal with cloning or anything.

ahal (changes requested):

treescript/src/treescript/gecko/l10n.py
bhearsum: We're not committing anything to Github here. The new repository is where the translated strings are held, and we need to find the latest revision of them when we run l10n bumper or merge automation, and commit that back to Gecko.

When we migrated Gecko to GitHub I think we'll want to do what you suggest to avoid needing a checkout  for these tasks though?

ahal: Thanks, that's what I was missing

ahal (approved):

PR #1032, opened 2024-06-20
Bump dependencies

jcristau:
Major version bump for numpy (pushapkscript), see
https://numpy.org/doc/stable/release/2.0.0-notes.html


bhearsum (approved):
I dug in a bit. AFAICT we get numpy via matplotlib via androguard. And _that_ is only used to [create a call graph](https://github.com/androguard/androguard/blob/b8ec68bf2ea3d78640a91ad083002f00b6f737f7/androguard/cli/cli.py#L554) in a command we don't use. So this is probably safe.

PR #1033, opened 2024-06-20
Bug 1882556 - Remove .deb hacks to upload ESR .deb packages to the archive

gabrielbusta:
This PR removes the hacky things done to the beetmover script to control which channel .debs would get uploaded to the archive. We want to start publishing these so we need to remove them.

### DIFF
--- a/beetmoverscript/src/beetmoverscript/constants.py
+++ b/beetmoverscript/src/beetmoverscript/constants.py
@@ -103,7 +103,6 @@ RELEASE_EXCLUDE = (
     r"^.*/mar-tools/.*$",
     r"^.*contrib.*",
     r"^.*/beetmover-checksums/.*$",
-    r"^.*\.deb$",
 )
 
 CACHE_CONTROL_MAXAGE = 3600 * 4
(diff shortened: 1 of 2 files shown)


gabrielbusta:
ah but I guess we only want ESR128+ debs :(

bhearsum (commented):
We also have `repackage-deb` tasks for esr115: https://treeherder.mozilla.org/jobs?repo=mozilla-esr115&searchStr=repack%2Cdeb&revision=6b05ad1f5f2dbb0d47ac169115e250ff3776289c

Should they still be excluded there?

bhearsum:
> ah but I guess we only want ESR128+ debs :(

Presumably we could handle this by tweaking https://searchfox.org/mozilla-esr115/rev/60c67692b28cc62964942c64ecbb20ab3430bfae/taskcluster/gecko_taskgraph/manifests/firefox_candidates.yml#227 ?

gabrielbusta:
I will give it a go, but I have a feeling these hacks were introduced because the manifest didn't do the trick :/

gabrielBusta:
Moar hacks #1034

gabrielbusta:
> We also have `repackage-deb` tasks for esr115: https://treeherder.mozilla.org/jobs?repo=mozilla-esr115&searchStr=repack%2Cdeb&revision=6b05ad1f5f2dbb0d47ac169115e250ff3776289c
> 
> Should they still be excluded there?

Yeah, we don't want to archive those.

bhearsum (approved):
I see you removed deb publishing from esr115 in https://phabricator.services.mozilla.com/D214475, which seems like it should work. It doesn't look like `beetmover-repackage` has run since it landed yet though, so you may want to wait until that happens before merging this.

PR #1035, opened 2024-06-26
Bug 1904738 - "hg add" new android-l10n files

gbrownmozilla:
Update android-l10n-import and android-l10n-sync such that new files (resource files which exist in the source android-l10n repo but not in the destination mercurial repo) are "hg add"-ed to the commit.


jcristau (approved):
We could also just call `hg addremove` once before commit (assuming we don't create any temp files in the repo that shouldn't be committed, which hopefully we don't).

But this is fine, thanks!

gbrownmozilla:
Thanks, I strongly considered addremove, but decided I felt slightly better about explicitly adding each file.

PR #1036, opened 2024-06-27
Bug 1904738 - use correct repo_path for "hg add"

gbrownmozilla:
My earlier patch added a call to "hg add" for new files imported from the android-l10n repo; that call to `run_hg_command` includes a `repo_path` parameter, but the value of that path is incorrect.

This PR:
 - for clarity, renames `copy_android_l10n_files()` parameters src_repo_path -> src_path and dest_repo_path -> dest_path
 - adds a new `repo_path` parameter to `copy_android_l10n_files()`
 - passes the correct `repo_path` from the script through to `copy_android_l10n_files()` and `run_hg_command`
 - updates test expectations.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -162,7 +162,7 @@ deps =
     ruff
 commands =
     ruff --version
-    ruff --verbose {toxinidir}
+    ruff check --verbose {toxinidir}
 
 [testenv:ruff-format]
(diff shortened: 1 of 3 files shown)


gabrielbusta (approved):
LGTM

PR #1037, opened 2024-07-03
feat(beetmover): Support expiration tag for GCS artifacts

hneiva:


beetmoverscript/src/beetmoverscript/gcloud.py
```diff
@@ -240,4 +246,10 @@ def move_artifacts(client, bucket_name, blobs_to_copy, candidates_blobs, release
         else:
             log.info("Copying {} to {}".format(source, destination))
             source_blob = bucket.blob(source)
-            bucket.copy_blob(source_blob, bucket, destination, retry=DEFAULT_RETRY)
+            dest_blob = bucket.blob(destination)
+            # We need to set the data payload with some information so the metadata is NOT copied over.
+            # This prevents custom_time metadata from being copied unintentionally
+            # https://cloud.google.com/storage/docs/json_api/v1/objects/rewrite#request-body
+            dest_blob._properties["name"] = destination
```
bhearsum, line 253: Woohoo! I'm glad this ended up working out.

beetmoverscript/tests/conftest.py
```diff
@@ -61,7 +61,6 @@ async def _fake_request(resp_status, method, url, *args, **kwargs):
     return resp
 
 
-@pytest.mark.asyncio
```
bhearsum, line 64: Thanks for the fix-up (these are not tests, so they don't need this mark AIUI).

bhearsum (changes requested):
Nice work here! I know it was a struggle to find a good solution to avoid having `custom_time` set for the files in `releases`.

This patch looks fine, but it needs a bit more in the way of tests. specifically, tests that with some asserts that:
- custom_time is passed when an expiry is present
- custom_time is not passed when an expiry is not present
- Both of these cases mixed together in the same artifact map
- custom_time is not set when moving artifacts

I'd also like to see a staging release run through `push-to-releases` so that we can inspect the state of uploads run with this in a real bucket.

I'm also wondering if we need or want to support this for other schemas? `maven_beetmover_task_schema.json` and  `import-from-gcs-to-artifact-registry` stand out as places where we might want to. (Even if we do, this is probably best handled as a follow-up.)

hneiva:
Yeah adding more tests is a great idea.

I've already tested this in a staging release: [treeherder](https://treeherder.mozilla.org/jobs?repo=try&revision=52e37092acb1be6a4874cfb971a192ee09cdd7c3), [Example task](https://firefox-ci-tc.services.mozilla.com/tasks/UjP6JLDiT963jkXRqjW_JA/runs/3/logs/public/logs/live_backing.log#L22), [gcloud bucket files](https://console.cloud.google.com/storage/browser/moz-fx-productdelivery-no-7d6a-productdelivery/pub/firefox/candidates/130.0b9-candidates/build2?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&project=moz-fx-productdelivery-no-7d6a)

I don't think we have a use case for other schemas yet; but if we do, adding it would be quite easy. Agree we should do it on a follow-up.

bhearsum:
> I've already tested this in a staging release: [treeherder](https://treeherder.mozilla.org/jobs?repo=try&revision=52e37092acb1be6a4874cfb971a192ee09cdd7c3), [Example task](https://firefox-ci-tc.services.mozilla.com/tasks/UjP6JLDiT963jkXRqjW_JA/runs/3/logs/public/logs/live_backing.log#L22), [gcloud bucket files](https://console.cloud.google.com/storage/browser/moz-fx-productdelivery-no-7d6a-productdelivery/pub/firefox/candidates/130.0b9-candidates/build2?pageState=(%22StorageObjectListTable%22:(%22f%22:%22%255B%255D%22))&project=moz-fx-productdelivery-no-7d6a)

Thanks! I will take a look there.

> I don't think we have a use case for other schemas yet; but if we do, adding it would be quite easy. Agree we should do it on a follow-up.

I agree we don't need to do it now unless we have specific plans for cleanup of artifacts that are uploaded through those methods already.

hneiva:
`build/` is part of `.gitignore` - had to force add to make sure the file was added ¯\_(ツ)_/¯

bhearsum (approved):
Thank you for the test improvements! I looked at your staging release as well, and it appears to have done all the right things with the artifacts on ftp.stage.mozaws.net.

PR #1038, opened 2024-07-04
signingscript: remove expired authenticode cert

jcristau:
This switches the autograph_authenticode_sha2{_stub,} formats to use the newer (202404) certificate.

[edit: added missing `_sha2` to the format names]


bhearsum (approved):
> This switches the autograph_authenticode{_stub,} formats to use the newer (202404) certificate.

This looks like it should be a no-opt, as we're using the date-specified formats everywhere I looked (central, esr115, vpn).

This makes me wonder if we should drop the non-dated formats altogether. That would certainly not be something to tackle here though.

PR #1040, opened 2024-07-08
shipitscript: run on python 3.11

jcristau:


bhearsum (approved):
thank you for continuing to run this down!

PR #1041, opened 2024-07-08
signingscript: run on python 3.11 (and debian 12)

jcristau:


signingscript/Dockerfile
```diff
@@ -1,6 +1,6 @@
-# python:3.9.7 docker image contains osslsigncode 2.1.0.
+# python:3.11.9 docker image contains osslsigncode 2.5.
 # if the image changes, verify version of osslsigncode and make sure winsign works as well
```
bhearsum, line 2: I assume you've done some verification of this?

IIRC, last time we took a bump of osslsigncode we neede to make [changes here](https://github.com/mozilla-releng/scriptworker-scripts/pull/436) and [winsign](https://github.com/mozilla-releng/winsign/pull/25).

One thing that would probably be good to manually verify is that attribution still works, which can be done by running an `attribution` task on try, installing with that installer, and seeing if you have attribution data in `about:telemetry`.

jcristau: I used https://treeherder.mozilla.org/jobs?repo=try&revision=190706c21cecc4dbe6311bff48085c1893c97b75&selectedTaskRun=JOF9jjpzSPCQ-MkujV7VkA.0 and about:telemetry says `attribution.dlsource=mozillaci`.  That seems to be the same as an installer from the `attribution` job on central.

bhearsum: Sounds right to me.

bhearsum (approved):

PR #1044, opened 2024-07-11
shipitscript: support creating firefox-android releases (bug 1884740)

jcristau:
Replace long-dead fennec special-case with firefox-android.


jcristau:
This is an alternative to https://github.com/mozilla-releng/shipit/pull/1523; I'm not sure which is least ugly...

bhearsum (approved):

PR #1045, opened 2024-07-12
Bump scriptworker to 60.2.0

bhearsum:
Note this being done against `production` to avoid picking up any other changes with it. This means deploys will happen when it merges.

This bump was done with a local patch to ensure only scriptworker was picked up:
```
diff --git a/maintenance/pin-helper.sh b/maintenance/pin-helper.sh
index 7cc71ef2..ab3d6f7e 100755
--- a/maintenance/pin-helper.sh
+++ b/maintenance/pin-helper.sh
@@ -18,12 +18,12 @@ for dir in $DIRS; do
     if [ "$dir" = "pushflatpakscript" ]; then
         ARGS="$ARGS -g flat-manager"
     fi
     if [ "$dir" = "." ]; then
         ARGS="$EXTRA_ARGS -g test -g docs"
     fi
     echo $ARGS
     pushd "$dir"
-    pip-compile-multi --backtracking -u -o "$SUFFIX" $ARGS
+    pip-compile-multi --backtracking --upgrade-package scriptworker -o "$SUFFIX" $ARGS
     chmod 644 requirements/*.txt
     popd
 done
```


bhearsum:
The ruff check is:
```
[task 2024-07-12T13:05:02.671Z] error: `ruff <path>` has been removed. Use `ruff check <path>` instead.
```

Which is pretty clearly unrelated.

jcristau (approved):
The base.in changes aren't required, I think?

bhearsum:
> The base.in changes aren't required, I think?

Technically not I guess...

hneiva (approved):
LGTM

PR #1047, opened 2024-07-16
chore: upgrade to scriptworker 60.3.0

bhearsum:
This picks up the pointer to the new fxci-config repo for projects.yml, which will ensure chain of trust stays working.

This is targeting `master`. I will push (or graft, in the case of ship it script) this to individual `production` branches to avoid accidental backouts like last time.


jcristau (approved):
I think we should just push master to prod tbh.  AFAICT treescript and beetmoverscript also have changes on their respective production branches that aren't on `production`, so arguably grafting to production is the less safe option.

bhearsum:
> I think we should just push master to prod tbh. AFAICT treescript and beetmoverscript also have changes on their respective production branches that aren't on `production`, so arguably grafting to production is the less safe option.

Ah, okay. I'll just do the usual thing in that case. Thanks.

PR #1049, opened 2024-07-19
Move remaining scripts/tests from python 3.9 to 3.11

jcristau:


README.rst
```diff
@@ -190,7 +190,7 @@ Update python dependencies
 
 ::
 
-  # from scriptworker-scripts/ ; this will run docker for py38 and py39
+  # from scriptworker-scripts/ ; this will run docker for py38 and py311
```
bhearsum, line 193: Is python3.8 still needed because of iscript, or some other reason?

bhearsum (approved):

README.rst
jcristau: iscript is still on 3.8, yes.

hneiva: Sadly 3.8 still.
If the move to Autograph + rcodesign fails for some reason we should look into upgrading the mac signers (os + python versions)

PR #1050, opened 2024-07-19
treescript: run with python 3.11

jcristau:


maintenance/pin.sh
```diff
@@ -37,7 +37,6 @@ PY_38_SCRIPTS=(
 PY_39_SCRIPTS=(
     configloader
     scriptworker_client
```
hneiva, line 39: since treescript is the last script on 39, should we remove 39?

hneiva (approved):
nice!
If we decide to remove configloader and scriptworker_client, I think it should probably be on a different PR..

maintenance/pin.sh
jcristau: That's coming next in #1049. :)

PR #1054, opened 2024-07-23
Bug 1909434 - Convert JSON with json.dumps() instead of with str()

JohanLorenzo:


pushmsixscript/src/pushmsixscript/microsoft_store.py
```diff
@@ -214,11 +215,9 @@ def _update_submission(config, channel, session, submission_request, headers, fi
             submission_request["packageDeliveryOptions"]["packageRollout"]["isPackageRollout"] = True
             submission_request["packageDeliveryOptions"]["packageRollout"]["packageRolloutPercentage"] = config["release_rollout_percentage"]
 
-    # The Store expects all-lower-case 'true' and 'false' in the submission request.
-    submission_request = str(submission_request)
-    submission_request = submission_request.replace("True", "true").replace("False", "false")
+    submission_json_string = json.dumps(submission_request)
     url = _store_url(config, f"{application_id}/submissions/{submission_id}")
-    response = session.put(url, submission_request.encode(encoding), headers=headers)
+    response = session.put(url, submission_json_string.encode(encoding), headers=headers)
```
jcristau:
```suggestion
    response = session.put(url, json=submission_request, headers=headers)
```

jcristau (changes requested):
Can we add a test for this?

pushmsixscript/src/pushmsixscript/microsoft_store.py
```diff
@@ -157,8 +159,30 @@ def _update_submission(config, channel, session, submission_request, headers, fi
...
+        zip_file_name = os.path.join(work_dir, "pushmsix.zip")
+        with zipfile.ZipFile(zip_file_name, "w") as zf:
+            for file_path in file_paths:
+                zf.write(file_path, arcname=upload_file_names[file_path])
+        # Note that simple HTTP uploads fail for large files (like ours!), so
+        # using the BlobClient is required.
+        blob_client = BlobClient.from_blob_url(upload_url)
+        with open(zip_file_name, "rb") as f:
+            d = blob_client.upload_blob(f, blob_type="BlockBlob", logging_enable=False)
+            log.debug(f"upload response: {d}")
```
JohanLorenzo, line 181: I didn't change this block. git believes I added these lines because `_craft_new_submission_request_and_upload_file_names()` reuses many lines above. It turns out it's more lines than this block alone. That's why git made this decision.

JohanLorenzo:
I recommend reviewing each commit individually. Each of them does one thing.

gbrownmozilla (approved):
I appreciate all the new test cases!

I liked jcristau's `json=` suggestion but I think your solution allows for a little better testing; I'm okay with it either way.

jcristau (approved):
nit; s/enconding/encoding/ in a commit message.

PR #1056, opened 2024-07-25
fix(treescript): test for GITHUB_PRIVKEY variable where required

ahal:


jcristau (approved):
Maybe we could move this test to the `if [ "$NEEDS_GIT" == "1" ]` block, instead of each consumer?

PR #1059, opened 2024-08-08
Bug 1912260 - Add Fenix configs for adhoc-signing

gbrownmozilla:


bhearsum (approved):
This is fine, but you'll need to update the 3 adhoc envs in the sops repo before this can be deployed.

PR #1061, opened 2024-08-13
Bump dependencies

gbrownmozilla:


bhearsum (approved):
Major bumps:
* attrs, which we don't use directly
* tenacity, for: "Bump major version to warn API breakage on statistics attribute" - seems harmless
* some sphinx packages I didn't dig into

Looks very safe to take.

PR #1066, opened 2024-09-02
Bug 1858212 - Remove Autograph staging in reference-browser prod signing

JohanLorenzo:
Depends on https://github.com/mozilla-mobile/reference-browser/pull/3056


jcristau (changes requested):
We need to make autograph_apk work again before landing this.

jcristau (approved):
This is good to go after https://github.com/mozilla-mobile/reference-browser/pull/3137

PR #1071, opened 2024-09-10
ci: switch pre-commit autoupdate to monthly schedule

jcristau:


ahal (approved):
Agreed, let's make this cadence standard across all repos.

PR #1072, opened 2024-09-10
chore(deps): Update pinned dependencies

hneiva:


bhearsum (approved):
Only major bump I see is deepmerge, which is for old Python version deprecations.

PR #1073, opened 2024-09-12
signingscript: fix xpi dep-signing keyids (bug 1916503)

jcristau:
The key ids in stage and prod autograph differ, but I missed this when changing the URL in #1068.


jcristau:
Ref: https://gist.github.com/hwine/9949f9e58dd514161cbecf7e6263a01a#file-prod-authorizations_app-csv-L78

bhearsum (approved):

PR #1075, opened 2024-09-24
chore(deps): Update pinned dependencies

hneiva:


bhearsum (approved):
`simple-github` is a major bump, but it looks like only for a change to `client.request` which appears to be unused here.

`taskcluster` is a major bump, but only because the services have caused it AFAICT.

No other major bumps.

ahal (changes requested):
Hang on, the `simple-github` bump almost for sure needs some code changes.

ahal (approved):
Oh sorry, you're right. I forgot treescript only uses the GraphQL endpoint. I wanted to do a quick r- just in case Heitor was about to merge it.

hneiva:
Thanks for double checking! :)

PR #1076, opened 2024-09-26
feat(signingscript): Bug 1710599 - Add support to xz archives

hneiva:


signingscript/src/signingscript/sign.py
```diff
@@ -731,6 +743,19 @@ def _owner_filter(tarinfo_obj):
     return tarinfo_obj
 
 
+def _create_xz_tarfile(to, files, tmp_dir):
```
jcristau: This might not be what we want here.  E.g. hardcoding the x86 bcj filter is almost certainly not ideal for the arm64 binaries.
The compression settings for shipping artifacts almost certainly don't belong in signingscript, so we might want to follow https://bugzilla.mozilla.org/show_bug.cgi?id=1710599#c20 and 1) output tar.zst from signingscript, 2) let the repackage task turn that into whatever ends up shipping.

cc @glandium

jcristau (changes requested):

signingscript/src/signingscript/sign.py
hneiva: I left my thoughts in the bug regarding zstd.

Also good to note, beetmover uploads the tarball from the signing task and not repackaging (at least for linux builds).

jcristau (approved):

PR #1082, opened 2024-10-16
feat(bouncer): Bug 1710599 Add support to xz linux builds

hneiva:


bhearsum:
Is this fully decided already? [This most recent comment in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1710599#c23) suggests that there's still some consideration for zst.

hneiva:
The final package compression is decided on being xz.
The comment is more around using zst in between tasks for better performance compressing/decompressing. I'm treating that as a potential future improvement, since supporting multiple compression formats would require a lot more work.

I'll reply on the thread.

bouncerscript/src/bouncerscript/constants.py
```diff
@@ -66,84 +66,124 @@
...
+        r"mac\.dmg|"
+        r"win(32|64|64-aarch64)\.installer\.exe"
+        r"))$"
     ),
 }
 
 _BOUNCER_PATH_REGEXES_PER_ALTERNATIVE_PACKAGE_FORMAT = {
     **_BOUNCER_PATH_REGEXES_PER_PRODUCT_DEFAULT,
     "firefox-nightly-msi-latest-ssl": (
-        r"^(/firefox/nightly/latest-(mozilla-central|try)/firefox-\d+\.0a1\.en-US\." r"(?:win32\.installer\.msi|win64(?:|-aarch64)\.installer\.msi))$"
```
bhearsum: Please move the changes like this (and many others further down) that are unrelated to xz support to their own PR. This helps reduces noise here, and ensures that a backout of xz support doesn't back this out, and vice versa.

bhearsum (changes requested):
The xz parts look fine here, but please separate the other part out per below.

Also, are beetmoverscript changes needed for this as well (eg: https://github.com/mozilla-releng/scriptworker-scripts/blob/fa61ba745fadd9ce1b1515b09a2deb667805c323/beetmoverscript/src/beetmoverscript/constants.py#L138, https://searchfox.org/mozilla-central/source/taskcluster/gecko_taskgraph/manifests/firefox_candidates.yml, etc.)?

iscript and signingscript also have bz2 references, but I think those are only intermediate packages, and likely don't need any changes.

hneiva:
I've already added what I needed for signingscript. iscript shouldn't need anything, since this is for linux only.
Still need some fixes for beetmover tho, will add that in a separate PR

bhearsum (approved):

PR #1083, opened 2024-10-17
feat(beetmover): Bug 1710599 Add support to xz linux builds

hneiva:


beetmoverscript/src/beetmoverscript/templates/firefox_partner.yml
```diff
@@ -8,6 +8,8 @@ mapping:
   {{ 'en-US' }}:
     target.tar.bz2: firefox-{{ version }}.tar.bz2
     target.tar.bz2.asc: firefox-{{ version }}.tar.bz2.asc
+    target.tar.xz: firefox-{{ version }}.tar.xz
```
bhearsum, line 11: I'm actually not sure if these are used anymore...but this won't hurt. You'll definitely need to update the in tree manifests (which I assume you already know.)

bhearsum (approved):
Have you done / will you do a staging release with this & the bouncer changes for some verification?

hneiva:
> Have you done / will you do a staging release with this & the bouncer changes for some verification?

Yes. I've been just fixing the problems that I found with the current scriptworkers, the plan is to deploy these fixes to dev and run another staging release

PR #1085, opened 2024-10-24
AUT-293: add stage formats for all dev/fake-prod signingscript formats

bhearsum:
This is part 1 of the plan from [this document](https://docs.google.com/document/d/1nOZoNfAg96pPUMU1_NGOoRQD0gQcWgxyMYRnKnOgyJg/edit?tab=t.0). The latest version of this has been tested on dev scriptworkers in https://treeherder.mozilla.org/jobs?repo=try&revision=3856381ba7e5d539896bc5dca31aa7a4df43c7ed. There are a handful of failures that we can ignore:
- MAR signing tasks fail (but they get past the actual autograph communication part) due to not having the real dep MAR key on autograph stage. This will be fixed as part of https://mozilla-hub.atlassian.net/browse/AUT-339, but ought not to block this PR.
- [linux64 signing](https://treeherder.mozilla.org/jobs?repo=try&revision=3856381ba7e5d539896bc5dca31aa7a4df43c7ed&selectedTaskRun=LvDHzEN0Rx2fuC_thXDEgg.0) failed due to a bad keyid, which I fixed and [reran](https://treeherder.mozilla.org/jobs?repo=try&revision=3856381ba7e5d539896bc5dca31aa7a4df43c7ed&selectedTaskRun=LvDHzEN0Rx2fuC_thXDEgg.1) for.
- [part of mac signing failed](https://treeherder.mozilla.org/jobs?repo=try&revision=3856381ba7e5d539896bc5dca31aa7a4df43c7ed&selectedTaskRun=V-_bIkh4R2uQbR8do-gBdw.0) due to me not configuring `stage_autograph_langpack` on the dev mac signing worker. I could, theoretically, do this and rerun, but I honestly don't think it's worth the effort. All of the non ja-jP-mac signing worked fine, because we don't do langpacks for them. I have no reason to believe this will fail in fake-prod or prod.
- [focus signing failed](https://treeherder.mozilla.org/jobs?repo=try&revision=3856381ba7e5d539896bc5dca31aa7a4df43c7ed&selectedTaskRun=BR5HuprLSUKcNbCN53SNzA.0) due to a bad secret. I fixed this and [reran](https://treeherder.mozilla.org/jobs?repo=try&revision=3856381ba7e5d539896bc5dca31aa7a4df43c7ed&selectedTaskRun=BR5HuprLSUKcNbCN53SNzA.1).


bhearsum:
This PR is still in draft while we wait for all of the necessary keys to be populated in autograph stage. (I'll probably also do some sanity checking with scriptworker dev instances before having this reviewed and landed.)

signingscript/src/signingscript/task.py
```diff
@@ -163,10 +163,20 @@ async def sign(context, path, signing_formats, **kwargs):
...
 
 
-def _get_signing_function_from_format(fmt):
+def _get_signing_function_from_format(fmt_and_key_id):
+    if ":" in fmt_and_key_id:
+        fmt, _ = fmt_and_key_id.split(":")
+    else:
+        fmt = fmt_and_key_id
+
     if fmt.startswith("autograph_xpi"):
```
jcristau: Should this handle the `stage_autograph_xpi` prefix as well?

signingscript/src/signingscript/task.py
```diff
@@ -163,10 +163,20 @@ async def sign(context, path, signing_formats, **kwargs):
     return output
 
 
-def _get_signing_function_from_format(fmt):
+def _get_signing_function_from_format(fmt_and_key_id):
+    if ":" in fmt_and_key_id:
+        fmt, _ = fmt_and_key_id.split(":")
+    else:
+        fmt = fmt_and_key_id
```
jcristau: Consider using `signingscript.utils.split_autograph_format`?

jcristau (approved):

signingscript/docker.d/passwords.yml
```diff
@@ -12,12 +12,81 @@ in:
...
+      # strive to keep a 1:1 mapping between hawkids or signers with production.
+      # This is OK; as long as we can sign with the same type of signer as production
+      # it's good enough for stage testing.
       '(ENV == "dev" || ENV == "fake-prod") && (COT_PRODUCT == "firefox" || COT_PRODUCT == "thunderbird")':
         $let:
           firefox_and_thunderbird_nonprod_autograph:
+            # GCP Autograph stage
+            - ["https://stage.autograph.nonprod.webservices.mozgcp.net",
+               "releng_authenticode_dep_sha256_2023",
+               {"$eval": "AUTOGRAPH_STAGE_AUTHENTICODE_SHA2_PASSWORD"},
```
jcristau: I'd like to understand why we're putting the autograph username in here, vs keeping it in sops along with the password.  To me that seems like it'd make rotations harder.

jcristau (changes requested):
Oops, I only looked at the first commit initially.

signingscript/src/signingscript/task.py
bhearsum: Indeed; I think it didn't come up in my testing, but clearly we should have it...

signingscript/docker.d/passwords.yml
bhearsum: This is debatable for sure (I had a thread about in Matrix, but admittedly that might not have been the right place to discuss it...).
 
The truth is that all of: username, password, and keyid are tightly coupled. There's a 1:1 relationship between username and password, and a M:1 between username and keyid. We've had issues in the past where we use the wrong username+keyid combination, and I thought it was preferable to keep those close together to help minimize it. Rotations of both username and password at the same time obviously become more annoying and error prone. Rotations of password only are maybe a little less error prone than that?

Maybe also of note is that ahal and I have been talking a bit about trying to get rid of the vast majority of username and passwords, and switching to at most 1 per trust domain, which would have access to all of the keyids that trust domain needs. At that point, password rotations are much less onerous, and arguably it makes keeping the usernames in this file more logically? (I guess you could still argue it the other way though...)

I'm not necessarily tied to keeping the usernames here, but I'm interested in your thoughts on the above tradeoffs.

jcristau: IMO rotations of password only should not be a thing ever.  They would require simultaneous deployments of autograph and signingscript, which might be acceptable in case of a really bad incident where the password got leaked and is being misused, but can just as easily be remedied by rotating both username and password.

So in my mind, username and password are tightly coupled, but keyid not so much.  If we're moving to a new keyid, we'd first configure it in autograph and deploy that, leaving credentials unchanged, then add it to signingscript and deploy it.  If we're rotating credentials, we'd add new ones to autograph, with access to the same keyids as the old user, and once it's deployed, update sops and deploy signingscript, then remove the old user from autograph (only combining the two autograph changes if the situation is bad enough that downtime is OK).

bhearsum: > IMO rotations of password only should not be a thing ever.

Good point.

I also agree entirely about the processes you've outline above. The flip side of the coin is that from a debugging and reading point of view, the passwords are the least useful piece of data. So, if, eg: you're trying to debug a signingscript error, having the usernames in passwords.yml allows you to reduce the number of places you need to piece together information from. (I'm not sure this is a strong enough reason to override keeping the usernames and passwords together.)

bhearsum: Just to follow-up here: I'm going to move usernames back to sops. I may also make a separate patch to make sure that we log usernames + keyids by default, which will satisfy most concerns I have about ease of debugging.

bhearsum: I've updated this patch to pull usernames from the env again. https://github.com/mozilla-services/cloudops-infra/pull/6093 is open for the helm part, and the `signingscript-autograph-gcp-stage-usernames` branch in the sops repo has that part prepared. Once those two are reviewed, I'll land them and re-test this patch in dev.

jcristau: I think you meant https://github.com/mozilla-services/cloudops-infra/pull/6109 :)

Thanks!

signingscript/docker.d/init_worker.sh
```diff
@@ -81,77 +81,113 @@ case $ENV in
...
         test_var_set 'AUTHENTICODE_CA_PATH'
         test_var_set 'AUTHENTICODE_CA_TIMESTAMP_PATH'
         test_var_set 'AUTHENTICODE_CROSS_CERT_PATH'
         test_var_set 'AUTHENTICODE_TIMESTAMP_STYLE'
+        test_var_set 'GPG_PUBKEY_PATH'
+        test_var_set 'AUTOGRAPH_AUTHENTICODE_SHA2_PASSWORD'
+        test_var_set 'AUTOGRAPH_AUTHENTICODE_SHA2_USERNAME'
         test_var_set 'AUTOGRAPH_GPG_PASSWORD'
         test_var_set 'AUTOGRAPH_GPG_USERNAME'
-        test_var_set 'GPG_PUBKEY_PATH'
```
jcristau: Should GPG_PUBKEY_PATH stay next to the corresponding credentials?

signingscript/docker.d/init_worker.sh
```diff
@@ -81,77 +81,113 @@ case $ENV in
...
+            test_var_set 'AUTOGRAPH_STAGE_FENIX_USERNAME'
+            test_var_set 'AUTOGRAPH_STAGE_FENIX_PASSWORD'
+            test_var_set 'AUTOGRAPH_STAGE_FOCUS_USERNAME'
+            test_var_set 'AUTOGRAPH_STAGE_FOCUS_PASSWORD'
         fi
         ;;
       mobile)
-        test_var_set 'AUTOGRAPH_GPG_PASSWORD'
-        test_var_set 'AUTOGRAPH_GPG_USERNAME'
         test_var_set 'GPG_PUBKEY_PATH'
```
jcristau: Might as well remove GPG_PUBKEY_PATH here if it is unused.

signingscript/docker.d/init_worker.sh
jcristau: Or remove the GPG_PUBKEY_PATH check entirely since it's set in this file, it doesn't come from the deployment env.

signingscript/src/signingscript/task.py
```diff
@@ -163,10 +164,17 @@ async def sign(context, path, signing_formats, **kwargs):
     return output
 
 
-def _get_signing_function_from_format(fmt):
-    if fmt.startswith("autograph_xpi"):
+def _get_signing_function_from_format(fmt_and_key_id):
+    fmt, _ = split_autograph_format(fmt_and_key_id)
+
+    if fmt.startswith("autograph_xpi") or fmt.startswith("stage_autograph_xpi"):
```
jcristau: nit: `fmt.startswith(("autograph_xpi", "stage_autograph_xpi"))`

signingscript/src/signingscript/task.py
```diff
@@ -185,7 +193,7 @@ def _sort_formats(formats):
     """
     # Widevine formats must be after other formats other than macapp; GPG must
     # be last.
-    for fmt in ("widevine", "autograph_widevine", "autograph_omnija", "macapp", "autograph_rsa", "autograph_gpg"):
+    for fmt in ("widevine", "autograph_widevine", "stage_autograph_widevine", "autograph_omnija", "stage_autograph_omnija", "macapp", "autograph_rsa", "autograph_gpg", "stage_autograph_gpg"):
```
jcristau: `stage_autograph_rsa`?  Not sure it's actually necessary, since AIUI it's only used for mozillavpn addons(?), but for consistency's sake.

bhearsum:
https://treeherder.mozilla.org/jobs?repo=try&revision=09c0318c363c2043d61db910ba924de5c2a68e7c&searchStr=sign has my latest tests with this patch + up-to-date sops + cloudops-infra. I've also prepared the `signingscript-fake-prod` branch in the sops repo with the necessary secrets changes we'll need for fake-prod before this lands, as well as some tests to ensure that dev + fake-prod entries (where both exist) have the same autograph secrets.

signingscript/docker.d/passwords.yml
bhearsum: Er, yes. Thank you!

tests/test_init.py
```diff
@@ -83,54 +83,68 @@
...
+        "AUTOGRAPH_STAGE_XPI_PASSWORD": "1",
+        "AUTOGRAPH_STAGE_XPI_USERNAME": "1",
+        "AUTOGRAPH_STAGE_XPI_PRIVILEGED_PASSWORD": "1",
+        "AUTOGRAPH_STAGE_XPI_PRIVILEGED_USERNAME": "1",
         "AUTOGRAPH_WIDEVINE_PASSWORD": "1",
         "AUTOGRAPH_WIDEVINE_USERNAME": "1",
+        "AUTOGRAPH_XPI_PASSWORD": "1",
         "AUTOGRAPH_XPI_PRIVILEGED_PASSWORD": "1",
         "AUTOGRAPH_XPI_PRIVILEGED_USERNAME": "1",
+        "AUTOGRAPH_XPI_USERNAME": "1",
```
jcristau: Keeping username and password next to each other might be better even if it means the list is not quite sorted?

jcristau (approved):

signingscript/src/signingscript/task.py
bhearsum: TIL! Yes, let's do that.

signingscript/src/signingscript/task.py
bhearsum: Ah; I think I originally didn't add this because it wasn't in the hardcoded list of `stage` formats, but now that those are gone it definitely makes sense to update this...

tests/test_init.py
bhearsum: Fair!

bhearsum:
I'm going to fix up the things you mentioned, I also want to note here that I'm looking at a strange failure in [this linux l10n signing job](https://firefox-ci-tc.services.mozilla.com/tasks/APmQ_w0sT36xron2C3b9tw). In just this one chunk, we ended up with:
```
aiohttp.client_exceptions.ClientConnectorDNSError: Cannot connect to host autograph-external.prod.autograph.services.mozaws.net:443 ssl:default [Name or service not known]
```

Normally I'd write this off as network gremlins, but the fact that it's trying to connect to autograph prod is very surprising, as it's configured only to use stage formats. I've pushed an additional change with some logging improvements to help diagnose this, and see if in fact _all_ of the chunks are doing some signing against prod.

bhearsum:
> I'm going to fix up the things you mentioned, I also want to note here that I'm looking at a strange failure in [this linux l10n signing job](https://firefox-ci-tc.services.mozilla.com/tasks/APmQ_w0sT36xron2C3b9tw). In just this one chunk, we ended up with:
> 
> ```
> aiohttp.client_exceptions.ClientConnectorDNSError: Cannot connect to host autograph-external.prod.autograph.services.mozaws.net:443 ssl:default [Name or service not known]
> ```
> 
> Normally I'd write this off as network gremlins, but the fact that it's trying to connect to autograph prod is very surprising, as it's configured only to use stage formats. I've pushed an additional change with some logging improvements to help diagnose this, and see if in fact _all_ of the chunks are doing some signing against prod.

This turned out to be another case of a format being hardcoded deep in the bowels of signingscript. I've fixed it, and found a bunch more along the way. I've also added some logging on the success cases. I'm kicking off some widespread testing again, and I'll verify that a) things continue to work, and b) everything is actually using stage when it should.

Regardless, this will need review again, as the changes I've made are not trivial.

bhearsum:
My [latest try push](https://treeherder.mozilla.org/jobs?repo=try&revision=1ad6ddd6d1002d80330a299fbd867cb0807cc1cd&searchStr=sign) is still finishing up, but it's already done every type of signing AFAICT, and I'm quite confident it will turn out fine. (The scattered failures are DNS resolution issues, which I've reported back to the Autograph team about.)

signingscript/src/signingscript/sign.py
```diff
@@ -537,22 +537,24 @@ def _get_mac_sigpath(from_):
...
 # _get_widevine_signing_files {{{1
 def _get_widevine_signing_files(file_list):
-    """Return a dict of path:signing_format for each path to be signed."""
+    """Return a dict of path:is_blessed for each path to be signed."""
     files = {}
     for filename in file_list:
-        fmt = None
         base_filename = os.path.basename(filename)
+        if base_filename not in _WIDEVINE_BLESSED_FILENAMES and base_filename not in _WIDEVINE_NONBLESSED_FILENAMES:
+            log.debug(f"{filename} is not a widevine signing file")
```
jcristau: That is going to be quite noisy...

signingscript/src/signingscript/sign.py
```diff
@@ -537,22 +537,24 @@ def _get_mac_sigpath(from_):
...
-        if fmt:
-            log.debug("Found {} to sign {}".format(filename, fmt))
-            sigpath = _get_mac_sigpath(filename)
-            if sigpath not in file_list:
-                files[filename] = fmt
-            else:
-                log.debug("{} is already signed! Skipping...".format(filename))
+            blessed = True
+
+        log.debug("Found {} to sign {}".format(filename, blessed))
```
jcristau: "Found firefox/firefox-bin to sign False" is a bit of an odd message

jcristau (approved):
LGTM, just a couple of logging nits.

jcristau:
Another failure on the try push is https://firefox-ci-tc.services.mozilla.com/tasks/FIpZBU_ZSUqWZ5KbKncx1g, where iscript doesn't know about stage_autograph_langpack; probably something to handle (or not) separately from this PR.

signingscript/src/signingscript/sign.py
bhearsum: You're right, this could be improved.

signingscript/src/signingscript/sign.py
bhearsum: Yeah, that's fair. It's probably not necessary to add this in the end; we already log the other path.

bhearsum:
> Another failure on the try push is https://firefox-ci-tc.services.mozilla.com/tasks/FIpZBU_ZSUqWZ5KbKncx1g, where iscript doesn't know about stage_autograph_langpack; probably something to handle (or not) separately from this PR.

Yeah. I did some best effort testing of stage with iscript, but that configuration and set-up is a whole other beast that I'm not prepared to tackle at the moment. (I did do enough testing and set-up to verify that iscript machines can route to the new autograph envs, and that the other types of signing they do with autograph works.)

PR #1086, opened 2024-10-24
fix(signingscript): remove formats no longer referenced in project repositories

bhearsum:
autograph_authenticode_sha2 was used by now expired authenticode certs autograph_mar384 was used for non-hash mar signing, and is fully deprecated by hash only signing, and no longer even used in an ESR repo


hneiva (approved):
LGTM

PR #1087, opened 2024-10-24
fix(signingscript): update README

bhearsum:
* Remove the formats that were eliminated in #1086
* Replace the contents of the Signers & Keys section with a pointer to the sources of truth, as this section is virtually impossible to keep up to date


jcristau (approved):
Perfect, thank you!

PR #1091, opened 2024-11-19
chore: dependency upgrades

bhearsum:
Some major bumps:
- taskcluster (generally safe)
- sphinx (docs only)
- pytest-cov (tests only)
- pillow
  - this appears to be about dropping python 3.8
- async-timeout
  - changelog says: `Drop deprecated sync context manager support, use async with timeout(...): ... instead. (https://github.com/aio-libs/async-timeout/issues/421)`
- alabaster (sphinx theme, docs only)


bhearsum:
afaict we're always using `async with timeout` already, so the async-timeout bump should be safe

jcristau (approved):
rs=me

PR #1093, opened 2024-11-22
fix: pick up new version of flat-manager-client

bhearsum:
This is to grab https://github.com/flatpak/flat-manager/pull/139, which ought to fix the bustage we saw in https://bugzilla.mozilla.org/show_bug.cgi?id=1932766 caused by an upgrade to aiohttp.


jcristau (changes requested):
1) taskcluster/docker/pushflatpakscript/Dockerfile needs to be updated, this one is only for local testing at this point iirc
2) the sha256 hash needs to be updated to match

jcristau (approved):

PR #1098, opened 2024-12-03
Bump dependencies

jcristau:


jcristau:
Major bumps: 
- asttokens 2.4.1 → 3.0.0: removed python 3.8 support; this is only pulled in by pushapkscript, which uses 3.11
- async-timeout 4.0.3 → 5.0.1: already done in #1091, we're just re-landing the bump for pushflatpakscript after #1092
- cryptography 43.0.3 → 44.0.0: dropped support for libressl < 3.9; we don't use that.
- taskcluster 74.0.1 → 75.0.0: the major change is in generic-worker, not the python client library

bhearsum (approved):

PR #1099, opened 2024-12-03
signingscript: autograph gcp migration step 2: test against gcp prod

bhearsum:
This patch adds another set of new formats that point at Autograph GCP prod. These entries contain equivalents for all current production formats, and use the exact same credentials the existing production formats. Where they differ are:
* Different formats (so we can opt into them)
* Different autograph URL
* Ensure we use explicit keyids everywhere

Note that this patch landing and being deployed should _not_ end up changing how we sign anything on its own, it will merely make gcp prod available to dev and fake-prod workers.

### DIFF
--- a/signingscript/docker.d/passwords.yml
+++ b/signingscript/docker.d/passwords.yml
@@ -81,6 +81,56 @@ in:
                "dummyapp_android",
             ]
 
+            # GCP Autograph prod
+            - ["https://prod.autograph.prod.webservices.mozgcp.net",
+               {"$eval": "AUTOGRAPH_AUTHENTICODE_SHA2_USERNAME"},
+               {"$eval": "AUTOGRAPH_AUTHENTICODE_SHA2_PASSWORD"},
+               ["gcp_prod_autograph_authenticode_202404", "gcp_prod_autograph_authenticode_202404_stub"],
+               "authenticode_dep_sha256"
+            ]
+            - ["https://prod.autograph.prod.webservices.mozgcp.net",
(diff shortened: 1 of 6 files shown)


bhearsum:
TODO: update a whole bunch of hardcodes in code that deal with formats

bhearsum:
This has been tested as thoroughly as possible. I covered:
* gecko: https://treeherder.mozilla.org/jobs?repo=try&revision=15f5fc77e1c8a17426954f67ae8b07e2645208a7
* staging-xpi-manifest: https://firefox-ci-tc.services.mozilla.com/tasks/Ts9ohSzPQsGDii2oJOynFQ and https://firefox-ci-tc.services.mozilla.com/tasks/X8E8BYZwTlevgL0LWZa-QA
* vpn: https://firefox-ci-tc.services.mozilla.com/tasks/groups/Z6u6AbcxT-qR4bnst5Jotg
* adhoc signing: https://github.com/mozilla-releng/adhoc-signing/pull/219

I was unable to test app services and glean due to not having anything except prod worker pools set up for them. (I don't think this is worth fixing as a blocker here, but I could be convinced otherwise.)

I was also unable to test the mobile trust domain via staging-reference-browser due to it not being in projects.yml (being addressed in https://github.com/mozilla-releng/fxci-config/pull/259), and I suspect that it is also does not have the firefoxci app installed. Again, I don't think this is worth blocking this on, but I could be convinced otherwise.

jcristau (approved):

PR #1100, opened 2024-12-04
Allow beetmover to receive optional exclude patterns

oskirby:
This patch attempts to add the ability for tasks to specify artifacts which should be excluded from beetmover. This has previously been done by updating the `RELEASE_EXCLUDE` in `constants.py` but that is a bit of a fragile way of making changes, and it would be handy if the tasks could provide this exclusion information in the payload.

With this change, a task can provide an optional `exclude` field in the payload, which contains an array of regex patterns to exclude from the beetmover action. For example:
```
payload:
   exclude:
      - somepath/*.exe
```

As perhaps a bit of an open question, should this be a literal path instead of a regex pattern? I went with regex simply because that's what is currently done with `RELEASE_EXCLUDE`

### DIFF
--- a/beetmoverscript/src/beetmoverscript/gcloud.py
+++ b/beetmoverscript/src/beetmoverscript/gcloud.py
@@ -203,6 +203,7 @@ async def push_to_releases_gcs(context):
 
     # Weed out RELEASE_EXCLUDE matches, but allow partners specified in the payload
     push_partners = context.task["payload"].get("partners", [])
+    exclude = context.task["payload"].get("exclude", []) + list(RELEASE_EXCLUDE)
 
     for blob_path in candidates_blobs.keys():
         if "/partner-repacks/" in blob_path:
@@ -214,7 +215,7 @@ async def push_to_releases_gcs(context):
                 )
             else:
                 log.debug("Excluding partner repack {}".format(blob_path))
-        elif not matches_exclude(blob_path, RELEASE_EXCLUDE):
+        elif not matches_exclude(blob_path, exclude):
             blobs_to_copy[blob_path] = blob_path.replace(candidates_prefix, releases_prefix)
         else:
             log.debug("Excluding {}".format(blob_path))
--- a/beetmoverscript/src/beetmoverscript/data/artifactMap_beetmover_task_schema.json
+++ b/beetmoverscript/src/beetmoverscript/data/artifactMap_beetmover_task_schema.json
@@ -89,6 +89,14 @@
                     "minItems": 1,
                     "uniqueItems": true
                 },
+                "exclude": {
+                    "type": "array",
+                    "minItems": 0,
+                    "uniqueItems": false,
+                    "items": {
+                        "type": "string"
+                    }
+                },
                 "artifactMap": {
                     "type": "array",
                     "items": {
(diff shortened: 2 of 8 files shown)


beetmoverscript/src/beetmoverscript/gcloud.py
```diff
@@ -203,6 +203,7 @@ async def push_to_releases_gcs(context):
 
     # Weed out RELEASE_EXCLUDE matches, but allow partners specified in the payload
     push_partners = context.task["payload"].get("partners", [])
+    exclude = context.task["payload"].get("exclude", []) + list(RELEASE_EXCLUDE)
```
bhearsum, line 206: It isn't a problem now, but I will note that this will subject VPN pushes to the existing `RELEASE_EXCLUDE` patterns, which could theoretically be a problem in the future. Hopefully we'll have removed this hardcode by the time any such thing comes up.

bhearsum (changes requested):
Thank you for doing this! It will enable some cleanup for Firefox as well.

I'm assuming that it's tasks like https://firefox-ci-tc.services.mozilla.com/tasks/TI0KzPifSVuVLjPBuydzyw that we want to exclude from uploading the `.deb` packages - please let me know if that's incorrect.

What's here looks good. https://github.com/mozilla-releng/scriptworker-scripts/blob/master/beetmoverscript/src/beetmoverscript/data/release_beetmover_task_schema.json will need to updated for this as well.

Once that change is done, we can push it to the `dev` beetmoverscript workers to test it against https://github.com/mozilla-releng/staging-mozilla-vpn-client. (You'll need to suffix [the worker-type](https://github.com/mozilla-releng/staging-mozilla-vpn-client/blob/5af9dcc72565d5133487d84e8cd1a367251ba30a/taskcluster/config.yml#L99) with `-dev` to do that, as well.)

beetmoverscript/src/beetmoverscript/gcloud.py
oskirby: From my understanding of the beetmover scripts, that was always the case and I felt that this approach was the best way to minimize the potential for unintended changes in behaviour. Once this PR lands we can probably work towards removing `RELEASE_EXCLUDE` by migrating the patterns into their associated tasks.

The other way we could have done this would be to use `RELEASE_EXCLUDE` as the default value for existing tasks that rely on `RELEASE_EXCLUDE` to omit files, for example:
```
exclude = context.task["payload"].get("exclude", list(RELEASE_EXCLUDE))
```

bhearsum: You're right - this is already the case. Sorry for the confusion. I think what you have here is fine, but the above suggestion would be fine as well, if you want to go that route.

bhearsum (approved):
Sorry for the delay here.

hneiva:
is this still valid? should we merge it before bit rot?

oskirby:
The feature we were hoping to use this for on the VPN has since been abandoned. So I no longer have a reason to push this patch through. I am not opposed to merging it if y'all think it will cleanup something in the Firefox side of things though.

hneiva:
I can't think of a place we could use it. Let's just close it and we can revisit later on if an use case arises.

PR #1102, opened 2024-12-11
Bug 1786807 - [treescript] Correctly apply esr suffix in version-bump.

jfx2006:
Applies to "release-version-bump" task. The "new" version comes from Shipit during release promotion and will have the esr suffix for "esr" release types.

### DIFF
--- a/treescript/src/treescript/gecko/versionmanip.py
+++ b/treescript/src/treescript/gecko/versionmanip.py
@@ -4,6 +4,7 @@
 import logging
 import os
 
+import attr
 from mozilla_version.gecko import FirefoxVersion, GeckoVersion, ThunderbirdVersion
 from mozilla_version.mobile import MobileVersion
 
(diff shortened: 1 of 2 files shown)


jfx2006:
From the bug description, version_display is the only file that should ever have the "esr" suffix.
My strategy is to check each file to see if it has the esr suffix and if the new version set in the payload has the suffix but the current file does not, drop the suffix from next_version before writing the new file.

I adjusted some tests accordingly. Most tests were writing only to milestone.txt, which by definition should never have an esr (or b_n_) suffix. So I added in the other files in a couple of places. The "test_bump_version" test with its "should_append_esr" parameter is removed since `bump_version` will always have the new version passed as an argument.

One could argue that `bump_version` is a poorly named function as it really sets the new version it was given.

In the interest of minimal change, I did not reuse the merge automation logic as suggested because merge automation works differently - given a file and a version field, it calculates the new version and then calls `do_bump_version` whereas version-bump is told what version to use in the payload.

jfx2006:
It occurred to me that this may break the release2esr merge automation. I will double check that.

PR #1105, opened 2024-12-18
Cancel bitrise tasks on taskcluster cancellation

### LINKED ISSUE #924: [bitrisescript] Attempt to cancel Bitrise builds when Task is pre-empted or cancelled
When a bitrisescript task is cancelled or has an infrastructure related error (e.g is pre-empted), we should attempt to cancel the in-progress Bitrise builds as well as they won't be useful anymore.

I think we at least need to handle SIGTERM in `script.py`, then cancel any in-progress futures, as well as add done callbacks to them to trigger the Bitrise API to abort their associated build.

I'm not 100% sure if a user cancelling the task will send SIGTERM or not..

### LINKED ISSUE #971: bitrisescript: support aborting bitrise builds
In certain circumstances (e.g user cancel), we should automatically abort the build in Bitrise as well to save resources.

Eijebong:
Fixes #971
Fixes #924

An example run:

image
image

### DIFF
--- a/bitrisescript/src/bitrisescript/bitrise.py
+++ b/bitrisescript/src/bitrisescript/bitrise.py
@@ -268,15 +268,20 @@ async def run_build(artifacts_dir: str, workflow_id: str, **build_params: Any) -
         "build_params": build_params,
     }
 
-    response = await client.request("/builds", method="post", json=data)
-    if response.get("status", "") != "ok":
-        raise Exception(f"Bitrise status for '{workflow_id}' is not ok. Got: {response}")
+    build_slug = None
+    try:
+        response = await client.request("/builds", method="post", json=data)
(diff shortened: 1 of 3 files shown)


bitrisescript/src/bitrisescript/bitrise.py
```diff
@@ -236,14 +236,19 @@ async def wait_and_download_workflow_log(artifacts_dir: str, build_slug: str) ->
         artifacts_dir (str): Directory to download artifacts to.
         build_slug (str): Identifier of workflow to run.
     """
+    skip_log_retrieval = False
```
jcristau: nit: `retrieve_log = True`, to avoid double negation.

hneiva (approved):
LGTM

PR #1108, opened 2024-12-19
feat(pushmsix): Bug 1817658 - Add TB to pushmsix

hneiva:


pushmsixscript/docker.d/init_worker.sh
```diff
@@ -15,29 +15,56 @@ export LOGIN_URL=https://login.microsoftonline.com
...
-    export TENANT_ID=`echo $TENANT_ID | base64 -d`
-    export CLIENT_ID=`echo $CLIENT_ID | base64 -d`
-    export CLIENT_SECRET=`echo $CLIENT_SECRET | base64 -d`
+  thunderbird)
+    export BETA_APPLICATION_ID=9PFQWVMSS45P
+    export RELEASE_APPLICATION_ID=9PM5VM1S3VMQ
     ;;
   *)
-    exit 1
+    # Product validation below will handle this case
```
bhearsum: Please keep the `exit 1` here. It may be currently true that product validation will handle this now, but it's not guaranteed that it will in the future.

pushmsixscript/docker.d/init_worker.sh
```diff
@@ -15,29 +15,56 @@ export LOGIN_URL=https://login.microsoftonline.com
...
+  prod)
+    case $COT_PRODUCT in
+      firefox|thunderbird)
+        # For firefox prod environment
+        test_var_set 'TENANT_ID'
+        test_var_set 'CLIENT_ID'
+        test_var_set 'CLIENT_SECRET'
+        export TENANT_ID=$(echo "$TENANT_ID" | base64 -d)
+        export CLIENT_ID=$(echo "$CLIENT_ID" | base64 -d)
+        export CLIENT_SECRET=$(echo "$CLIENT_SECRET" | base64 -d)
```
bhearsum, line 57: Do we have values for these already, or are we waiting on someone from Thunderbird or elsewhere to provide them?

bhearsum (changes requested):

pushmsixscript/docker.d/init_worker.sh
hneiva: Yes, already added them to SOPS.

bhearsum (approved):

PR #1111, opened 2024-12-27
signingscript: remove debsign support

bhearsum:
As far as I can tell we have not used this since it was disabled in https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4773. I believe we have no desire to use this again, as Debian packages are being signed through Google Artifact Registry instead at this point.


bhearsum:
Flagging @gabrielBusta and @oskirby for visibility.

gabrielbusta (approved):

PR #1113, opened 2025-01-02
Bug 1939751 - Use proper keyids for omnija signing

ahal:
The default keyids here were still using the old root CA. These should be using the new CA.


hneiva (approved):
LGTM

PR #1116, opened 2025-01-09
chore: reformat with ruff 0.9.0

bhearsum:


beetmoverscript/src/beetmoverscript/constants.py
```diff
@@ -69,7 +69,7 @@
 
 HASH_BLOCK_SIZE = 1024 * 1024
 
-RELEASE_BRANCHES = ("mozilla-central", "mozilla-beta", "mozilla-release", "mozilla-esr52" "comm-central", "comm-beta", "comm-esr60")
+RELEASE_BRANCHES = ("mozilla-central", "mozilla-beta", "mozilla-release", "mozilla-esr52comm-central", "comm-beta", "comm-esr60")
```
hneiva: I think this is not intentional.
Shouldn't it be `"mozilla-esr52", "comm-central"` ?

hneiva (changes requested):

beetmoverscript/src/beetmoverscript/constants.py
bhearsum: Indeed. Which actually means this has been broken for awhile :laughing:

hneiva (approved):

PR #1117, opened 2025-01-09
fix: pin ruff correctly in tox.ini and pre-commit

bhearsum:


beetmoverscript/src/beetmoverscript/constants.py
```diff
@@ -69,7 +69,7 @@
 
 HASH_BLOCK_SIZE = 1024 * 1024
 
-RELEASE_BRANCHES = ("mozilla-central", "mozilla-beta", "mozilla-release", "mozilla-esr52" "comm-central", "comm-beta", "comm-esr60")
+RELEASE_BRANCHES = ("mozilla-central", "mozilla-beta", "mozilla-release", "mozilla-esr52comm-central", "comm-beta", "comm-esr60")
```
hneiva: Copy-pasting the comment here :)
I think this is not intentional.
Shouldn't it be `"mozilla-esr52", "comm-central"` ?

hneiva (changes requested):

beetmoverscript/src/beetmoverscript/constants.py
bhearsum: argh, reformatting wasn't supposed to happen here...that's happening in https://github.com/mozilla-releng/scriptworker-scripts/pull/1116

bhearsum:
This is now stacked on https://github.com/mozilla-releng/scriptworker-scripts/pull/1116; I'll rebase it after that merges.

hneiva (approved):
LGTM!

PR #1122, opened 2025-01-16
bug 1942344: signingscript: import new authenticode certs

bhearsum:
These new certs are associated with the new authenticode keys in the Autograph GCP production environment, so we need new formats to along with them. Both the EV and non-EV cert are using same intermediates as before, so we don't need any changes to the CA bundles.

I've slightly simplified the `sign.py` blocks that deal with certs and cafiles as a part of this, which I'm _fairly_ certain is safe, but I'll be testing in dev to gain some more confidence in that.


signingscript/src/signingscript/sign.py
```diff
@@ -1364,10 +1364,15 @@ async def signer(digest, digest_algo):
     if fmt in ("autograph_authenticode_ev", "gcp_prod_autograph_authenticode_ev", "stage_autograph_authenticode_ev"):
         cafile_key = f"{cafile_key}_ev"
         cert_key = f"{cert_key}_ev"
-    elif fmt.startswith(("autograph_authenticode_202404", "gcp_prod_autograph_authenticode_202404", "stage_autograph_authenticode_202404")):
```
bhearsum, line 1367: This line being replaced actually fixes a possible footgun. With the current code, `stub` formats would not select the `_202404` version of the key. The only reason this hasn't caused problems is because the `authenticode_cert` and `authenticode_ca` values were the same as their `_202404` counterparts.

signingscript/src/signingscript/sign.py
```diff
@@ -1389,7 +1394,7 @@ async def signer(digest, digest_algo):
     else:
         timestamp_style = context.config["authenticode_timestamp_style"]
     timestamp_url = context.config["authenticode_timestamp_url"]
-    if fmt.endswith(("authenticode_stub", "authenticode_sha2_stub", "authenticode_202404_stub")):
+    if fmt.endswith("stub"):
```
bhearsum, line 1397: There is some theoretical world where this might be too inclusive, but at the moment I think this is totally fine.

signingscript/docker.d/init_worker.sh
```diff
@@ -29,22 +29,30 @@ export GPG_PUBKEY_PATH=$APP_DIR/signingscript/src/signingscript/data/gpg_pubkey_
...
 if [ "$ENV" == "prod" ]; then
   export AUTHENTICODE_TIMESTAMP_STYLE=old
+  export AUTHENTICODE_CERT_PATH_202501=$APP_DIR/signingscript/src/signingscript/data/authenticode_prod_202501.crt
   export AUTHENTICODE_CERT_PATH_202404=$APP_DIR/signingscript/src/signingscript/data/authenticode_prod_202404.crt
   export AUTHENTICODE_CERT_PATH="$AUTHENTICODE_CERT_PATH_202404"
   export AUTHENTICODE_CERT_PATH_EV=$APP_DIR/signingscript/src/signingscript/data/authenticode_prod_ev.crt
-  export AUTHENTICODE_CA_PATH_202404=$APP_DIR/signingscript/src/signingscript/data/authenticode_prod_ca_202404.crt
+  export AUTHENTICODE_CERT_PATH_EV_202501=$APP_DIR/signingscript/src/signingscript/data/authenticode_prod_ev_202501.crt
+  export AUTHENTICODE_CA_PATH_202501=$APP_DIR/signingscript/src/signingscript/data/authenticode_prod_ca_202404.crt
+  export AUTHENTICODE_CA_PATH_202404="$AUTHENTICODE_CA_PATH_202404"
```
jcristau:
```suggestion
  export AUTHENTICODE_CA_PATH_202404="$AUTHENTICODE_CA_PATH_202501"
```

signingscript/docker.d/init_worker.sh
```diff
@@ -29,22 +29,30 @@ export GPG_PUBKEY_PATH=$APP_DIR/signingscript/src/signingscript/data/gpg_pubkey_
...
   export AUTHENTICODE_CA_PATH="$AUTHENTICODE_CA_PATH_202404"
-  export AUTHENTICODE_CA_PATH_EV=$APP_DIR/signingscript/src/signingscript/data/authenticode_prod_ev.crt
+  export AUTHENTICODE_CA_PATH_EV_202501=$APP_DIR/signingscript/src/signingscript/data/authenticode_prod_ev.crt
```
jcristau: We could use the opportunity to fix the EV bundles.

AUTHENTICODE_CA_PATH_EV_202501 would have just the root (DigiCert Trusted Root G4), and AUTHENTICODE_CERT_PATH_EV_202501 would have our cert and the intermediate (DigiCert Trusted G4 Code Signing RSA4096 SHA384 2021 CA1).

signingscript/docker.d/passwords.yml
```diff
@@ -85,7 +86,8 @@ in:
             - ["https://prod.autograph.prod.webservices.mozgcp.net",
                {"$eval": "AUTOGRAPH_AUTHENTICODE_SHA2_USERNAME"},
                {"$eval": "AUTOGRAPH_AUTHENTICODE_SHA2_PASSWORD"},
-               ["gcp_prod_autograph_authenticode_202404", "gcp_prod_autograph_authenticode_202404_stub"],
+               ["gcp_prod_autograph_authenticode_202404", "gcp_prod_autograph_authenticode_202404_stub",
```
jcristau: Having `gcp_prod_autograph_authenticode_202404` here might actually not be a good idea, since the prod environment won't have that key?

signingscript/docker.d/init_worker.sh
bhearsum: erm, thank you

signingscript/docker.d/passwords.yml
bhearsum: Yeah, that's a good call. Will do.

bhearsum:
I've tested this as much as I think it makes sense to in dev, with this try push: https://treeherder.mozilla.org/jobs?repo=try&revision=ff97dbc8c201da61d5981e04bf5cd41cf64ac005. This has verified that the `202501` flows are working on a basic level. We can't do any real tests of the new cert until we add `gcp_prod` to L3 workers. I intend to do this as a follow up. Once this patch and that land in production I'll do testing of the `202501` cert on a project branch, and we can probably test the EV one with adhoc-signing at that point too.

(I guess I could also test vpn dev in staging...but it doesn't seem worthwhile to do.)

bhearsum:
I ended up renaming the new formats to `202412`; as it turns out, that's what's in the new keyid, so I thought it best to be consistent. A new try push for that is in https://treeherder.mozilla.org/jobs?repo=try&revision=aef4e362ab7f469a6a98f671017b5128504507e4 (the first `gd-s` failure is because that task ran before the latest code here was pushed to dev).

signingscript/docker.d/init_worker.sh
```diff
@@ -29,22 +29,30 @@ export GPG_PUBKEY_PATH=$APP_DIR/signingscript/src/signingscript/data/gpg_pubkey_
...
   export AUTHENTICODE_TIMESTAMP_STYLE=old
+  export AUTHENTICODE_CERT_PATH_202412=$APP_DIR/signingscript/src/signingscript/data/authenticode_prod_202412.crt
   export AUTHENTICODE_CERT_PATH_202404=$APP_DIR/signingscript/src/signingscript/data/authenticode_prod_202404.crt
   export AUTHENTICODE_CERT_PATH="$AUTHENTICODE_CERT_PATH_202404"
   export AUTHENTICODE_CERT_PATH_EV=$APP_DIR/signingscript/src/signingscript/data/authenticode_prod_ev.crt
-  export AUTHENTICODE_CA_PATH_202404=$APP_DIR/signingscript/src/signingscript/data/authenticode_prod_ca_202404.crt
-  export AUTHENTICODE_CA_PATH="$AUTHENTICODE_CA_PATH_202404"
-  export AUTHENTICODE_CA_PATH_EV=$APP_DIR/signingscript/src/signingscript/data/authenticode_prod_ev.crt
+  export AUTHENTICODE_CERT_PATH_EV_202412=$APP_DIR/signingscript/src/signingscript/data/authenticode_prod_ev_202412.crt
+  export AUTHENTICODE_CA_PATH_202412=$APP_DIR/signingscript/src/signingscript/data/authenticode_prod_ca_202412.crt
```
jcristau: This file doesn't seem to exist, should we point at `authenticode_prod_ca_202404.crt` instead?  Or did you mean to rename `_202404` to `_202412`?

jcristau (approved):
Just one issue with the CA config, otherwise lgtm, thanks!

signingscript/docker.d/init_worker.sh
bhearsum: Thanks for catching that.

To be honest, I'm not sure what I intended here. With fresh eyes though, it seems like removing the date is the best option seeing as there's only one version of it after https://github.com/mozilla-releng/scriptworker-scripts/commit/e4fc7ac12f31e44f5e25e1da7edf5e568d3388ad.

PR #1123, opened 2025-01-17
bug 1942364: signingscript: autograph gcp migration step 4: add gcp_prod entries for L3 workers

bhearsum:
This adds `gcp_prod` prefixes for everything in the production workers, except for the authenticode stub format in `vpn`, which it doesn't actually use.

The new authenticode keyids require new hawk credentials. There is a `gcp-prod-level-3` branch in the sops repo with the necessary changes staged in it that needs to be pushed before this rolls out.

This is step 4 of https://docs.google.com/document/d/1nOZoNfAg96pPUMU1_NGOoRQD0gQcWgxyMYRnKnOgyJg/edit?tab=t.0#heading=h.xv0nvnha14du, and relates to https://mozilla-hub.atlassian.net/browse/AUT-300.

PR #1127, opened 2025-01-28
use separate env vars for gcp prod authenticode formats

bhearsum:
And also fix the lingering `202501` formats that I somehow missed in #1123.

There's a necessary cloudops-infra and secrets change coming as part of this, too.


bhearsum:
New secrets have been added; https://github.com/mozilla-services/cloudops-infra/pull/6209 covers the cloudops-infra change.

jcristau (approved):
Interesting; I thought we were going to use the same creds for aws/gcp.

bhearsum:
> Interesting; I thought we were going to use the same creds for aws/gcp.

Yeah, that was the plan, and in fact is the case for all of the other secrets. We ended up with new ones for authenticode because of the "name them after the sole signer they use" convention. (We could change things on the autograph side instead of doing this, but I think it's just as easy to deal with it here.) So, this changed after the original plan was written :(.

PR #1128, opened 2025-01-30
fix(signingscript): gcp prod langpack keyid

bhearsum:
I bungled this in my original patch; it has a different month than both the webextensions dep key and the systemaddons prod key.


jcristau (approved):
More of a rubberstamp than anything else...

PR #1133, opened 2025-02-10
fix(signingscript): use newer systemaddon dep keyid for firefox dep/f…

bhearsum:
…ake-prod

We're already using this elsewhere; this was a simple typo


hneiva (approved):
lgtm

PR #1097, opened 2024-12-03
WIP: add tests that operate at the API level

bhearsum:
This is inspired by recent frustrations of surprise behaviour by signingscript, and feeling like the current test-every-tiny-little-function-independently tests wouldn't allow for any sort of confident refactoring. These new tests that I'm adding use the same entry point that `scriptworker` calls into: `async_main`, and they mock out as little as they possible can. Currently, this is calls to Autograph, and `verify_mar_signature`, but there will probably be other things mocked out (eg: anything that depends on a signature from a real certificate will need to be mocked).

Running just these new tests alone already achieves 35% coverage overall, and 27% coverage of `sign.py`:
```
tests/test_script.py::test_gpg_signing[autograph_gpg] PASSED                                                                                                                                [ 25%]
tests/test_script.py::test_gpg_signing[stage_autograph_gpg] PASSED                                                                                                                          [ 50%]
tests/test_script.py::test_mar_signing[autograph_hash_only_mar384] cPASSED                                                                                                                   [ 75%]
tests/test_script.py::test_mar_signing[stage_autograph_hash_only_mar384] PASSED                                                                                                             [100%]

---------- coverage: platform linux, python 3.11.6-final-0 -----------
Name                              Stmts   Miss Branch BrPart  Cover   Missing
-----------------------------------------------------------------------------
src/signingscript/__init__.py         0      0      0      0   100%
src/signingscript/exceptions.py      11      3      0      0    73%   17, 30, 43
src/signingscript/rcodesign.py       71     58     28      0    13%   25-51, 62-69, 82-90, 99-112, 121-132, 143-152
src/signingscript/script.py          72     29     34      7    55%   32, 34, 37-38, 41-43, 55, 57, 71-74, 112-130, 135-137
[...]

PR #1135, opened 2025-02-13
add landoscript

bhearsum:
This (fairly hefty) pull request adds a new scriptworker: landoscript. This new scriptworker will be replacing treescript when writes to hg.mozilla.org for Gecko repositories cease to happen.

Because it ultimately fulfills the same use cases as treescript, I've chosen to keep the payload as similar as possible. There are some necessary changes (mostly around using lando repository names instead of references to hg.mozilla.org repositories). I've also removed support for some things that are not used (eg: `ignore_closed_tree` and `dontbuild` are either not present for most actions, or always have the same value).

Some notes on the tests included here:
* Most of the success tests have payloads taken directly from real world treescript tasks (with some small tweaks, per the above note). I fished around and tried to find all variations of these tasks that exist (most notably for `merge_day` and `version_bump` tasks, which are often slightly different depending on which repository they are running from).
* The approach I've taken to testing here is the same as what I've been attempting to do for signingscript in https://github.com/mozilla-releng/scriptworker-scripts/pull/1097: execute tests by running `async_main`, rather than deeply unit testing individual components. This does end up making the tests notable bigger, which can be intimidating, but I've found that once written they need little to no touching, and has made refactoring as I go along significantly easier. There is probably an argument to be made that certain helper pure functions (like `diff_contents`) could do with some separate unit testing - but I've not gone to that length yet.
* I'm not terribly happy with how the responses for the github client are set-up. Because the GraphQL API uses a single endpoint for all requests, and I'm simply returning responses in a fixed order, the ordering of the `aioresponses` calls is important. In an ideal world we would be able to key responses off of parts of the requests, but I haven't had time to make this happen.

At the time of writing I have not yet been able to test it in a dev environment - I've merely been working off of the spec at https://docs.google.com/document/d/1F3xYp6v4YLLHsX7zFkYoygDyvIUPzujxeMW7W4qte0o/edit?tab=t.0#heading=h.c29tth48xps4. Seeing as this is completely unused so far, I'm hoping to get what's here merged, and then follow up with more targeted pull requests as I fix issues that I encounter when testing in the real world.


landoscript/src/landoscript/script.py
```diff
@@ -0,0 +1,73 @@
...
+    }
+    return default_config
+
+
+async def async_main(context):
+    config = context.config
+    payload = context.task["payload"]
+    artifact_dir = config["artifact_dir"]
+    public_artifact_dir = os.path.join(artifact_dir, "public", "build")
+    owner, repo = extract_github_repo_owner_and_name(payload["source_repo"])
```
ahal: I think this will need to be extracted out of the scopes instead of the payload. Or else we'll need to validate scopes independently (but then `source_repo` would need to be defined in the task definition twice)

landoscript/src/landoscript/script.py
```diff
@@ -0,0 +1,73 @@
...
+
+    os.makedirs(public_artifact_dir)
+
+    lando_actions = []
+    async with GithubClient(context.config["github_config"], owner, repo) as gh_client:
+        for action in payload["actions"]:
```
ahal: Same here, I assume we'll want to protect actions via scopes too.

E.g, maybe we want to allow generating the changelog, but not bumping the version (just thinking of possible future use cases)

landoscript/src/landoscript/github.py
```diff
@@ -0,0 +1,84 @@
...
+
+        Returns:
+            Dict: The dictionary of file contents of the form `{<path>: <contents>}`.
+        """
+        branch = branch or "HEAD"
+
+        if isinstance(files, str):
+            files = [files]
+
+        # Periods are not legal GraphQL key names.
```
ahal: nit: Periods and slashes

landoscript/src/landoscript/github.py
```diff
@@ -0,0 +1,84 @@
...
+from simple_github import AppClient
+
+log = logging.getLogger(__name__)
+
+
+class UnknownBranchError(Exception):
+    pass
+
+
+class GithubClient:
```
ahal: Maybe it's worth putting this in `scriptworker_client.github`? The slashes fix you incorporated here is evidence that it might be

landoscript/src/landoscript/actions/version_bump.py
```diff
@@ -0,0 +1,122 @@
+import logging
```
ahal, line 1: I think an argument could also be made to share the actual file manipulation logic between landoscript and treescript/githubscript. But that can be a future cleanup (and probably won't ever be a priority)

The actions you're concerned with are all Gecko specific anyway, so there's not much point. But in the future, we could have e.g a "pre-commit" action that applies lint fixes, or a "commitizen" action that generates a changelog. That kind of stuff might benefit from being shared across both scripts.

landoscript/src/landoscript/script.py
bhearsum: Ah, indeed! I have a "add scopes" todo item still; I'll make sure to address this specific point as part of that.

landoscript/src/landoscript/github.py
bhearsum: Good call, yeah, will do.

landoscript/src/landoscript/actions/version_bump.py
bhearsum: That's a fair point. I don't really mind moving that sort of thing to shared code. It's not terribly harmful to live in `scriptworker_client` (or perhaps a new module?) even if there's never any re-use?

I know we've talked about this a bit at points, but it would probably be good to write down our long term vision for lando/github/treescript going forward. If I recall correctly, we anticipate treescript going away entirely, and any lando-enabled repo operations will move to landoscript (which will include all current supported hg repos?), and the remainder will move to githubscript? (I say this mainly selfishly, because I often get myself confused between the three of them...) The answers here may influence whether or not it's worthwhile to move some code out.

ahal: Good point, could be worth writing up a quick RFC. I think you should just leave it here for now though, no need to get distracted by that yet.

landoscript/src/landoscript/github.py
bhearsum: I moved this part out to https://github.com/mozilla-releng/scriptworker-scripts/pull/1147.

landoscript/src/landoscript/script.py
bhearsum: I pushed an update that deals with this and the next point. I ended up adding separate `repo` and `branch` scopes (this will be needed for the main Firefox repository, where we'll have separate branches in it for various L3 repos), as well as `action` scopes. If you have any concerns please let me know!

bhearsum: ...actually, no `branch` scope. (After a conversation with Connor, there was a change to the Lando API: we'll be using short repo identifiers in the URLs that map to an actual github repository+branch.)

landoscript/src/landoscript/lando.py
```diff
@@ -0,0 +1,77 @@
...
+
+def create_commit_action(commitmsg, diff):
+    author = "Release Engineering Landoscript <release+landoscript@mozilla.com>"
+    timestamp = datetime.datetime.now(datetime.timezone.utc).isoformat()
+
+    return {"action": "create-commit", "commitmsg": commitmsg, "diff": diff, "date": timestamp, "author": author}
+
+
+async def submit(session, lando_api, lando_repo, actions, sleeptime_callback=calculate_sleep_time):
+    url = f"{lando_api}/api/v1/{lando_repo}"
```
cgsheeh: We no longer use the `v1` in the URL, so this should instead look like `f"{lando_api}/api/repo/{lando_repo}"`.

landoscript/src/landoscript/lando.py
```diff
@@ -0,0 +1,92 @@
...
+        log.info(pprint(json))
+
+        submit_resp = await retry_async(
+            session.post,
+            args=(url,),
+            kwargs={
+                "json": json,
+                "raise_for_status": True,
+                # TODO: is this a bearer token?
+                "headers": {"Authorization": f"Bearer {lando_token}"},
```
cgsheeh: You will need to add a `User-Agent` here that looks like `Lando-User/{username}`.

landoscript/src/landoscript/lando.py
```diff
@@ -0,0 +1,92 @@
...
+        status_resp = await session.get(status_url)
+
+        # just retry if something went wrong...
+        if not status_resp.ok:
+            log.info(f"lando response is not ok (code {status_resp.status}), trying again...")
+            continue
+
+        if status_resp.status == 200:
+            body = await status_resp.json()
+            if body.get("status") != "completed":
```
cgsheeh: You should look for `"LANDED"` as the status here.

cgsheeh (changes requested):

landoscript/src/landoscript/lando.py
bhearsum: Can I use whatever value I want here?

cgsheeh: Any value is fine, but something like the email or username for the bot would be preferred. :+1:

landoscript/src/landoscript/lando.py
```diff
@@ -0,0 +1,95 @@
...
+async def submit(
+    session: ClientSession,
+    lando_api: str,
+    lando_token: str,
+    lando_repo: str,
+    actions: list[LandoAction],
+    sleeptime_callback: Callable[..., Any] = calculate_sleep_time,
+) -> str:
+    """Submit the provided `actions` to the given `lando_repo` through the `lando_api`."""
+    url = f"{lando_api}/api/{lando_repo}"
```
cgsheeh: This should look like `/api/repo/{lando_repo}` (unless `lando_repo` already includes `/repo`).

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -173,3 +174,107 @@ async def get_branch_head_oid(self, branch: str) -> str:
...
+                          $obj_expr
+                        }
+                      }
+                    }
+                    """
+            )
+        )
+        file_expr = leaf_expr
+        for _ in range(depth_per_query - 1):
+            file_expr = recursive_expr.substitute(obj_expr=file_expr)
```
ahal: Maybe one day I will understand GraphQL.. but that day is not today ><.

I'm guessing you considered using https://docs.github.com/en/rest/repos/contents?apiVersion=2022-11-28 but this can get the data more efficiently?

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -173,3 +174,107 @@ async def get_branch_head_oid(self, branch: str) -> str:
             raise UnknownBranchError(f"branch '{branch}' not found in repo!")
 
         return repo["object"]["oid"]
+
+    async def get_file_listing(self, path: str = "", branch: Optional[str] = None, depth_per_query=10) -> List[str]:
```
ahal: Please add a docstring similar to the other functions

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -173,3 +174,107 @@ async def get_branch_head_oid(self, branch: str) -> str:
...
+                # it will already be included in the files returned.
+                prefix = str(refetch.parent)
+                # Then we combine the proper prefix with each returned file, giving us
+                # the full path to the file from the root of the repository.
+                files.append(f"{prefix}/{deep_file}")
+
+        return files
+
+
+def process_file_listings(entries: List[Dict[str, Any]], prefix: Path = Path("")) -> Tuple[List[str], List[Path]]:
```
ahal: nit: it seems a bit strange for this to be a module function. Seems like it should be a private method on the client class.

landoscript/docker.d/worker.yml
```diff
@@ -0,0 +1,12 @@
...
+verbose: { "$eval": "VERBOSE == 'true'" }
+lando_api: { "$eval": "LANDO_API" }
+lando_token: { "$eval": "LANDO_TOKEN" }
+# maps the repo+branch portion of lando URLs to github repo information
+# TODO: switch this to lookup in lando when that API exists
+lando_name_to_github_repo:
+  autoland:
+    owner: mozilla-firefox
+    repo: firefox
+    branch: autoland
```
cgsheeh: The current output from this endpoint (the URL is `/api/repoinfo/<phab_short_name>`) looks like this:

```
{
	"repo_url": "https://github.com/mozilla-firefox/firefox",
	"branch_name": "main",
	"scm_level": "main.scm_level_3"
}
```

I guess the `owner` and `repo` fields you have here are derived from the `repo_url` we have. Should I split those out on our side so they don't need to be parsed from the URL?

I can also remove the `main.` part from the `scm_level`, that's a detail of our permissions implementation.

landoscript/src/landoscript/actions/version_bump.py
```diff
@@ -0,0 +1,143 @@
...
+    "browser/": FirefoxVersion,
+    "config/milestone.txt": GeckoVersion,
+    "mobile/android/": MobileVersion,
+    "mail/": ThunderbirdVersion,
+}
+
+
+def log_file_contents(contents):
+    for line in contents.splitlines():
+        log.info(line)
```
ahal: It's not clear to me why this is better than just logging contents directly. Having the logger preamble in front of each line just makes it harder to copy/paste imo.

Feel free to ignore this if this was just copy/pasted and you want to keep it as similar as possible.

cgsheeh (changes requested):

landoscript/src/landoscript/lando.py
```diff
@@ -0,0 +1,76 @@
...
+
+    status_url = (await submit_resp.json()).get("status_url")
+    if not status_url:
+        raise LandoscriptError("couldn't find status url!")
+
+    return status_url
+
+
+async def poll_until_complete(session: ClientSession, poll_time: int, status_url: str):
+    while True:
```
ahal, line 72: So I guess the idea is that we poll until we hit the Taskcluster timeout?

I think we might be losing some of the idempotent-ness of  the old approach here. It will certainly be possible for this task to fail due to timeout, but Lando could still land it afterwards. This means a failed task doesn't necessarily indicate a failed landing.

Do you know what would happen if we re-ran the task in such a case? For version bump, it might be worth ensuring that if the requested version is already what's committed, that we exit successfully. I'm not really sure what to do in the general case..

Another approach might be to timeout within the task, and abort the landing prior to failing. That way there's more of a guarantee that a failed landoscript task means the push didn't get landed. I don't think this needs to be fully solved for the MVP

landoscript/src/landoscript/script.py
```diff
@@ -0,0 +1,104 @@
...
+    }
+    missing = expected_scopes - scopes
+    if missing:
+        raise scriptworker.client.TaskVerificationError(f"required scope(s) not present: {', '.join(missing)}")
+
+
+# `context` is kept explicitly untyped because all of its members are typed as
+# Optional. This never happens in reality (only in tests), but as things stand
+# at the time of writing, it means we need noisy and unnecessary None checking
+# to avoid linter complaints.
```
ahal, line 44: Might be worth a test fixture to provide dummy values for everything in context? I don't really care that much though, if it's super annoying don't bother.

landoscript/src/landoscript/script.py
```diff
@@ -0,0 +1,104 @@
...
+async def async_main(context):
+    config = context.config
+    payload = context.task["payload"]
+    scopes = set(context.task["scopes"])
+    artifact_dir = config["artifact_dir"]
+    public_artifact_dir = os.path.join(artifact_dir, "public", "build")
+
+    # Note: `lando_repo` is not necessarily the same as the repository's name
+    # on Github.
+    lando_repo = payload["lando_repo"]
```
ahal: It's a common pattern to just parse this out of the scopes in other scriptworkers, that way you don't need to validate it independently, nor have it defined twice in the task definition.

I'm a bit on the fence on which way is better, so if you prefer this approach, I'm fine with leaving it.

landoscript/src/landoscript/script.py
```diff
@@ -0,0 +1,104 @@
...
+    validate_scopes(scopes, lando_repo, payload["actions"])
+
+    os.makedirs(public_artifact_dir)
+
+    lando_actions: list[lando.LandoAction] = []
+    async with GithubClient(context.config["github_config"], owner, repo) as gh_client:
+        for action in payload["actions"]:
+            log.info(f"processing action: {action}")
+
+            if action == "version_bump":
```
ahal: I think it'll be cleaner to create some kind of module level mapping to call these, rather than if/else chaining.

landoscript/src/landoscript/script.py
```diff
@@ -76,6 +76,9 @@ async def async_main(context):
                 # sometimes version bumps are no-ops
                 if version_bump_action:
                     lando_actions.append(version_bump_action)
+            elif action == "tag":
+                tag_actions = tag.run(payload["tags"])
+                lando_actions.extend(tag_actions)
```
ahal: Going back to the module level mapping, I don't think it would hurt to generalize the `None` test to all actions here (even if the action implementation would never return `None`)

landoscript/src/landoscript/actions/l10n_bump.py
```diff
@@ -0,0 +1,231 @@
...
+    path: str
+
+
+class L10nBumpInfo(TypedDict):
+    path: str
+    name: str
+    l10n_repo_url: str
+    l10n_repo_target_branch: str
+    ignore_config: dict[str, list[str]]
+    platform_configs: list[PlatformConfig]
```
ahal, line 33: Tbh I'm not really sure what `TypedDict` is meant for, but these look like they should be dataclasses.

landoscript/src/landoscript/script.py
```diff
@@ -64,42 +67,58 @@ async def async_main(context):
...
+                        if not await is_tree_open(session, config["treestatus_url"], lando_repo, config["sleeptime_callback"]):
+                            log.info("Treestatus is closed; skipping l10n bump.")
+                            continue
+
+                    l10n_bump_actions = await l10n_bump.run(
+                        gh_client, context.config["github_config"], public_artifact_dir, branch, payload["l10n_bump_info"], dontbuild, ignore_closed_tree
+                    )
+                    # sometimes nothing has changed!
+                    if l10n_bump_actions:
+                        lando_actions.extend(l10n_bump_actions)
```
ahal, line 113: I guess the actions are different enough that my previous suggestion of generalizing this might be a bit of a pain?

I still think it would be do-able, and that there's probably some benefit to having a standardized interface around calling actions and handling their results.. But I also don't want to cause a major refactor of this stack if things are already working.

So I guess feel free to leave it for now :)

landoscript/src/landoscript/actions/merge_day.py
```diff
@@ -0,0 +1,285 @@
...
+    to_branch: str
+    from_branch: str
+    base_tag: str
+    end_tag: str
+    merge_old_head: bool
+    fetch_version_from: str
+    touch_clobber_file: bool
+    version_files: list[VersionFile]
+    replacements: list[list[str]]
+    regex_replacements: list[list[str]]
```
ahal: Same comment, these should likely be dataclasses.

landoscript/src/landoscript/actions/android_l10n_import.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class TomlInfo(TypedDict):
+    toml_path: str
+    dest_path: str
+
+
+class AndroidL10nImportInfo(TypedDict):
+    from_repo_url: str
+    toml_info: list[TomlInfo]
```
ahal, line 30: These look like dataclasses.

landoscript/src/landoscript/actions/android_l10n_import.py
```diff
@@ -0,0 +1,106 @@
...
+        # we always take the tip of the default branch when importing new strings
+        toml_contents = await l10n_github_client.get_files(toml_files)
+        l10n_files: list[L10nFile] = []
+
+        for info in android_l10n_import_info["toml_info"]:
+            toml_file = info["toml_path"]
+            log.info(f"processing toml file: {toml_file}")
+
+            if toml_contents[toml_file] is None:
+                raise LandoscriptError(f"toml_file '{toml_file}' is not present in repository")
```
ahal, line 59: The variable repetition is a bit odd here. Couldn't you just iterate over `toml_contents.items()` instead of accessing `android_l10n_import_info` multiple times and then needing validation?

landoscript/src/landoscript/actions/android_l10n_import.py
```diff
@@ -0,0 +1,106 @@
+import logging
+import os.path
+from pathlib import Path
+from typing import TypedDict
+
+import tomli
```
ahal, line 8: IIRC we're now using Python 3.11 for all scriptworkers other than iscript right?

If that's the case, just use the stdlib:
https://docs.python.org/3/library/tomllib.html

landoscript/src/landoscript/actions/android_l10n_sync.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class TomlInfo(TypedDict):
+    toml_path: str
+
+
+class AndroidL10nSyncInfo(TypedDict):
+    from_repo_url: str
+    from_branch: str
+    toml_info: list[TomlInfo]
```
ahal, line 28: Ok, so I finally looked it up. I guess one reason to use `TypedDict` is if, well, you need a dict rather than a class. I'm not really sure why that would be necessary, but maybe avoiding converting everything is reason enough.

It's worth noting that you can convert a dataclass to a dict with the `dataclasses.asdict` module function.

landoscript/src/landoscript/actions/android_l10n_sync.py
```diff
@@ -0,0 +1,95 @@
...
+    toml_files = [info["toml_path"] for info in android_l10n_sync_info["toml_info"]]
+    toml_contents = await github_client.get_files(toml_files, branch=from_branch)
+    l10n_files: list[L10nFile] = []
+
+    for info in android_l10n_sync_info["toml_info"]:
+        toml_file = info["toml_path"]
+        log.info(f"processing toml file: {toml_file}")
+
+        if toml_contents[toml_file] is None:
+            raise LandoscriptError(f"toml_file '{toml_file}' is not present in repository")
```
ahal, line 51: Same comment about the odd variable shuffling as before.

landoscript/src/landoscript/actions/android_l10n_sync.py
```diff
@@ -0,0 +1,95 @@
...
+
+    # fetch l10n_files from gecko repo
+    dst_files = [f["dst_name"] for f in l10n_files]
+    log.info(f"fetching original files from l10n repository: {dst_files}")
+    orig_files = await github_client.get_files(dst_files, branch=to_branch)
+
+    diff = ""
+    for l10n_file in l10n_files:
+        if l10n_file["dst_name"] not in orig_files:
+            log.warning(f"WEIRD: {l10n_file['dst_name']} not in dst_files, continuing anyways...")
```
ahal: Lol, I approve of this new log level.

landoscript/tests/conftest.py
```diff
@@ -126,26 +126,51 @@ async def run_test(
...
-def setup_fetch_files_responses(aioresponses, file_contents):
-    for fc in file_contents:
-        setup_fetch_files_response(aioresponses, 200, fc)
+def get_file_listing_payload(paths):
+    def make_entry(path):
+        parts = path.split("/", 1)
+        type_ = "blob" if len(parts) == 1 else "tree"
+        obj = {}
+        if type_ == "tree":
+            # this obviously does not handle multiple files in the same directory
```
ahal: I'm not sure "obviously" applies here ><

landoscript/docker.d/init_worker.sh
```diff
@@ -1,2 +1,20 @@
...
+test_var_set() {
+  local varname=$1
+
+  if [[ -z "${!varname}" ]]; then
+    echo "error: ${varname} is not set"
+    exit 1
+  fi
+}
+
+# TODO: real URLs
```
ahal, line 13: Comment just to help remember this.

ahal (changes requested):
Looking good! Did a first pass

landoscript/docker.d/worker.yml
bhearsum: Thank you! Feel free to structure this however you want; I don't mind needing to do a bit of parsing on our side as long the information is there, and unambigious (as it is now).

On a related note: is there going to be a `central` phab short name? Or is `mozilla-central` going to go by `main` in the new world?

scriptworker_client/src/scriptworker_client/github_client.py
bhearsum: I did look at https://docs.github.com/en/rest/git/trees?apiVersion=2022-11-28#get-a-tree, yeah. It has a noted limit on the maximum response size (similar to the GraphQL API), and because of that I didn't bother pursuing it further, as we most likely would've needed to recursively fetch through it as well, and I thought it was nicer to avoid needing an HTTP and GraphQL client.

bhearsum: (also, I should've mentioned this upfront: this part of the change is in its own pull request over at https://github.com/mozilla-releng/scriptworker-scripts/pull/1161)

landoscript/src/landoscript/actions/version_bump.py
bhearsum: I don't recall now...my only goal is to make sure newlines are rendered. That can probably be done without the splitting though!

landoscript/src/landoscript/lando.py
bhearsum: Yeah, polling until we timeout is what is happening here, and that's a really good point you've made.

And to be honest, I'm not actually sure what the best thing to do here is. I think in the short term, attempting to cancel the request at shutdown for any non-success makes sense. But this is also less than ideal if, eg: we get spot terminated and have to redo the work because of it. There's also no guarantees that we'll be able to cancel the request in time in the event of a spot termination.

Another obvious option (for the future) would be to take the notarization approach, and do the polling in a separate task. But I also wonder if we even need to bother to poll in some cases? For example, l10n bumps don't have anything downstream of them...maybe we could just spit out the status URL, and leave it at that for cases like that?

landoscript/src/landoscript/script.py
bhearsum: I think changing this would be a larger effort around improving types in scriptworker (the problematic class being referenced here is https://github.com/mozilla-releng/scriptworker/blob/bba07b47036bfa4870741ba54c6d5e98e91134f3/src/scriptworker/context.py#L57-L64, and I haven't looked recently, but I think there's assumptions in tests over there as well around certain things being nullable that don't really make sense in reality).

landoscript/src/landoscript/script.py
bhearsum: If we want to do it for consistency I could be convinced, but my personal view here is that it's weird and surprising that scopes shouldn't affect behaviour. (Sorry...I should've called this out upfront, as it is different than what we do in at least some other scriptworkers.)

landoscript/src/landoscript/script.py
bhearsum: I think this is a good idea...it will require the `run` functions to all have the same interface though. This is easy enough when it comes to the return values, but as things stand now the inputs vary. (Obviously I could just pass the entire `config` rather than the specific bits needed, but that harms readability in the `run` functions IMO.)

landoscript/src/landoscript/actions/android_l10n_import.py
bhearsum: Ah, excellent!

landoscript/src/landoscript/actions/android_l10n_import.py
bhearsum: For checking for None here? Yeah, you're right, that would be better. (We still need this loop to pull `dest_path` for each toml file though, of course, but I don't think you were suggesting otherwise?)

landoscript/src/landoscript/script.py
bhearsum: (Oh, you touched on this in a later comment...whoops!)

scriptworker_client/src/scriptworker_client/github_client.py
bhearsum: I do recall now doing some experiments with the HTTP API as well, and finding that it may have also been slower, in addition to being just as complex...although that is a weak memory at this point.

scriptworker_client/src/scriptworker_client/github_client.py
bhearsum: Yeah, that's fair. At the very least it needs to be prefixed; but I don't see any reason it can't go on the class.

scriptworker_client/src/scriptworker_client/github_client.py
bhearsum: I addressed this and the other comments about this commit in https://github.com/mozilla-releng/scriptworker-scripts/pull/1161.

landoscript/src/landoscript/actions/version_bump.py
bhearsum: Oh, I recall why we have this now: it's to ensure that we get one log line per line in the file, while ensuring each line is prefixed. (Arguably we could simply log with a leading newline to achieve a similar effect...)

scriptworker_client/src/scriptworker_client/github_client.py
bhearsum: I addressed this and the other comments about this commit in https://github.com/mozilla-releng/scriptworker-scripts/pull/1161.

landoscript/src/landoscript/lando.py
jcristau: Spot termination is not a concern for scriptworkers.  We might get killed if we exceed terminationGracePeriodSeconds though, in which case we should still aim to preserve idempotence.  It looks like we only ever have a single lando request, which hopefully is handled atomically at the other end (TIL git push has a `--atomic` option, nice, I'm only 9 years late to notice)?  (I was wondering how this worked for merge day.  Currently we have 2 hg pushes, e.g. once to tag central and another to merge to beta.  I guess that goes away because we know they're the same underlying git repo and so it doesn't matter where the tag goes?)

We could try and cancel the submission on shutdown (bitrisescript seems to do something similar: https://github.com/mozilla-releng/scriptworker-scripts/blob/9867329bf5b2956a90bd990ded424473c42fc2f2/bitrisescript/src/bitrisescript/script.py#L47), and there should be a couple of minutes between SIGTERM and SIGKILL.  I would be a bit worried about races between the attempt to cancel and the request landing (or maybe the cancel attempt itself times out, or ...).

So there's a few scenarios that could happen AFAICT:
1. maybe the initial request didn't go through, we can rerun everything, all good
2. maybe the initial request failed, or was canceled, we can safely retry.
3. maybe the initial request succeeded before the rerun, we should detect that and not do anything
4. maybe the initial request is still in flight.  We could potentially find it and poll on it, or cancel it and retry?  If we can't find it, we may have 2 requests in flight, hopefully only one can succeed?

In treescript we also had the case where one push went through but not the other, and the case where when the next run checked version numbers, the first push had gone through to hgssh but was not yet mirrored out to hgweb.

bhearsum: Thanks for the additional thoughts and strategies, @jcristau! Not needing to worry about spot termination certainly simplifies things, as we can make some safer assumptions.

I'm going to give this a bit more thought - but clearly we need some sort of special handling for the rerun scenario, if not separating polling altogether.

> In treescript we also had the case where one push went through but not the other, and the case where when the next run checked version numbers, the first push had gone through to hgssh but was not yet mirrored out to hgweb.

If we're talking about two pushes to the same repository, this case will go away AFAIK - all changes to a given repository (which in the new world means branch within the new github repo) will either happen or not.

landoscript/src/landoscript/actions/android_l10n_sync.py
bhearsum: Yeah, there's no reason these have to be TypedDicts - I've converted them. (It turns out TypedDict handles nesting a bit more gracefully, but dataclasses deal with required/optional/defaults much better IMO, so it's still preferable.)

landoscript/docker.d/worker.yml
cgsheeh: Okay, I removed the `main.` part from the scm level, it will now return a string like `"scm_level_3"`. I'll keep the `repo_url` field the same and let you handle parsing the `owner` and `repo` values out of that field. :)

> On a related note: is there going to be a central phab short name? Or is mozilla-central going to go by main in the new world?

`mozilla-central` will go by `main` in the new world, and the others repos will use names that match their bookmarks in mozilla unified (ie `mozilla-beta` is the `beta` branch, `mozilla-esr128` is the `esr128` branch).

bhearsum: Cheers!

landoscript/src/landoscript/actions/version_bump.py
bhearsum: Just for clarity: I plan to leave this as is unless you feel it's preferable to log with a preceding newline.

landoscript/src/landoscript/lando.py
bhearsum: >  Currently we have 2 hg pushes, e.g. once to tag central and another to merge to beta. I guess that goes away because we know they're the same underlying git repo and so it doesn't matter where the tag goes?)

Yeah, that's right. Although that does bring to mind something a bit odd about the API: `tag` actions are applied to a repository+branch - not to a repository overall. @cgsheeh - is this something you've thought about? (I don't think we necessarily need to change anything here, I'm just noting it because it's a bit odd.)

bhearsum: The more I think about how to handle termination, the more it seems like moving polling out of the landoscript tasks might be the best thing to do. Something like:
* Return immediately from landoscript after a successful submission; attaching the status url in a well enough artifact
* Add a new task downstream of each landoscript task which does the polling. This doesn't need to be a scriptworker task AFAIK, because polling doesn't require credentials.

It would also avoid the need for any cancellation in landoscript, and thus any race conditions associated with that. If we trust that a landoscript task would never be rerun unless a human has inspected the state of things first, this would eliminate the need for special rerun handling in landoscript, I think. (Although perhaps it's too much a leap to assume it would never be errantly rerun, and we'd want at least some of the rerun safety checks?)

The main downside here is that if the polling task finds that the landoscript submission has failed, we'd need to rerun the upstream landoscript task, and then force a new polling task afterwards.

scriptworker_client/src/scriptworker_client/github_client.py
ahal: > I thought it was nicer to avoid needing an HTTP and GraphQL client.

Fyi the simple-github client supports both. IIRC they even share an aiohttp session. Not that that changes anything here :)

ahal (approved):
Thanks, lgtm!

bhearsum:
To help make additional changes easier to review I'm going to merge this initial patch. There are still a few things to address from comments here (most notably - how to handle unfinished lando submissions) that will come in follow-up PRs next week. Thanks to everyone who reviewed this giant patch so far!

landoscript/docker.d/worker.yml
bhearsum: I'm getting rid of the `central` references in https://github.com/mozilla-releng/scriptworker-scripts/pull/1164.

landoscript/src/landoscript/actions/android_l10n_import.py
bhearsum: This is getting handled in https://github.com/mozilla-releng/scriptworker-scripts/pull/1165

landoscript/src/landoscript/actions/android_l10n_sync.py
bhearsum: This is getting handled in https://github.com/mozilla-releng/scriptworker-scripts/pull/1165

landoscript/src/landoscript/actions/android_l10n_import.py
bhearsum: Handling this in https://github.com/mozilla-releng/scriptworker-scripts/pull/1166

landoscript/src/landoscript/lando.py
bhearsum: For posterity, I discovered after writing this that all Lando endpoints (including status polling) require authentication. So at least in the immediate term, moving polling outside of landoscript is not an option. (I guess we could poll in a separate landoscript task, but that probably doesn't provide any benefits...)

PR #1136, opened 2025-02-13
remove unused push_image.sh

bhearsum:
as far as i can tell, this stopped being used in https://github.com/mozilla-releng/scriptworker-scripts/pull/932


jcristau (approved):
Makes sense to me, taskcluster/docker/skopeo/push_image.sh is used instead.

PR #1138, opened 2025-02-24
Bug 1950203 - Add a behavior that notarizes openh264 plugins for macosx

gabrielbusta:
See [Bug 1950203](https://bugzilla.mozilla.org/show_bug.cgi?id=1950203) for details.

### DIFF
--- a/signingscript/src/signingscript/sign.py
+++ b/signingscript/src/signingscript/sign.py
@@ -1604,6 +1604,20 @@ async def apple_notarize_geckodriver(context, path, *args, **kwargs):
 
 
 @time_async_function
+async def apple_notarize_openh264_plugin(context, path, *args, **kwargs):
+    """
+    Notarizes given openh264 plugin using rcodesign.
+    """
+    # Setup workdir.
+    notarization_workdir = os.path.join(context.config["work_dir"], "apple_notarize")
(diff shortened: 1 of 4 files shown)


signingscript/src/signingscript/sign.py
```diff
@@ -1603,6 +1610,19 @@ async def apple_notarize_geckodriver(context, path, *args, **kwargs):
     return await _notarize_geckodriver(context, path, notarization_workdir)
 
 
+@time_async_function
+async def apple_notarize_openh264_plugin(context, path, *args, **kwargs):
+    """
+    Notarizes given geckodriver package using rcodesign.
```
hneiva: `s/geckodriver/openh264`

Also.. should we just move the `_notarize_openh264_plugin` logic inside this function?

hneiva (changes requested):

hneiva (approved):
lgtm

ahal:
I was able to get a task that uses this going on maple:
https://firefox-ci-tc.services.mozilla.com/tasks/FiXB7OjIR7-0m9KeNV0uDg/runs/0/logs/public/logs/live_backing.log

Unfortunately there are no L1 notarization credentials, so the failure is expected and we've tested this as much as we are able. We'll need to just go ahead and deploy to prod and then I can try it again.

PR #1140, opened 2025-02-26
Bug 1950585 - Add hiro-redux

JohanLorenzo:
Blocks https://github.com/mozilla-releng/adhoc-signing/pull/229

### DIFF
--- a/iscript/src/iscript/mac.py
+++ b/iscript/src/iscript/mac.py
@@ -27,7 +27,7 @@ from scriptworker_client.utils import get_artifact_path, makedirs, rm, run_comma
 log = logging.getLogger(__name__)
 
 
-KNOWN_ARTIFACT_PREFIXES = ("public/", "releng/partner/", "private/openh264/")
+KNOWN_ARTIFACT_PREFIXES = ("public/", "releng/partner/", "private/openh264/", "releng/adhoc/")
 
 
 # App {{{1
--- a/iscript/src/iscript/constants.py
+++ b/iscript/src/iscript/constants.py
@@ -29,6 +29,14 @@ MAC_PRODUCT_CONFIG = {
         "create_pkg": True,
         "hardened_runtime_only_files": "geckodriver",
     },
+    "hiro-redux": {
+        "designated_requirements": """=designated => certificate leaf[subject.OU] = "%(subject_ou)s" """,
+        "sign_dirs": ("MacOS", "Resources"),
+        "skip_dirs": tuple(),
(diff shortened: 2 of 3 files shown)


iscript/src/iscript/mac.py
```diff
@@ -1120,9 +1120,9 @@ async def download_entitlements_file(config, sign_config, task):
         None: if not ``sign_config["sign_with_entitlements"]``
 
     """
-    if not sign_config["sign_with_entitlements"]:
+    url = task["payload"].get("entitlements-url")
+    if not sign_config["sign_with_entitlements"] or not url:
```
hneiva: This feels a bit counter-intuitive to me.
If `sign_with_entitlements` is true, I feel like a url should be provided.
If it expects to sign with entitlements but the url is not provided, I think it should error out.

JohanLorenzo: I totally understand. For the record, I quickly hacked that condition to see if I could green up an adhoc signing task. I managed to do so, thus I agree it's time to make things cleaner. 

I looked into the history and actually, not doing anything when `entitelements-url` is missing from the payload used to be the behavior. It was changed in https://github.com/mozilla-releng/scriptworker-scripts/commit/ee1d457134d363190b8ef2ea548a9440cdcf67b0 but I couldn't find more context about why this was changed. 

Back to this PR, I ended up in a situation where: 

1. Hiro Redux doesn't have any entitlements 
2. The `sign_config` is configured at the worker level (and not at the product one) 

https://github.com/mozilla-releng/scriptworker-scripts/blob/2a1d96ea13ad21653a097d7efe6469caeffaf75d/iscript/config_example.yaml#L21 

3. `sign_with_entitlements` cannot be overridden by another config
4. The adhoc worker has `sign_with_entitlements` set to true (for valid reasons) 

So, I'm not sure what the right approach should be. I agree we should error out when we do expect the `entitlements-url`. However, we should likely make this config per product. 

@hneiva What do you think?

JohanLorenzo: Discussed over Element with @hneiva. He recommended to switch the `mac-behavior` to  `mac_sign_hardened` because it's less convoluted and easier to follow. I did so and it yielded a green task: https://firefox-ci-tc.services.mozilla.com/tasks/d-ifanOwRR-nFbM_M9oPMQ/runs/6/logs/public/logs/live_backing.log

Thus, I'm removing controversial if-statement.

hneiva (approved):
LGTM
No harm in merging this in

PR #1141, opened 2025-02-27
support uploading translations artifacts

bhearsum:
The first commit here is a sortof companion to https://github.com/mozilla-releng/scriptworker/pull/644, where wildcard support was added in scriptworker. It adds a helper function to take a beetmover artifact map with wildcards, and turn it into one with those resolved.

The second commit is a fairly straightfoward addition of a new action for uploading the translations artifacts.

I've been testing this in https://github.com/mozilla/translations/pull/590 / https://firefox-ci-tc.services.mozilla.com/tasks/groups/c-dILkwwTA-SxTdTZ22oLA. The beetmover parts of that are all working well (the failures are intermittent issues in the pipeline).

### DIFF
--- a/beetmoverscript/src/beetmoverscript/data/upload_translations_artifacts_task_schema.json
+++ b/beetmoverscript/src/beetmoverscript/data/upload_translations_artifacts_task_schema.json
@@ -1,0 +1,97 @@
+{
+  "title": "Taskcluster beetmover upload translations artifacts task schema",
+  "type": "object",
+  "properties": {
+    "dependencies": {
+      "type": "array",
+      "minItems": 1,
+      "uniqueItems": true,
+      "items": {
+        "type": "string"
+      }
+    },
+    "payload": {
+      "type": "object",
+      "properties": {
+        "releaseProperties": {
+            "type": "object",
+            "properties": {
+                "appName": {
+                    "type": "string"
+                }
+            },
+            "required": [
+                "appName"
+            ],
+            "additionalProperties": false
+        },
+        "upstreamArtifacts": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "additionalProperties": false,
+                "properties": {
+                    "taskType": {
+                        "type": "string"
+                    },
+                    "taskId": {
+                        "type": "string"
+                    },
+                    "paths": {
+                        "type": "array",
+                        "minItems": 1,
+                        "uniqueItems": true,
+                        "items": {
+                            "type": "string"
+                        }
+                    },
+                    "optional": {
+                        "type": "boolean"
+                    }
+                },
+                "required": ["taskId", "taskType", "paths"]
+            },
+            "minItems": 1,
+            "uniqueItems": true
+        },
+        "artifactMap": {
+            "type": "array",
+            "items": {
+                "type": "object",
+                "additionalProperties": false,
+                "properties": {
+                    "taskId": {
+                        "type": "string"
+                    },
+                    "paths": {
+                        "type": "object",
+                        "minProperties": 1,
+                        "additionalProperties": {
+                            "type": "object",
+                            "additionalProperties": false,
+                            "properties": {
+                                "destinations": {
+                                    "type": "array",
+                                    "minItems": 1,
+                                    "items": {
+                                        "type": "string"
+                                    }
+                                }
+                            }
+                        }
+                    }
+                },
+                "required": [
+                    "taskId",
+                    "paths"
+                ]
+            },
+            "minItems": 1,
+            "uniqueItems": true
+        }
+      },
+      "required": ["releaseProperties", "upstreamArtifacts", "artifactMap"]
+    }
+  },
+  "required": ["payload", "dependencies"]
+}
--- a/beetmoverscript/src/beetmoverscript/script.py
+++ b/beetmoverscript/src/beetmoverscript/script.py
@@ -3,11 +3,16 @@
 
 import asyncio
 import base64
+import copy
+import fnmatch
+import json
 import logging
 import mimetypes
 import os
+import os.path
 import re
 import sys
+from collections import defaultdict
 from io import BytesIO
 from multiprocessing.pool import ThreadPool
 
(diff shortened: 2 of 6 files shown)


beetmoverscript/src/beetmoverscript/data/upload_translations_artifacts_task_schema.json
```diff
@@ -0,0 +1,91 @@
...
+                        "properties": {
+                            "destinations": {
+                                "type": "array",
+                                "minItems": 1,
+                                "items": {
+                                    "type": "string"
+                                }
+                            }
+                        }
+                    }
```
jcristau: This looks wrong?
- minItems is for arrays not objects afaict
- "destinations" is not a property of the "paths" object, there's another thing in between

I think this should more or less mirror artifactMap_beetmover_task_schema.json's `artifactMap`?

beetmoverscript/src/beetmoverscript/data/upload_translations_artifacts_task_schema.json
```diff
@@ -0,0 +1,91 @@
...
+                    "taskId",
+                    "paths"
+                ]
+            },
+            "minItems": 1,
+            "uniqueItems": true
+        },
+        "maxRunTime": {
+          "type": "number"
+        }
```
jcristau: nit: we don't do anything with maxRunTime, and we also don't set additionalProperties: false, so this is kind of useless (I know we have it elsewhere though)

jcristau: (we should probably have additionalProperties disabled where possible though...)

jcristau (changes requested):

beetmoverscript/src/beetmoverscript/data/upload_translations_artifacts_task_schema.json
bhearsum: I'm not sure how this got so out of whack; I'm guessing I was playing with a different format at some point, and didn't come back around to fix this up. (Obviously the schema here doesn't match what the code expects...)

bhearsum:
Also included in the schema is support for the `optional` attribute, which is needed to satisfy scriptworker...

beetmoverscript/src/beetmoverscript/data/upload_translations_artifacts_task_schema.json
```diff
@@ -0,0 +1,96 @@
...
+            "type": "array",
+            "items": {
+                "type": "object",
+                "additionalProperties": false,
+                "properties": {
+                    "taskId": {
+                        "type": "string"
+                    },
+                    "paths": {
+                        "type": "object",
```
jcristau, line 67: nit: we could probably have `minProperties: 1` on the paths object

beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -285,6 +290,208 @@ async def upload_data(context):
         await retry_data_upload(context, dataUpload["destinations"], data, dataUpload["contentType"])
 
 
+def get_concrete_artifact_map_from_globbed(upstreamArtifactPaths, artifactMap, strip_segments=["public/build", "public/logs"]):
```
jcristau: don't use a mutable type as default argument, e.g. replace the list with a tuple

beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -285,6 +290,208 @@ async def upload_data(context):
...
+    """Given a list of artifacts from tasks (`upstreamArtifactPaths`) and a
+     mapping of names and/or patterns to destinations (`artifactMap`), return a
+     mapping that maps all of the specific artifacts from tasks int
+     destinations. Artifacts will have `strip_segments` removed from them before
+     being compared against `artifactMap` patterns.
```
jcristau: The strip_segments stuff seems a bit magic/surprising.  Could we avoid it, or pull it from the payload maybe?
Also AIUI strip_segments are removed from the eventual destination path, not just for comparison with the patterns?

beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -285,6 +290,208 @@ async def upload_data(context):
...
+                # We need to look at non-'*' paths separate from '*' paths.
+
+                for input_path, output in other_paths.items():
+                    # Skip any input paths that don't match the artifact name.
+                    if "*" in input_path:
+                        if not fnmatch.fnmatch(dest_artifact, input_path):
+                            continue
+                    else:
+                        if input_path != artifact:
+                            continue
```
jcristau: Is there a reason we need to treat globs and non-globs differently here, where one is compared with `artifact` and the other with `dest_artifact`?  Couldn't we just always use fnmatch against either (not sure which makes more sense)

beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -285,6 +290,208 @@ async def upload_data(context):
...
+                    destinations.extend(full_glob_destinations)
+
+                # If there are destinations, create a fully concrete entry,
+                # mapping the full path to the file on disk to destinations
+                # it belongs, with any requested segments removed.
+                if destinations:
+                    concretePaths[artifact] = {"destinations": []}
+                    for d in destinations:
+                        if not d.endswith(dest_artifact):
+                            d = os.path.normpath(f"{d}{dest_artifact}")
```
jcristau: Should this require that `d` ends in `/`?  Or join the paths?  Or are there cases where the paths do need to be glued together like that?

beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -285,6 +290,208 @@ async def upload_data(context):
...
+
+    return concreteArtifactMap
+
+
+def remove_segments(string, prefixes):
+    for p in prefixes:
+        if p in string:
+            string = string[string.find(p) + len(p) :]
+
+    return string
```
jcristau: This doesn't seem to do what it says on the tin, i.e. the "prefixes" aren't treated as prefixes, but as "if this shows up anywhere in `string`, drop it and anything before it".

jcristau (changes requested):
I've got a couple of questions on the shape of the API.

I'll also note that AIUI we assume that writes anywhere in the destination bucket are legitimate for that task, or that they're prevented by permissions on the bucket? (I have no objection to that just wanted to make sure it's explicit)

beetmoverscript/src/beetmoverscript/script.py
bhearsum: We probably could come to think of it; I'll give it a try.

bhearsum:
> I'll also note that AIUI we assume that writes anywhere in the destination bucket are legitimate for that task, or that they're prevented by permissions on the bucket? (I have no objection to that just wanted to make sure it's explicit)

That's correct. Theoretically, we could do something more fine grained (eg: only allow certain tasks to write to `models`), but in practice anyone who has scopes for one of these has scopes for all, so there's no security benefit to doing so AFAICT.

beetmoverscript/src/beetmoverscript/script.py
bhearsum: That's all correct - yes.

I don't have any objection to having it in the payload, seeing as this is not a translations-specific method. I think that providing a default if it's not there still might be sensible though (I imagine that most future use cases of this are likely to have the same prefixes on their artifacts). I could be convinced otherwise though.

beetmoverscript/src/beetmoverscript/script.py
bhearsum: > Should this require that `d` ends in `/`? Or join the paths?

Yes - I think that's sensible.

> Or are there cases where the paths do need to be glued together like that?

I'm not aware of any cases like this. I guess theoretically there could be some use case where you'd want to apply a common prefix to all matching filenames (instead of a common leading directory), but we can cross that bridge if we ever come to it.

beetmoverscript/src/beetmoverscript/script.py
bhearsum: You're right. The original version of this did actually only deal in prefixes; I realized later that we had to deal with absolute paths, and ended up renaming most things. At the very least, I can adjust the argument's name. Ideally there would be a less ugly/hacky way to do this, but I haven't come up with one yet :(.

bhearsum: I made a change which I think is an improvement: it now strips prefixes, but gains some smarts to always strip away the scriptworker-specific leading directories. Let me know what you think.

beetmoverscript/src/beetmoverscript/script.py
bhearsum: I ended up moving this up a level, and hardcoding the prefixes only for translations. (I don't see a need to override this in the payload for it at the moment, so I don't particularly want to bother supporting that...). Happy to discuss this more if needed.

beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -285,6 +290,206 @@ async def upload_data(context):
...
+    if errors:
+        raise ScriptWorkerTaskException(*errors)
+
+    return concreteArtifactMap
+
+
+def remove_prefixes(string, prefixes):
+    for p in prefixes:
+        if string.startswith(p):
+            string = string[len(p):]
```
jcristau: nit: could return here unless there's a case where we want to strip multiple prefixes from the same path?

beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -285,6 +290,206 @@ async def upload_data(context):
...
+            if map_["taskId"] != taskId:
+                continue
+
+            for artifact in artifacts:
+                # the artifact path with the scriptworker-specific leading
+                # directories removed. ie: in the same form it is in the task
+                # it was pulled from
+                leading_dir = os.path.join(work_dir, "cot", taskId) + "/"
+                if not artifact.startswith(leading_dir):
+                    raise ScriptWorkerTaskException("cannot determine relative artifact path")
```
jcristau: suggestion: include `artifact` in the exception message so it's easier to figure out if this ever happens?

beetmoverscript/src/beetmoverscript/task.py
```diff
@@ -30,6 +30,8 @@ def get_schema_key_by_action(context):
         return "import_from_gcs_to_artifact_registry_schema_file"
     elif utils.is_upload_data_action(action):
         return "upload_data_schema_file"
+    elif utils.is_upload_translations_artifacts_action(action):
+        return "upload_translations_artifacts"
```
jcristau, line 34: Should this action only be available for the translations cot product?  If not, 1) the name is going to be confusing, 2) what are the implications of that payload shape being allowed in, say, firefox?  (And now I think about it, the same applies to upload-data...)

beetmoverscript/src/beetmoverscript/script.py
bhearsum: Good idea; will do.

beetmoverscript/src/beetmoverscript/task.py
bhearsum: Yeah, making this exclusive to the translations production makes sense to me.

bhearsum: Sigh...so we have no access to the cot product in the script itself (only the `scriptworker` process has it in the env/config). Perhaps this is guarded enough by the `upload-translations-artifacts` scope? I can wire this in further if you like though.

jcristau: It's fine, thanks.

beetmoverscript/src/beetmoverscript/script.py
```diff
@@ -285,6 +291,222 @@ async def upload_data(context):
...
+                    raise ScriptWorkerTaskException(f"cannot determine relative artifact path for {artifact}")
+
+                relative_artifact = artifact[len(leading_dir) :]
+
+                destinations = []
+                # We need to look at non-'*' paths separate from '*' paths.
+
+                for input_path, output in other_paths.items():
+                    # Skip any input paths that don't match the artifact name.
+                    if not fnmatch.fnmatch(artifact, input_path):
```
jcristau: Should this look at relative_artifact?

jcristau (approved):

beetmoverscript/src/beetmoverscript/script.py
bhearsum: Yes, I think it should. (And this also highlights that the non-glob entries in the tests shouldn't be full paths, as they come from task definitions which would have to use the relative paths...)

PR #1142, opened 2025-02-27
feat(beetmover): allow not failing on unknown mimetypes

bhearsum:
This is needed to support work from https://github.com/mozilla-releng/scriptworker-scripts/pull/1141 where we're uploading all artifacts from upstream tasks, and can't reasonably predict all possible file types and extensions.

This also fixes a minor bug in the `upload_to_s3` tests, where they implicitly rely on `setup_mimetypes()` being called elsewhere to pass (which means they fail mysteriously if run independently).

I considered making this the default behaviour, but ultimately decided against it because we most likely want to notice and fix mime types for files from other projects, rather than have them uploaded so generically.

### DIFF
--- a/beetmoverscript/src/beetmoverscript/gcloud.py
+++ b/beetmoverscript/src/beetmoverscript/gcloud.py
@@ -102,11 +102,14 @@ def setup_gcs_credentials(raw_creds):
     os.environ["GOOGLE_APPLICATION_CREDENTIALS"] = fp.name
 
 
-async def upload_to_gcs(context, target_path, path, expiry=None):
+async def upload_to_gcs(context, target_path, path, expiry=None, fail_on_unknown_mimetype=True):
     product = get_product_name(context.task, context.config)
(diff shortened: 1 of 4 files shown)


jcristau (changes requested):
This seems incomplete.  I think we want this to come with a way for `retry_upload` to be called with that new argument?

Also, should we explicitly set content-type to application/octet-stream if we can't guess, instead of relying on whatever the cloud service does with it?

bhearsum:
> This seems incomplete. I think we want this to come with a way for `retry_upload` to be called with that new argument?

Well, actually....the work I'm doing in #1141 uses `retry_upload` directly, because its existing callers can't reasonable be augmented to work for that use case. I can wire this up further if necessary, but there won't be a use case for it at the moment.
 
> Also, should we explicitly set content-type to application/octet-stream if we can't guess, instead of relying on whatever the cloud service does with it?

This seems fine and sensible to me.

jcristau (approved):

PR #1143, opened 2025-02-28
feat(bitrise): Download workflow metadata after work is finished

hneiva:
This data will come in handy for subsequent workflows when reusing build artifacts in bitrise.

### DIFF
--- a/bitrisescript/src/bitrisescript/bitrise.py
+++ b/bitrisescript/src/bitrisescript/bitrise.py
@@ -185,6 +185,23 @@ async def download_log(build_slug: str, artifacts_dir: str, poll_interval: int =
         log.error(f"Bitrise has no log for build '{build_slug}'. Please check https://app.bitrise.io/build/{build_slug}")
 
 
+async def download_workflow_info(build_slug: str, target_file: str) -> None:
+    """Download and save a workflow json.
+
+    Args:
+        build_slug (str): The Bitrise build slug id
+        target_file (str): The artifacts path
+    """
+    client = BitriseClient()
+    endpoint = f"/builds/{build_slug}"
+
+    workflow_data = await client.request(endpoint)
+    os.makedirs(os.path.dirname(target_file), exist_ok=True)
+    with open(target_file, "wt") as fd:
+        # Save the workflow data as pretty-printed json
+        fd.writelines(pformat(workflow_data))
+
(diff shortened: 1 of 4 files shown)


bitrisescript/tests/test_script.py
```diff
@@ -35,15 +36,18 @@ async def test_async_main(mocker, config, task, expectation, expected_num_future
     client_mock.close.return_value.set_result(None)
     client_mock.set_app_prefix.return_value.set_result(None)
     mocker.patch("bitrisescript.script.BitriseClient", return_value=client_mock)
+    mocker.patch("bitrisescript.script.get_running_builds", return_value=mocker.AsyncMock())
+    mocker.patch("bitrisescript.script.find_running_build", return_value=mocker.Mock())
+    mocker.patch("bitrisescript.script.wait_and_download_workflow_log", return_value=mocker.AsyncMock())
+    mocker.patch("bitrisescript.script.run_build", return_value=mocker.AsyncMock())
+
+    # Mock signal handler
+    mocker.patch("bitrisescript.script.asyncio.get_event_loop", return_value=mocker.Mock())
```
Eijebong, line 45: I'm curious, why is this needed now?

hneiva: Because pytest and event loops are a major pain to deal with.
While adding the tests I kept getting an error saying the "event loops is closed". It's usually related to scoping fixtures, but in this case mocking out the signal handler seemed to fix it for me.

bitrisescript/src/bitrisescript/bitrise.py
```diff
@@ -185,6 +185,24 @@ async def download_log(build_slug: str, artifacts_dir: str, poll_interval: int =
...
+
+    Args:
+        build_slug (str): The Bitrise build slug id
+        target_file (str): The artifacts path
+    """
+    client = BitriseClient()
+    endpoint = f"/builds/{build_slug}"
+
+    workflow_data = await client.request(endpoint)
+    if not os.path.isdir(os.path.dirname(target_file)):
```
Eijebong: nit: `os.makedirs(os.path.dirname(target_file), exist_ok=True)`

https://docs.python.org/3/library/os.html#os.makedirs

Eijebong (approved):

PR #1147, opened 2025-03-18
move GithubClient to scriptworker_client; add support for slashes in key names

bhearsum:
A few things going on here, mostly boring:
* Move the existing GithubClient to scriptworker_client
* Move the associated test fixtures to a new `pytest-scriptworker-client` package
* Necessary adjustments in treescript to let it be able to find and use the new packages

Although not ideal, this also includes a minor fix to allow slashes in GraphQL key names that I had already incorporated into the version of this code in https://github.com/mozilla-releng/scriptworker-scripts/pull/1135. I can pull it out if necessary, but I think it's probably small and safe enough to just keep in here?

This is all motivated by the fact that [Landoscript needs a GithubClient](https://github.com/mozilla-releng/scriptworker-scripts/pull/1135), and we ought to share the existing one between it and treescript.


ahal (approved):
Thanks, lgtm!

PR #1149, opened 2025-03-25
Add new gpg signing subkey (bug 1947765)

jcristau:


bhearsum (approved):
It looks like your plan is to land this now, and then flip https://github.com/mozilla-releng/scriptworker-scripts/blob/df364bcbae8c4ad40eaa3a8deb21d160f9a943f5/signingscript/docker.d/passwords.yml#L493C17-L493C48 over to the new keyid after the Autograph side is done? (Unless they're going to re-use the existing key id with new key material, in which case this needs to be landed and deployed before that happens AIUI.)

jcristau:
Right, planning to get a new autograph keyid for the new subkey.
The sequence as I see it will look like:
- [x] land/deploy this
- [x] announce the upcoming switch
- [x] (optionally?) switch adhoc to the new signer; sign something for verification
- [x] wait a few days
- [x] switch nightly-signing to the new signer
- [x] switch release-signing to the new signer

PR #1153, opened 2025-04-07
chore(signingscript): remove AWS signing formats

bhearsum:
We're now using Autograph GCP prod everywhere except ESR. Once we start using it on ESRs (this week, hopefully), we can remove the old AWS formats.


ahal (approved):
Sweet.

PR #1155, opened 2025-04-08
chore: bump dependencies

bhearsum:
The only major bumps here are taskcluster, which is always safe, and mercurial. AFAICT the latter is mainly for changes to its build system, which are unlikely to be relevant to our usage in `robustcheckout`: https://wiki.mercurial-scm.org/Release7.0


jcristau:
mercurial doesn't actually use semver AFAIK, 7.0 is just what comes after 6.9.  But indeed there were large build system changes this time around.

jcristau (approved):

PR #1157, opened 2025-04-10
Add support for release bodies to githubscript

Eijebong:
Github generates release notes that are less than idea (basically `HEAD commit message`). This adds a `releaseBody` parameter to the payload for githubscript to allow tasks to override the generated one.

I made the parameter optional as to not break anything currently using this, if it's not provided then it will continue using the automatically generated release body.

Github also provides a `generate_release_notes` parameter which would append better release notes using the commit list automatically. I chose against using that because it doesn't work well with the assumption that you should be able to rerun this task and have the release be the same (the body would always be different than what's expected since it would be generated by github, not by the task)


Eijebong:
Working task:

https://firefox-ci-tc.services.mozilla.com/tasks/C9qBEsoJRdGEWaOzWRPKww created 
https://github.com/mozilla-mobile/staging-firefox-ios/releases/tag/v139.0

hneiva (approved):
LGTM

PR #1158, opened 2025-04-11
Add firefox iOS repos to treescript known repos

Eijebong:


jcristau (commented):
Sort of related: we should probably add treescript scopes to `cot_restricted_scopes` in scriptworker... it seems to be missing for mobile and comm trust domains :/

treescript/src/treescript/github/versionmanip.py
```diff
@@ -23,8 +23,7 @@
 _VERSION_CLASS_PER_BEGINNING_OF_PATH = {}
 
 _VERSION_CLASS_PER_END_OF_SOURCE_REPO = {
-    # previously, this mapped mobile github repos to the MobileVersion; now there are no remaining
-    # repos on github requiring version bumps
+    "firefox-ios": MobileVersion,
```
Eijebong: iOS doesn't use the same version scheme as other mobile projects.

Instead of bumping `139.0` -> `139.0.1`, they go `139.0` -> `139.1`. Need to figure out if mozilla-versions has a scheme that works or if we need to create one

hneiva:
> Sort of related: we should probably add treescript scopes to `cot_restricted_scopes` in scriptworker... it seems to be missing for mobile and comm trust domains :/

created https://github.com/mozilla-releng/scriptworker/pull/693

treescript/src/treescript/github/versionmanip.py
Eijebong: https://github.com/mozilla-releng/mozilla-version/pull/134

hneiva (approved):

hneiva (approved):

PR #1160, opened 2025-04-11
fix(scriptworker_client): gracefully handle missing files in github_client.get_files

bhearsum:
At the moment we end up throwing an exception if a file is missing, which means we don't get the contents of _any_ of the files fetched, which is probably not what we want.
    
This patch changes things such that missing files will still have a key in the returned dict, but with a None value. (This allows for distinguishing between empty files and missing files.)

Also included here is a minor improvement to handle dashes in identifiers, which was too annoying too split out into its own pull request.

### DIFF
--- a/scriptworker_client/src/scriptworker_client/github_client.py
+++ b/scriptworker_client/src/scriptworker_client/github_client.py
@@ -87,7 +87,7 @@ class GithubClient:
 
         await retry_async(_execute, attempts=3, retry_exceptions=(TransportQueryError,), sleeptime_kwargs={"delay_factor": 0})
 
-    async def get_files(self, files: Union[str, List[str]], branch: Optional[str] = None) -> Dict[str, str]:
+    async def get_files(self, files: Union[str, List[str]], branch: Optional[str] = None) -> Dict[str, Union[str, None]]:
         """Get the contents of the specified files.
 
         Args:
@@ -137,7 +137,15 @@ class GithubClient:
         str_query = query.substitute(owner=self.owner, repo=self.repo, fields=",".join(fields))
 
         contents = (await self._client.execute(str_query))["repository"]
-        return {k.replace(sentinel_dot, ".").replace(sentinel_slash, "/").replace(sentinel_dash, "-"): v["text"] for k, v in contents.items()}
+        ret: Dict[str, Union[str, None]] = {}
+        for k, v in contents.items():
+            if v is None:
+                ret[k] = None
+            else:
+                subbed_k = k.replace(sentinel_dot, ".").replace(sentinel_slash, "/").replace(sentinel_dash, "-")
+                ret[subbed_k] = v["text"]
+
+        return ret
 
     async def get_branch_head_oid(self, branch: str) -> str:
         """Get the revision of the tip of the given branch.
--- a/scriptworker_client/src/scriptworker_client/github_client.py
+++ b/scriptworker_client/src/scriptworker_client/github_client.py
@@ -103,9 +103,10 @@ class GithubClient:
         if isinstance(files, str):
             files = [files]
 
-        # Periods and slashes are not legal GraphQL key names.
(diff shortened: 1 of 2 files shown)


scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -106,6 +106,7 @@ async def get_files(self, files: Union[str, List[str]], branch: Optional[str] =
         # Periods and slashes are not legal GraphQL key names.
         sentinel_dot = "__dot__"
         sentinel_slash = "__slash__"
+        sentinel_dash = "__dash__"
```
ahal, line 109: I think this is fine for now, but it's starting to get into "turn these variables into a dict and iterate over them" territory.

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -106,6 +106,7 @@ async def get_files(self, files: Union[str, List[str]], branch: Optional[str] =
         # Periods and slashes are not legal GraphQL key names.
```
ahal: Maybe edit this comment to say "Certain characters"

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -87,7 +87,7 @@ async def _execute():
 
         await retry_async(_execute, attempts=3, retry_exceptions=(TransportQueryError,), sleeptime_kwargs={"delay_factor": 0})
 
-    async def get_files(self, files: Union[str, List[str]], branch: Optional[str] = None) -> Dict[str, str]:
+    async def get_files(self, files: Union[str, List[str]], branch: Optional[str] = None) -> Dict[str, Union[str, None]]:
```
ahal, line 90: nit: if I'm not mistaken, we should be using Python 3.11 for scriptworkers now. Which means that `str | None` is valid syntax. Though I guess it's best to be consistent and I see `Union` used elsewhere.

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -137,7 +137,15 @@ async def get_files(self, files: Union[str, List[str]], branch: Optional[str] =
         str_query = query.substitute(owner=self.owner, repo=self.repo, fields=",".join(fields))
 
         contents = (await self._client.execute(str_query))["repository"]
-        return {k.replace(sentinel_dot, ".").replace(sentinel_slash, "/").replace(sentinel_dash, "-"): v["text"] for k, v in contents.items()}
+        ret: Dict[str, Union[str, None]] = {}
```
ahal, line 140: nit: I'm not opposed to annotating this, but it's not at any interface boundary and type checkers would have no problem inferring the type.. so I think annotating this variable is unnecessary.

ahal (approved):
Thanks! I have a few optional nits, but I'm ok with it landing as is

scriptworker_client/src/scriptworker_client/github_client.py
bhearsum: Hm, I added this because I thought I saw a complaint about it...but maybe I conflated it with a complaint about the return value of the function. I'll see about removing it.

scriptworker_client/src/scriptworker_client/github_client.py
jcristau: iscript (and therefore also scriptworker_client) is still on 3.8 iirc.

scriptworker_client/src/scriptworker_client/github_client.py
bhearsum: Yeah:
```
[task 2025-04-14T14:47:20.837Z] src/scriptworker_client/github_client.py:140: error: Need type annotation for "ret" (hint: "ret: dict[<type>, <type>] = ...")  [var-annotated]
```

PR #1161, opened 2025-04-11
feat(scriptworker_client): add support for remote file listings in github_client

bhearsum:
This adds support for fetching the directory structure of a Github repository or a subdirectory within one, something which will be necessary to support Android l10n updates and imports without requiring a full repository clone. (tl;dr - some of the `l10n.toml` files rely on globbing to find all localized files: https://github.com/mozilla-l10n/android-l10n/blob/main/mozilla-mobile/android-components/l10n.toml).

The GraphQL query here is inspired/copied from https://gist.github.com/MichaelCurrin/6777b91e6374cdb5662b64b8249070ea, but allows us to fetch to a configurable depth per query. This is necessary because for large repos (eg: Gecko, but even https://github.com/mozilla-l10n/android-l10n) we run afoul of the maximum GraphQL response size if we attempt to fetch everything at once.

### DIFF
--- a/scriptworker_client/tests/test_github_client.py
+++ b/scriptworker_client/tests/test_github_client.py
@@ -1,5 +1,6 @@
 import base64
 from pathlib import Path
+from string import Template
 from textwrap import dedent
 
 import pytest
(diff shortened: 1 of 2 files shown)


scriptworker_client/tests/test_github_client.py
```diff
@@ -214,3 +214,284 @@ async def test_get_branch_head_oid_branch_not_found(aioresponses, github_client)
...
+    aioresponses.assert_called()
+
+    key = ("POST", URL(GITHUB_GRAPHQL_ENDPOINT))
+    first_request = aioresponses.requests[key][-2][1]["json"]
+    second_request = aioresponses.requests[key][-1][1]["json"]
+
+    assert first_request == {
+        "query": dedent(
+            f"""
+            query RepoFiles {{
```
ahal, line 323: It might be worth using template strings here, just to avoid the double curly braces everywhere. But not a big deal.

scriptworker_client/tests/test_github_client.py
```diff
@@ -214,3 +214,284 @@ async def test_get_branch_head_oid_branch_not_found(aioresponses, github_client)
 
     with pytest.raises(UnknownBranchError, match="branch 'branchy' not found in repo!"):
         await github_client.get_branch_head_oid("branchy")
+
+
+@pytest.mark.asyncio
+async def test_get_repository_files(aioresponses, github_client):
```
ahal, line 221: Just because you asked for feedback on tests, this seems fine. It's definitely ugly, but I'm not really sure how else to do it.

ahal (approved):
Thanks!

scriptworker_client/tests/test_github_client.py
bhearsum: Yeah, that's a good call.

PR #1162, opened 2025-04-18
feat(landoscript): add schema file to landoscript config

bhearsum:
This ensures that scriptworker will check payload schemas. This is not realistically testable, because we don't run tests through `scriptworker.client.sync_main`.

Also fixes a minor issue with `config_path` (scriptworker wants it to be `None` in order for usage to be printed).

### DIFF
--- a/landoscript/docker.d/worker.yml
+++ b/landoscript/docker.d/worker.yml
@@ -3,3 +3,4 @@ artifact_dir: { "$eval": "ARTIFACTS_DIR" }
 verbose: { "$eval": "VERBOSE == 'true'" }
 lando_api: { "$eval": "LANDO_API" }
 lando_token: { "$eval": "LANDO_TOKEN" }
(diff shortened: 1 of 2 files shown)


landoscript/docker.d/worker.yml
```diff
@@ -3,3 +3,4 @@ artifact_dir: { "$eval": "ARTIFACTS_DIR" }
 verbose: { "$eval": "VERBOSE == 'true'" }
 lando_api: { "$eval": "LANDO_API" }
 lando_token: { "$eval": "LANDO_TOKEN" }
+schema_file: /app/landoscript/src/landoscript/data/landoscript_task_schema.json
```
jcristau, line 6: Isn't this redundant with having it in the default config?

bhearsum: uhh, yes!

PR #1163, opened 2025-04-18
feat(landoscript): respect closed tree for android l10n sync

bhearsum:
I originally thought this was only used for l10n-bump, but it turns out we need to check treestatus for android l10n sync tasks as well.

### DIFF
--- a/landoscript/src/landoscript/script.py
+++ b/landoscript/src/landoscript/script.py
@@ -5,9 +5,8 @@ import aiohttp
 import scriptworker.client
 from scriptworker.exceptions import TaskVerificationError
 
-from landoscript import lando
+from landoscript import lando, treestatus
 from landoscript.actions import android_l10n_import, android_l10n_sync, l10n_bump, merge_day, tag, version_bump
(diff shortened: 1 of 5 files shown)


jcristau (approved):
Do we somewhere test that "approval required" is treated as "open" for our purposes?  (Beta is rarely if ever actually "open")

bhearsum:
> Do we somewhere test that "approval required" is treated as "open" for our purposes? (Beta is rarely if ever actually "open")

What we [do in treescript](https://github.com/mozilla-releng/scriptworker-scripts/blob/ab059644643f954d8dc843bb44b9ae0c8d09e191/treescript/src/treescript/util/treestatus.py) (which I [pulled in here](https://github.com/mozilla-releng/scriptworker-scripts/blob/aa55cedae0e567c698cb43c502fbfe2952908489/landoscript/src/landoscript/treestatus.py) as well), is assume that we can land unless the status is "closed" - so we handle that on our side.

Oh, maybe you're saying that we should have tests that use the `approval required` status though...in which case the answer is no, but I'll add them.

PR #1165, opened 2025-04-18
fix(landoscript): check all android toml files at one time, before pr…

bhearsum:
…ocessing any of them

This is a follow to address something [@ahal called out in the initial review of landoscript](https://github.com/mozilla-releng/scriptworker-scripts/pull/1135#discussion_r2047655019).

### DIFF
--- a/landoscript/tests/test_android_l10n_import.py
+++ b/landoscript/tests/test_android_l10n_import.py
@@ -1,5 +1,7 @@
 import pytest
+from yarl import URL
 
+from landoscript.errors import LandoscriptError
 from landoscript.script import async_main
 from tests.conftest import (
     assert_add_commit_response,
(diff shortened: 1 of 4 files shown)


landoscript/tests/test_android_l10n_import.py
```diff
@@ -242,7 +244,6 @@ async def test_success(
         "android_l10n_import_info": android_l10n_import_info,
     }
     # done here because setup_test sets up github_installation_response too soon...argh
-    from yarl import URL
```
jcristau: Was the comment meant to go with the removed line?  (either way, seems unrelated with this patch?)

landoscript/tests/test_android_l10n_sync.py
```diff
@@ -1,6 +1,7 @@
 import pytest
-from scriptworker_client.github_client import TransportQueryError
+from yarl import URL
```
jcristau: Unrelated changes?

landoscript/tests/test_android_l10n_import.py
bhearsum: The comment is about repeating all of the test set up overall (instead of using `setup_test`), but I can see how it is...not particularly great. I'll improve it.

The import change is unrelated, yeah. It's not harmful, but I'll revert it to keep things clean. (Another follow up I'm planning is to add isort, which will help clean some of these things up.)

landoscript/tests/test_android_l10n_sync.py
bhearsum: Yeah. The removal is removing an unused import, but as you say, unrelated.

jcristau (approved):

PR #1171, opened 2025-04-21
feat(scriptworker_client): replace get_files sentinel values with hashes

bhearsum:
The sentinel values quickly run afoul of the maximum key length of 320 characters for longer file/path names.


hneiva (approved):
Awesome commenting!

LGTM

PR #1172, opened 2025-04-21
fix(landoscript): landoscript tests not running in CI

bhearsum:


bhearsum:
note: we can't trust tests at the moment due to https://bugzilla.mozilla.org/show_bug.cgi?id=1961671, but the tests _are_ indeed passing in CI

ahal (approved):

PR #1177, opened 2025-04-22
fix(landoscript): ensure esr bumps use `esr` suffix where appropriate

bhearsum:
While inspecting some diffs that were spat out in dry run mode, I noticed that `browser/config/version.txt` ended up with an `esr` suffix on it when running `release-to-esr` merge-automation for esr128. As it turns out, this was happening in tests as well, but masked by an incomplete assertion (now fixed).
    
The fix I've implementing here is not great: it's a flag that only really affects anything for the `release-to-esr` merge automation case, but it's the best I could come up with for now. One would think it would be preferable to drop the esr logic in `get_cur_and_next_version` - which indeed fixes this problem, but it it ends up causing issues with the regular `version-bump` on esr: we end up with an esr-suffixed version in `version.txt`, because there's no per-file granularity in the `version_bump` action -- all files take the same version number. (To be quite honest - I think this is a problem in treescript as well, and it's avoided because we always end up in the "Version bumping skipped due to conflicting values" code path: https://github.com/mozilla-releng/scriptworker-scripts/blob/419abe11d20ca5c5a123aaaa5d010f428dff9ca4/treescript/src/treescript/gecko/versionmanip.py#L164C26-L164C75, and end up bumping the version through the `bump-esr` merge automation instead -- but I didn't want to perpepuate the problem here, in case it ends up mattering in the future.)

### DIFF
--- a/landoscript/src/landoscript/actions/merge_day.py
+++ b/landoscript/src/landoscript/actions/merge_day.py
@@ -135,6 +135,7 @@ async def run(github_client: GithubClient, public_artifact_dir: str, merge_info:
                 bump_branch,
                 version_bump_infos,
                 dontbuild=False,
+                munge_next_version=False,
             )
         )
 
--- a/landoscript/src/landoscript/actions/version_bump.py
+++ b/landoscript/src/landoscript/actions/version_bump.py
@@ -1,8 +1,10 @@
 import logging
 import os.path
+import typing
 from dataclasses import dataclass
 
 from gql.transport.exceptions import TransportError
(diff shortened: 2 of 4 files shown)


jcristau:
This is indeed a longstanding issue in treescript, that shows up once a year (bug 1786807, and #349).

bhearsum:
> This is indeed a longstanding issue in treescript, that shows up once a year (bug 1786807, and #349).

Ah, thanks for those pointers! It's great to have some confirmation of this.

PR #1179, opened 2025-04-23
Split out mobile scopes for shipitscript

Eijebong:
When they got added by aaee32e6b68672c5f35d910606de58e38725f84b for some reasons they didn't contain the trust-domain. However the shipit feature in fxci inserts said trust domain in the given scopes.

The firefox android project had those scopes specifically given to it, but since no one else is using shipit in the mobile trust domain atm, let's just make it be coherent with the rest.


jcristau (approved):
makes sense to me

PR #1182, opened 2025-04-23
feat(scriptworker_client): batch get_files requests to avoid timeouts

bhearsum:
While doing real world testing of landoscript I discovered that the android l10n tasks need to pull thousands of files. When send a single `get_files` request with that many files in it, it continually times out.

### DIFF
--- a/scriptworker_client/tests/test_github_client.py
+++ b/scriptworker_client/tests/test_github_client.py
@@ -139,6 +139,60 @@ async def test_get_files(aioresponses, github_client):
 
 
 @pytest.mark.asyncio
+async def test_get_files_multiple_requests(aioresponses, github_client):
+    branch = "main"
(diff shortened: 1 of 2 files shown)


scriptworker_client/tests/test_github_client.py
```diff
@@ -138,6 +138,60 @@ async def test_get_files(aioresponses, github_client):
...
+
+    result = await github_client.get_files(files, branch, files_per_request=1)
+    assert result == expected
+
+    aioresponses.assert_called()
+    key = ("POST", URL(GITHUB_GRAPHQL_ENDPOINT))
+    first_request = aioresponses.requests[key][-2][1]["json"]
+    second_request = aioresponses.requests[key][-1][1]["json"]
+
+    files = list(map(Path, files))
```
jcristau: nit: `files = [Path(file) for file in files]`

jcristau (approved):

PR #1183, opened 2025-04-24
feat(landoscript): support target revisions in tagging, and correct tag action payload

bhearsum:
When I originally implemented the `tag` action I failed to use the right payload. This patch addresses that, including adding support for target revisions.
    
Of note with the latter is that for direct tagging (eg: `release-early-tagging`) we need to accept an hg revision (to avoid taskgraph needing to look up the git revision). This revision + repo url can be used to look up the git revision.
    
For merge day tagging, the revision is not known up front: we use whatever the tip of the to and/or from branches, which we can look-up directly in github.
    
When we move decision tasks over to firing from Github events, we can drop the hg support.

### DIFF
--- a/landoscript/src/landoscript/actions/merge_day.py
+++ b/landoscript/src/landoscript/actions/merge_day.py
@@ -8,6 +8,7 @@ from datetime import date
 from typing import Self
 
 import attr
+from aiohttp import ClientSession
 from mozilla_version.gecko import GeckoVersion
 from mozilla_version.version import BaseVersion
 from scriptworker.client import TaskVerificationError
@@ -51,7 +52,7 @@ class MergeInfo:
         return cls(**kwargs)
 
 
-async def run(github_client: GithubClient, public_artifact_dir: str, merge_info: MergeInfo) -> list[LandoAction]:
+async def run(session: ClientSession, github_client: GithubClient, public_artifact_dir: str, merge_info: MergeInfo) -> list[LandoAction]:
     to_branch = merge_info.to_branch
     from_branch = merge_info.from_branch
     end_tag = merge_info.end_tag
@@ -64,11 +65,12 @@ async def run(github_client: GithubClient, public_artifact_dir: str, merge_info:
     to_version = await get_version(github_client, version_file, to_branch)
     log.info(f"to_version is: {to_version}")
     if end_tag:
+        end_revision = await github_client.get_branch_head_oid(to_branch)
         # End tag specifically uses the `to_version` _before_ we bump it
         # (because we're declaring its current version as "done")
         end_tag_fmted = end_tag.format(major_version=to_version.major_number)
         log.info(f"Adding end_tag: {end_tag_fmted}")
-        actions.extend(tag.run([end_tag_fmted]))
+        actions.extend(await tag.run(session, tag.GitTagInfo(revision=end_revision, tags=[end_tag_fmted])))
 
     # We need to determine `bump_version`, which is what we will use when
     # performing version bumps later on. This version must be whatever version
@@ -80,6 +82,7 @@ async def run(github_client: GithubClient, public_artifact_dir: str, merge_info:
         bump_version = await get_version(github_client, version_file, from_branch)
         bump_branch = from_branch
         log.info(f"from_branch is present, got bump_version from it: {bump_version}")
+        end_revision = await github_client.get_branch_head_oid(from_branch)
 
         # base tagging _only_ happens when we have a `from_branch` -- these are
         # scenarios where we're uplifting one branch to another, and beginning a new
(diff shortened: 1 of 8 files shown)


bhearsum:
WIP for now, until the hg -> git look up is finalized.

landoscript/src/landoscript/actions/tag.py
```diff
@@ -1,17 +1,58 @@
...
+        git_commit = tag_info.revision
+    else:
+        # tag_info.revision is an hg revision; lando wants a git revision
+        resp = await retry_async(
+            session.get,
+            args=(f"{tag_info.hg_repo_url}/json-rev/{tag_info.revision}",),
+            kwargs={"raise_for_status": True},
+        )
+        hg_revision_info = await resp.json()
+        # TODO: fail if `git_commit` is None. We can only start doing this
```
bhearsum, line 45: I'll call out this TODO: explicitly. I expect to address it on Wednesday, after Try has been cut over; possibly before if I don't think I'll be testing this action again before cut-over.

landoscript/src/landoscript/actions/merge_day.py
```diff
@@ -85,9 +87,10 @@ async def run(github_client: GithubClient, public_artifact_dir: str, merge_info:
         # scenarios where we're uplifting one branch to another, and beginning a new
         # version number on the `to_branch`, which we declare with the `BASE` tag.
         if base_tag:
+            base_revision = await github_client.get_branch_head_oid(from_branch)
             base_tag_fmted = base_tag.format(major_version=bump_version.major_number)
             log.info(f"Adding base_tag: {base_tag_fmted}")
-            actions.extend(tag.run([base_tag_fmted]))
+            actions.extend(await tag.run(session, tag.GitTagInfo(revision=base_revision, tags=[base_tag_fmted])))
```
jcristau: Isn't there a race here if something lands between looking up the from_branch head and merging from_branch to to_branch?  Could we merge that base_revision to to_branch instead?

bhearsum: Hm...yes, possibly. I'll need to check if shas are supported for the merge action (but I don't see why they wouldn't be...).

bhearsum: (It works!)

landoscript/src/landoscript/actions/merge_day.py
```diff
@@ -80,18 +82,20 @@ async def run(github_client: GithubClient, public_artifact_dir: str, merge_info:
...
         if base_tag:
             base_tag_fmted = base_tag.format(major_version=bump_version.major_number)
             log.info(f"Adding base_tag: {base_tag_fmted}")
-            actions.extend(tag.run([base_tag_fmted]))
+            actions.extend(await tag.run(session, tag.GitTagInfo(revision=bump_ref, tags=[base_tag_fmted])))
     else:
         if merge_old_head:
             raise TaskVerificationError("'from_branch' is required when merge_old_head is True or version_files are present")
 
+        bump_ref = from_branch
```
jcristau: nit: dead store (we've just asserted that !merge_old_head so bump_ref won't get used)

landoscript/src/landoscript/actions/tag.py
jcristau: What happens right now if git_commit is None?  Looks like we add a tag action with no target?

landoscript/src/landoscript/script.py
```diff
@@ -98,10 +98,12 @@ async def async_main(context):
                     if version_bump_action:
                         lando_actions.append(version_bump_action)
                 elif action == "tag":
-                    tag_actions = tag.run(payload["tags"])
+                    tag_actions = await tag.run(session, tag.HgTagInfo(**payload["tag_info"]))
```
jcristau, line 103: nit: explicitly check that tag_info has a hg_repo_url?

jcristau (approved):

landoscript/src/landoscript/actions/tag.py
bhearsum: Yeah, which will tag the tip of the branch.

I suppose I can address this now, it will just need reverting to test on Try in the near term...

landoscript/src/landoscript/actions/merge_day.py
bhearsum: Hm, yes. And actually, I can simplify this and move the `if merge_old_head` code into the `if` branch above as it turns out...and get rid of the indirection.

landoscript/src/landoscript/script.py
bhearsum: Sure; we'll error out anyways due to the dataclass constructor, but it's probably better to have a TaskVerificationError than a TypeError.

PR #1188, opened 2025-04-26
fix(landoscript): add poll_time to config

bhearsum:


jcristau (commented):
Should this be in the default config instead?

jcristau (approved):

bhearsum:
> Should this be in the default config instead?

Yeah...that seems like a good idea. (Although I guess in practice, both locations are "default config" seeing as there's no other config source...)

PR #1190, opened 2025-04-28
fix(landoscript) include correct metadata for additions and removals in diffs

bhearsum:
These lines are necessary for the diff to apply correctly with `git apply`.

### DIFF
--- a/landoscript/src/landoscript/util/diffs.py
+++ b/landoscript/src/landoscript/util/diffs.py
@@ -3,6 +3,7 @@ from difflib import unified_diff
 
 def diff_contents(orig, modified, file):
     """Create a git-style unified diff of `orig` and `modified` with the filename `file`."""
+    add_remove_metadata = ""
     if orig:
         # orig exists already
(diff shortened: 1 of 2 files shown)


landoscript/src/landoscript/util/diffs.py
```diff
@@ -19,10 +21,11 @@ def diff_contents(orig, modified, file):
         # modified does not exist yet; ie: it will be added
         modified_contents = ""
         tofile = "/dev/null"
+        add_remove_metadata = "deleted file mode 100644\n"
```
jcristau, line 24: I guess in theory files might have a different mode, but in practice this is all just l10n info so unlikely to be executable?

bhearsum: Yeah. I suppose this could be more robust by looking at the file modes at the same time as their contents...probably as a future enhancement though? (I did test, and `git apply` will bail if the mode does not match, so we'll notice if this causes problems...)

jcristau (approved):

PR #1191, opened 2025-04-28
fix(landoscript): remove unnecessary spam in logs

bhearsum:
GQL is notably noisy, and the android l10n diffs are typically very large as well. All diffs are attached as artifacts anyways, so there's little point in logging them.

### DIFF
--- a/landoscript/src/landoscript/lando.py
+++ b/landoscript/src/landoscript/lando.py
@@ -1,7 +1,6 @@
 import asyncio
 import datetime
 import logging
-from pprint import pprint
 from typing import Any, Callable, Tuple
 
 from aiohttp import ClientResponseError, ClientSession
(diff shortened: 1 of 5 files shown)


landoscript/src/landoscript/lando.py
```diff
@@ -43,7 +43,13 @@ async def submit(
     url = f"{lando_api}/repo/{lando_repo}"
     json = {"actions": actions}
 
-    log.info(f"submitting actions to lando: {actions}")
+    log.info("submitting actions to lando:")
+    for action in actions:
+        print_action = action.copy()
+        if "diff" in print_action:
+            print_action["diff"] = "<omitted for brevity>"
+        log.info(action)
```
jcristau:
```suggestion
        log.info(print_action)
```

bhearsum: d'oh...thank you

jcristau (approved):

PR #1194, opened 2025-04-30
fix(landoscript): don't fail job when landing is still IN_PROGRESS

bhearsum:
Another fix based on a misunderstanding of the lando api.


hneiva (approved):
LGTM

PR #1195, opened 2025-04-30
Bug 1963695 - Use uv for pinning dependencies + repin

hneiva:
- Removes use of docker + pip-compile-multi
- Auto-installs required python version with uv command line


maintenance/pin.sh
```diff
@@ -55,33 +54,29 @@ PY_311_SCRIPTS=(
...
+            PY311_DIRS+=("${DIRS[$idx]}")
+        fi
+    done
 done
 
 if [ ${#PY38_DIRS} -gt 0 ]; then
-    printf "FROM python:3.8\n${RUNCMD}" | docker build --platform linux/x86_64 --pull --tag "scriptworker-script-pin:3.8" -
-    echo "${PY38_DIRS[@]}" | xargs -n4 -P4 docker run --platform linux/x86_64 --rm -t -v "$PWD":/src -e EXTRA_ARGS="$EXTRA_ARGS" -e SUFFIX=py38.txt -w /src scriptworker-script-pin:3.8 maintenance/pin-helper.sh
+    # Make sure python 3.8 is installed
+    uv python install 3.8
```
bhearsum: IMO we should still be doing this in `docker`, or finding some way to do this outside of it that doesn't affect local state. (It is surprising behaviour that running this script would affect what python versions I have available...)

bhearsum (changes requested):
> Removes use of docker + pip-compile-multi

Does `uv` behave the same regardless of the host platform?

maintenance/pin.sh
hneiva: Even though `uv` uses standalone python distributions ([ref](https://docs.astral.sh/uv/guides/install-python/#getting-started)), I agree this script shouldn't be doing that for you.

I pondered for a bit and decided to just error out in case it can't find the required python versions. I leaned towards this instead of keep using docker.

maintenance/pin.sh
```diff
@@ -55,33 +54,40 @@ PY_311_SCRIPTS=(
...
+            PY311_DIRS+=("${DIRS[$idx]}")
+        fi
+    done
 done
 
+INSTALLED_PYTHON_VERSIONS=$(uv python list)
+check_python_version() {
+    local version=$1
+    if ! echo $INSTALLED_PYTHON_VERSIONS | grep -q "$version"; then
+        echo "ERROR: Python $version is not installed. Hint: run 'uv python install $version'. If you are using a virtualenv, make sure to activate it first."
```
bhearsum, line 77: This works too :)

bhearsum (approved):

maintenance/pin.sh
jcristau: Can you please reconsider the docker part?

maintenance/pin-helper.sh
```diff
@@ -2,28 +2,55 @@
...
+    for req_file in "${REQ_FILES[@]}"; do
+        echo "$dir" "$req_file" "$SUFFIX"
+        if [ ! -f "requirements/$req_file.in" ]; then
+            echo "File $dir/requirements/$req_file.in does not exist"
+            exit 1
+        fi
+        uv pip compile "requirements/$req_file.in" \
+            --output-file "requirements/$req_file.$SUFFIX" \
+            --python-version "$PYTHON_VERSION" \
+            --universal \
```
Eijebong, line 51: Don't we want `-U` here to actually update stuff?

maintenance/pin.sh
hneiva: @jcristau could you please elaborate?
`uv` handles different OS's with `--universal`, which in my mind was the reason why we used docker to begin with. Maybe I'm missing other benefits?

jcristau: Mostly that I have docker, I don't have uv, I'd rather not have yet another tool to remember to maintain/update, so this doesn't work for me right now.

PR #1196, opened 2025-04-30
fix(landoscript): improve commit messages

bhearsum:
* adjust version bump and merge day commit messages to not include `Subject:`
* ensure `no bug` is in all commit messages, to avoid tripping over hg hooks
    
I'm not entirely sure how this prefix got into them, but it's not correct...

### DIFF
--- a/landoscript/src/landoscript/actions/android_l10n_import.py
+++ b/landoscript/src/landoscript/actions/android_l10n_import.py
@@ -109,5 +109,5 @@ async def run(
         log.info("adding android l10n import diff! contents omitted from log for brevity")
 
         # We always ignore closed trees for android l10n imports.
-        commitmsg = f"Import translations from {l10n_repo_url} CLOSED TREE"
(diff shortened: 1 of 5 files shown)


landoscript/src/landoscript/actions/android_l10n_import.py
```diff
@@ -109,5 +109,5 @@ async def run(
         log.info("adding android l10n import diff! contents omitted from log for brevity")
 
         # We always ignore closed trees for android l10n imports.
-        commitmsg = f"Import translations from {l10n_repo_url} CLOSED TREE"
+        commitmsg = f"no bug - Import translations from {l10n_repo_url} CLOSED TREE"
```
ahal: nit: I prefer `No Bug`

ahal (approved):

landoscript/src/landoscript/actions/android_l10n_import.py
bhearsum: done!

PR #1202, opened 2025-05-05
refactor(pin.sh): use docker instead of local uv

hneiva:


ahal:
I don't really understand what this solves, and seems like it'll make it more painful if we ever want to add tooling in the future (e.g, dependabot / renovate)

bhearsum:
I was about to go ahead and do this myself, primarily because pinning packages now depends on the state of the host system. In my case, it first meant installing some missing system packages that the pushflat dependencies require. After that, I ran into some weird inconsistencies when generating python 3.8 dependencies (trailing markers were getting removed and changed, which I presume was because I was running with a different version of uv than the last set was generated with?)

I think the point of dockerizing is to make sure that we have consistency regardless of where the pinning is done or who does it.

bhearsum:
My update to this PR seems to work; but maybe we need to have a discussion about this before proceeding?

addonscript/tests/test_api.py
```diff
@@ -18,7 +18,7 @@ def does_not_raise():
 
 
 @pytest_asyncio.fixture(scope="function")
-async def fake_session(event_loop):
```
Eijebong: You can probably unpin pytest-asyncio if you keep that in?

bhearsum: You're right...but on second thought it's probably best to avoid conflating; I'm going to drop this part.

Eijebong (approved):
Found the difference between the docker uv setup and the non docker one 🙃

So the one difference we're seeing in the lockfile is https://github.com/nedbat/coveragepy/blob/52407da7b9ed99c8bd1d2536df82ba8749664f50/setup.py#L98

I checked a bit and it turns out that the docker version passes `PYTHON_VERSION=3.11.9` to `uv pip compile --python-version` and the non docker one passed `3.11`. `3.11` matches that random alpha version that is referenced by coveragepy so tomli gets added as a dependency.

Technically this PR's output is more correct but it has nothing to do with docker, it's just because the version of python passed to `uv pip compile` is different.

With the following patch I get the exact same output.

```diff
diff --git a/maintenance/pin.sh b/maintenance/pin.sh
index b3bbee2c..cad0c41d 100755
--- a/maintenance/pin.sh
+++ b/maintenance/pin.sh
@@ -83,11 +83,11 @@ if [ ${#PY38_DIRS} -gt 0 ]; then
     # Make sure python 3.8 is installed
     # This is the version used in the old mac signers, once the new signers are in place, we can remove this
     # and just use python 3.11 for everything
-    check_python_version 3.8
-    PYTHON_VERSION=3.8 SUFFIX=py38.txt maintenance/pin-helper.sh "${PY38_DIRS[@]}"
+    check_python_version 3.8.3
+    PYTHON_VERSION=3.8.3 SUFFIX=py38.txt maintenance/pin-helper.sh "${PY38_DIRS[@]}"
 fi
 if [ ${#PY311_DIRS} -gt 0 ]; then
     # Make sure python 3.11 is installed
-    check_python_version 3.11
-    PYTHON_VERSION=3.11 SUFFIX=txt maintenance/pin-helper.sh "${PY311_DIRS[@]}"
+    check_python_version 3.11.9
+    PYTHON_VERSION=3.11.9 SUFFIX=txt maintenance/pin-helper.sh "${PY311_DIRS[@]}"
 fi
```


I'm fine with this PR as it doesn't seem to be a maintenance burden nor does it make pinning significantly slower (it's still magnitudes of order faster than the old pip-compile-multi stuff) and it means that people don't have to install uv and don't have to manage their python versions just to pin dependencies.

bhearsum:
@Eijebong dug into this more:

> Alright, finally figured this out... I could not for the life of me get it to call out to a compiler, tried on linux, macos, windows...
> Ended up compiling an old version of uv and wouldn't you know, a build happened \o/
> So the problematic dependency is pygobject, you can add pygobject<3.52 in a file and call uv pip compile req.in and reproduce. But you can only reproduce with uv 0.5.6 and lower. It got fixed by https://github.com/astral-sh/uv/commit/c314c68bff660cb3cd699e946e7c4ae387678a40 which hilariously lists the exact same dependency as the one we were having issues with.
> Solution: Update uv 

I guess it's theoretically possible we might end up with some similar situation in the future, but I'm fine with leaving this undockerized until we know it's going to be a real issue.

PR #1203, opened 2025-05-06
fix(landoscript): ensure diffs handle end of file newline (or not) correctly

bhearsum:
While testing some merge automation, which modifies the `CLOBBER` file (which does not have an end of file newline), I discovered that diffs were not applying correctly. Ultimately, I realized that the end of file handling in `diff_contents` was somewhat broken in a few scenarios. These were not caught in the existing tests because I didn't understand what the diffs should look like.

This update should resolve all of that, and adds some fairly extensive unit testing specifically for that function. Because it is as well tested as it on its own now, I relaxed some of the checks in the higher level tests, because handling the newline and non-newline cases there would be a huge pain in the butt.


bhearsum:
This is working well. For example, in https://firefox-ci-tc.services.mozilla.com/tasks/G0C_-MBxTpOxbYJDjPn65w

jcristau (approved):

PR #1207, opened 2025-05-06
fix(landoscript): do l10n bumps correctly for merge day

bhearsum:
I had hoped we could run these as their own separate action, but because they need to happen _after_ the merge operation, only the `merge_day` action has enough information to accurately generate a diff for them. (This is different than in treescript, where we're operating on files on disk, and we don't need to look-up file contents of remote branches.)

The test parts of this diff are very big, but they are essentially just:
- Updating the existing test in `test_script.py` to do l10n_bump this way -- this is mostly shuffling around the payload and when certain data is returned by the github API.
- Replacing the current test in `test_merge_day.py` with that one (now that it is not a cross-action test).


bhearsum:
Draft for now while I finish testing.

bhearsum:
> Draft for now while I finish testing.

It worked in https://firefox-ci-tc.services.mozilla.com/tasks/bIJ0rXqgSyKVDuPFbLL2Pg!

ahal (approved):
Thanks, looks reasonable to me!

PR #1209, opened 2025-05-13
bug 1965876: fix(landoscript): include l10n.toml changes in android l10n updates

bhearsum:
This is a pure oversight from my original work here: I didn't realize these files were automatically updated in addition to strings. Typically this happens when a locale is added or removed.

The fix is fairly straightforward: additionally fetch the l10n.toml files from the destination repository and diff them against the ones from the source.

I was able to test this pretty thoroughly on try: [import task](https://firefox-ci-tc.services.mozilla.com/tasks/cPhNifENRsyt6CZuMUKh8w) and [commit](https://github.com/mozilla-releng/staging-firefox/commit/d85abd381aa4475101cbbb7f13a85c3c0a595b49#diff-8174e188ffab98600baf1c46a58b3a3bf3a98eb2e216fb0b844a1798f1b2bee0), [sync task](https://firefox-ci-tc.services.mozilla.com/tasks/NbXlOcjiQAysMxU-HGRofQ) and [commit](https://github.com/mozilla-releng/staging-firefox/commit/fc4ebbc1df0a5117a80a74876b408486786cb7b3). In both cases they correctly updated the new `bqi` locale to android-components and fenix:
```
diff --git a/mobile/android/android-components/l10n.toml b/mobile/android/android-components/l10n.toml
index 99e4fea201d5..6eb5ab3f3e2c 100644
--- a/mobile/android/android-components/l10n.toml
+++ b/mobile/android/android-components/l10n.toml
@@ -19,6 +19,7 @@ locales = [
     "bg",
     "bn",
     "bo",
+    "bqi",
     "br",
     "bs",
     "ca",
diff --git a/mobile/android/fenix/l10n.toml b/mobile/android/fenix/l10n.toml
index d150ae38f79c..cdfbf2900afc 100644
--- a/mobile/android/fenix/l10n.toml
+++ b/mobile/android/fenix/l10n.toml
@@ -19,6 +19,7 @@ locales = [
   "bg",
   "bn",
   "bo",
+  "bqi",
   "br",
   "bs",
   "ca",
```

### DIFF
--- a/landoscript/src/landoscript/actions/android_l10n_import.py
+++ b/landoscript/src/landoscript/actions/android_l10n_import.py
@@ -76,20 +76,28 @@ async def run(
         src_files = [f.src_name for f in l10n_files]
         log.info(f"fetching updated files from l10n repository: {src_files}")
         new_files = await l10n_github_client.get_files(src_files)
+        # we also need to update the toml files with locale metadata. we've
+        # already fetched them above; so just add their contents by hand
+        new_files.update(toml_contents)
 
         # fetch l10n_files from gecko repo
+        # `l10n_files` will give us all of the files with translations in them
         dst_files = [f.dst_name for f in l10n_files]
+        # we also need the gecko locations of the toml files
+        for toml_info in android_l10n_import_info.toml_info:
+            dst_files.append(f"{toml_info.dest_path}/l10n.toml")
+
         log.info(f"fetching original files from l10n repository: {dst_files}")
         orig_files = await github_client.get_files(dst_files, branch=to_branch)
 
         diff = ""
         for l10n_file in l10n_files:
             if l10n_file.dst_name not in orig_files:
-                log.warning(f"WEIRD: {l10n_file.dst_name} not in dst_files, continuing anyways...")
+                log.warning(f"WEIRD: {l10n_file.dst_name} not in orig_files, continuing anyways...")
                 continue
 
             if l10n_file.src_name not in new_files:
-                log.warning(f"WEIRD: {l10n_file.src_name} not in src_files, continuing anyways...")
+                log.warning(f"WEIRD: {l10n_file.src_name} not in new_files, continuing anyways...")
                 continue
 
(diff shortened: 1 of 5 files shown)


landoscript/src/landoscript/actions/android_l10n_import.py
```diff
@@ -100,6 +108,25 @@ async def run(
 
             diff += diff_contents(orig_file, new_file, l10n_file.dst_name)
 
+        for info in android_l10n_import_info.toml_info:
```
jcristau: nit: `for info in android_l10n_import_info.toml_info` here vs `for toml_info in android_l10n_import_info.toml_info` above; might be better to stick to the same variable name?

landoscript/src/landoscript/actions/android_l10n_import.py
```diff
@@ -100,6 +108,25 @@ async def run(
...
+            dst_path = f"{info.dest_path}/l10n.toml"
+
+            if src_path not in new_files:
+                log.warning(f"WEIRD: {src_path} not in new_files...continuing anyways...")
+
+            if dst_path not in orig_files:
+                log.warning(f"WEIRD: {dst_path} not in orig_files...continuing anyways...")
+
+            orig_file = orig_files[dst_path]
+            new_file = new_files[src_path]
```
jcristau, line 119: This will crash if we hit one of the warnings above?  Seems like it would be better to either remove the warning or raise right away.

bhearsum: Or option 3: `continue`. This would more gracefully handle a theoretically case where we have a newly added `l10n.toml`. That's probably not worth optimizing for though; I think raising a more specific error is probably the right thing to do.

landoscript/src/landoscript/actions/android_l10n_import.py
bhearsum: Yeah, that's fair.

jcristau (approved):

PR #1210, opened 2025-05-15
fix(landoscript): bug 1966092: fail with intermittent-task when a merge conflict is encountered

bhearsum:


bhearsum:
I'm still attempting to repro and prove this fix on try, but this can probably be reviewed in the meantime.

jcristau (approved):

bhearsum:
I was finally able to repro this in staging. I found an issue with my original patch where `exit_code` wasn't set correctly (now fixed).

With that fixed, I got the expected behaviour given the current state of lando:
* [One task succeeded](https://firefox-ci-tc.services.mozilla.com/tasks/QDTOo57xTXSONbwC34qfZA), and pushed a tag and the version bump
* The other [two](https://firefox-ci-tc.services.mozilla.com/tasks/KQVN6oF9QsCbLAU4HCmFoA) [tasks](https://firefox-ci-tc.services.mozilla.com/tasks/Tt1QITf5T4uiERvoqPCyew/runs/2) failed with intermittent-task on their first attempt, because they tried to bump the version, and that resulted in a merge conflict. On their second attempts they only tried to push the tag, and ran into `tag xxx already exists` errors, which is what would expect until https://bugzilla.mozilla.org/show_bug.cgi?id=1966118 is fixed.

Issue #931, opened 2024-03-01
shipitscript: pass repo_url when creating a release


jcristau:
See https://github.com/mozilla-releng/shipit/pull/1411

PR #1215, opened 2025-05-21
Add support for creating new releases for a github repo in shipitscript

Eijebong:


shipitscript/src/shipitscript/script.py
```diff
@@ -42,20 +42,28 @@ def create_new_release_action(context):
...
+
+        log.info("Determining most recent shippable revision")
+        shippable_revision = ship_actions.get_shippable_revision(branch, last_shipped_revision, cron_revision)
+        if not shippable_revision:
+            log.info("No valid shippable revision found, silent exit ...")
+            return
+        log.info(f"The shippable revision found is {shippable_revision}")
+    elif repository_type == "github":
+        shippable_revision = cron_revision
+        log.info(f"Repository type is github, shipping `cron_revision` {shippable_revision} without checking")
```
jcristau: Can you explain?

shipitscript/src/shipitscript/script.py
```diff
@@ -42,20 +42,28 @@ def create_new_release_action(context):
...
+    else:
+        log.error(f"Unknown repository type: {repository_type}. Exiting...")
```
jcristau: Shouldn't this be a malformed payload error?

jcristau (changes requested):

shipitscript/src/shipitscript/script.py
Eijebong: When the repository type is hg, shipit checks if there's any important commit since the last release by parsing the push log. The main issue is that the push log of a github project won't contain any of the markers (`a=...`, `DONTBUILD`, `L10n bumps`) so I decided against even thinking about it. I thought I had put a TODO on this though 🤔 Maybe I never saved the file.

shipitscript/src/shipitscript/script.py
Eijebong: Yup, done

shipitscript/src/shipitscript/script.py
jcristau: Don't we still need to avoid building twice from the same rev?

Eijebong: I guess we could get the latest release on that branch if there's one, get the SHA for that and compare?

Eijebong:
Have to add support for the repo_url because shipit needs it to get the proper trust domain to figure out the decision task index path.

https://github.com/mozilla-releng/shipit/blob/cb43f6f9b01d26188996456a08d08b5ba8a677af/api/src/shipit_api/admin/tasks.py#L40

Edit: We can guess repository_type from this instead of having both fields

shipitscript/src/shipitscript/script.py
Eijebong: Done now

Eijebong: https://firefox-ci-tc.services.mozilla.com/tasks/D3ghBHPPQfeYQnTnObt0TA/runs/0/logs/public/logs/live_backing.log tested here where the only commit on the release branch was the version bump that had just happened

shipitscript/src/shipitscript/script.py
```diff
@@ -59,14 +60,14 @@ def create_new_release_action(context):
             log.info("No valid shippable revision found, silent exit ...")
             return
         log.info(f"The shippable revision found is {shippable_revision}")
-    elif repository_type == "github":
+    elif "github.com" in repository_url:
```
jcristau: Might want to use mozilla-repo-urls (as a followup is also fine)?

shipitscript/src/shipitscript/shipitapi.py
```diff
@@ -110,10 +110,12 @@ def get_disabled_products(self, headers={}):
                 log.error(f"Response code: {resp.status_code}")
             raise
 
-    def create_new_release(self, product, branch, version, build_number, revision, headers={}):
+    def create_new_release(self, product, branch, version, build_number, revision, repository_url, headers={}):
         """Method to map over the POST /releases/ API in Ship-it"""
         resp = None
         params = {"product": product, "branch": branch, "version": version, "build_number": build_number, "revision": revision, "partial_updates": "auto"}
+        if repository_url is not None:
+            params["repo_url"] = repository_url
```
jcristau, line 118: This might fix #931, at least partly?

shipitscript/src/shipitscript/data/create_new_release_task_schema.json
```diff
@@ -36,7 +36,7 @@
                 "cron_revision": {
                   "type": "string"
                 },
-                "repository_type": {
+                "repository_url": {
```
jcristau: I have a bit of a concern here with using an arbitrary repo url coming from the payload. Can we either pull it from something that can be verified to be the same repo that the task is coming from, or use scopes somehow, to prevent creating releases across repos?

jcristau (changes requested):

shipitscript/src/shipitscript/data/create_new_release_task_schema.json
Eijebong: I now extract the repo url from the metadata->source of the task as that gets validated early.

shipitscript/src/shipitscript/shipitapi.py
Eijebong: Actually I can fix it completely I think. I'll do that

shipitscript/src/shipitscript/script.py
```diff
@@ -43,22 +45,47 @@ def create_new_release_action(context):
...
+        log.info("Determining most recent shippable revision")
+        repository_url = f"https://hg.mozilla.org/{source_url.repo}"
+
+        shippable_revision = ship_actions.get_shippable_revision(branch, last_shipped_revision, cron_revision, source_url)
+    elif source_url.platform == "github":
+        if last_shipped_revision:
+            log.info(f"Last shipped revision is {last_shipped_revision}")
+
+            log.info("Determining if cron revision is shippable")
+            repository_url = f"https://github.com/{source_url.owner}/{source_url.repo}"
```
Eijebong: This should be above line 68

ahal:
Out of curiosity, what's the benefit of doing this vs having a `githubscript` task in the release graph that does it?

Eijebong:
This is to create a shipit release from a GitHub repo, not the opposite. It's to get automated betas for firefox ios

shipitscript/src/shipitscript/script.py
```diff
@@ -43,22 +45,48 @@ def create_new_release_action(context):
...
-    log.info("Determining most recent shippable revision")
-    shippable_revision = ship_actions.get_shippable_revision(branch, last_shipped_revision, cron_revision)
+    if source_url is None or source_url.platform == "hgmo":
+        if not last_shipped_revision:
+            log.error("Something is broken under the sun if no shipped revision")
+            sys.exit(1)
+        log.info(f"Last shipped revision is {last_shipped_revision}")
+
+        log.info("Determining most recent shippable revision")
+        repository_url = f"https://hg.mozilla.org/{source_url.repo}"
```
jcristau: That won't work if source_url is None.  Though I think you can assert that it is not?

jcristau: Also isn't repository_url just source_url.url, both in the hgmo and github cases?

Eijebong: > That won't work if source_url is None. Though I think you can assert that it is not?

Oops, good catch, adding a test for this. I don't think we should assert that it's there, scriptworker doesn't unless CoT gets verified which isn't a guarantee AFAIK.

> Also isn't repository_url just source_url.url, both in the hgmo and github cases?

Nope

```diff
  -     'https://hg.mozilla.org/releases/mozilla-beta',
  +     'https://hg.mozilla.org/releases/mozilla-beta/file/8a2c52b9488e5ae5a8dc8500d3430d5ebe3d9a85/taskcluster/kinds/maybe-release'
```

shipitscript/setup.py
```diff
@@ -27,5 +27,5 @@
     entry_points={"console_scripts": ["shipitscript = shipitscript.script:main"]},
     license="MPL2",
     install_requires=requirements,
-    classifiers=("Programming Language :: Python :: 3.6", "Programming Language :: Python :: 3.7"),
+    classifiers=["Programming Language :: Python :: 3.6", "Programming Language :: Python :: 3.7"],
```
jcristau, line 30: drive-by: that's a bit out of date :)

shipitscript/src/shipitscript/shipitapi.py
```diff
@@ -110,13 +110,15 @@ def get_disabled_products(self, headers={}):
                 log.error(f"Response code: {resp.status_code}")
             raise
 
-    def create_new_release(self, product, branch, version, build_number, revision, headers={}):
+    def create_new_release(self, product, branch, version, build_number, revision, repository_url, headers={}):
         """Method to map over the POST /releases/ API in Ship-it"""
         resp = None
         params = {"product": product, "branch": branch, "version": version, "build_number": build_number, "revision": revision, "partial_updates": "auto"}
+        if repository_url is not None:
+            params["repo_url"] = repository_url.format()
```
jcristau: question: why .format()?

shipitscript/setup.py
Eijebong: Hahaha, I fixed that when I made sure that I didn't break setup.py when I removed the scriptworker_client dependency in the middle of this patch, because it turns out that it wasn't working in the first place. I can fix the classifiers if you want though

shipitscript/src/shipitscript/shipitapi.py
Eijebong: Leftover and python is like "nah I don't care that it's empty" T_T
I'll fix that

jcristau (approved):

PR #1217, opened 2025-05-23
bug 1967787: feat: handle landoscript terminations better

bhearsum:
This improves landoscript reruns by looking for the lando status URL from an earlier run, and polling it instead of redoing work (that is likely to fail). I don't think we can be _perfect_ here, but this is much better than the current state of affairs.

### DIFF
--- a/landoscript/requirements/base.txt
+++ b/landoscript/requirements/base.txt
@@ -310,7 +310,7 @@ click==8.2.0 \
     --hash=sha256:6b303f0b2aa85f1cb4e5303078fadcbcd4e476f114fab9b5007005711839325c \
     --hash=sha256:f5452aeddd9988eefa20f90f05ab66f17fce1ee2a36907fd30b05bbb5953814d
     # via cookiecutter
-colorama==0.4.6 ; sys_platform == 'win32' \
+colorama==0.4.6 ; platform_system == 'Windows' \
     --hash=sha256:08695f5cb7ed6e0531a20572697297273c47b8cae5a63ffc6d6ed5c201be6e44 \
     --hash=sha256:4f1d9991f5acc0ca119f9d443620b77f9d6b33703e51011c16baf57afb285fc6
     # via click
--- a/landoscript/src/landoscript/script.py
+++ b/landoscript/src/landoscript/script.py
@@ -5,11 +5,14 @@ import os.path
 import aiohttp
 import scriptworker.client
 from scriptworker.exceptions import TaskVerificationError
+from scriptworker.utils import calculate_sleep_time
 
+import taskcluster
 from landoscript import lando, treestatus
 from landoscript.actions import android_l10n_import, android_l10n_sync, l10n_bump, merge_day, tag, version_bump
 from scriptworker_client.github import extract_github_repo_owner_and_name
 from scriptworker_client.github_client import GithubClient
+from scriptworker_client.utils import retry_async
 
 log = logging.getLogger(__name__)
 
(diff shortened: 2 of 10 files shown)


bhearsum:
I still have some real world testing to do here, and need to pick up the yet-to-be-released scriptworker 60.10.0 before this will actually work in practice.

bhearsum:
I've tested this on Try. Here is a [success case](https://firefox-ci-tc.services.mozilla.com/tasks/GeNLwOQlSBy2W4bB928juQ/runs/2) (run #1 of that contained a small bug in the rerun handling, which was fixed before run #2 ran). Here is a [failure case](https://firefox-ci-tc.services.mozilla.com/tasks/Ld7eRfpQTyeBxS3bG1p_kA/runs/1).

landoscript/src/landoscript/script.py
```diff
@@ -35,8 +35,89 @@ def validate_scopes(scopes: set, lando_repo: str, actions: list[str]):
         raise scriptworker.client.TaskVerificationError(f"required scope(s) not present: {', '.join(missing)}")
 
 
-def sanity_check_payload(payload):
-    pass
+def sanity_check_payload(payload, scopes, lando_repo):
+    """Additional verification past what the task schema does."""
+    # validate scopes - these raise if there's any scope issues
+    validate_scopes(scopes, lando_repo, payload["actions"])
+    if len(payload["actions"]) < 1:
```
jcristau, line 45: nit: the schema already says `"minItems": 1`
(I realize you're just moving this code, so fixing this should be done separately anyway)

landoscript/requirements/base.txt
```diff
@@ -4,88 +4,93 @@ aiohappyeyeballs==2.6.1 \
     --hash=sha256:c3f9d0113123803ccadfdf3f0faa505bc78e6a72d1cc4806cbd719826e943558 \
     --hash=sha256:f349ba8f4b75cb25c99c5c2d84e997e485204d2902a9597802b0371f09331fb8
     # via aiohttp
-aiohttp==3.11.18 \
```
jcristau: Unfortunate that we end up updating unrelated things in this commit.

landoscript/src/landoscript/script.py
```diff
@@ -120,6 +123,33 @@ async def process_actions(session, context, owner, repo, public_artifact_dir, br
...
+    while run_id > 0:
+        run_id -= 1
+        queue = taskcluster.Queue(options={"rootUrl": root_url})
+        try:
+            artifactInfo = queue.getArtifact(task_id, run_id, "public/build/lando-status.txt")
+            assert artifactInfo
+            url = artifactInfo.get("url")
+            assert url
+            resp = await session.get(str(url))
+            return (await resp.content.read()).decode()
```
jcristau: We probably need to check resp.status before using the contents?
And consider adding retries.

landoscript/src/landoscript/script.py
bhearsum: Yeah, good call.

landoscript/requirements/base.txt
bhearsum: I can probably fix this...

bhearsum: Oh no, we don't have any easy way to only add dependencies....meh.

Eijebong: You can `uv pip compile -P packagename requirements/base.in --output-file requirements/base.txt --python-version 3.11 --universal --generate-hashes` for each of the .in file, it should only touch things related to `packagename`

bhearsum: ah...this is probably easier to achieve now that we're using uv, indeed! (i don't think it was so easy with pip...)

landoscript/src/landoscript/script.py
```diff
@@ -35,8 +38,124 @@ def validate_scopes(scopes: set, lando_repo: str, actions: list[str]):
...
+                kwargs={"raise_for_status": True},
+                attempts=5,
+                retry_exceptions=aiohttp.ClientResponseError,
+                sleeptime_callback=calculate_sleep_time,
+            )
+            return (await resp.content.read()).decode()
+        except taskcluster.TaskclusterRestFailure as e:
+            if e.status_code >= 500:
+                log.error("taskcluster is unavailable...cannot continue!")
+                raise e
```
jcristau:
```suggestion
                raise
```

jcristau (approved):

PR #1220, opened 2025-05-28
Bug 1946095 - Add support for the samsung galaxy store to pushapkscript

Eijebong:
The target store was already part of the task schema as a leftover from amazon. I just added support for samsung there. Because it also has support for rollouts, I made that non google specific. I also overloaded the `commit` parameter although a dry run on the samsung galaxy store won't do anything beside checking that the arguments passed in and the APKs are correct.

### DIFF
--- a/pushapkscript/src/pushapkscript/publish.py
+++ b/pushapkscript/src/pushapkscript/publish.py
@@ -6,15 +6,16 @@ from mozapkpublisher.push_apk import push_apk
 log = logging.getLogger(__name__)
 
 
-def publish(product_config, publish_config, apk_files, contact_server):
+async def publish(product_config, publish_config, apk_files, contact_server):
     if apk_files:
-        push_apk(
+        await push_apk(
             apks=apk_files,
-            secret=publish_config["secret"],
+            secret=publish_config.get("secret"),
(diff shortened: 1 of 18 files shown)


pushapkscript/src/pushapkscript/publish.py
```diff
@@ -13,7 +13,7 @@ def publish(product_config, publish_config, apk_files, contact_server):
             secret=publish_config["secret"],
             expected_package_names=publish_config["package_names"],
             track=publish_config.get("google_track"),
-            rollout_percentage=publish_config.get("google_rollout_percentage"),
```
bhearsum, line 16: Thanks for making the effort to avoid this becoming a misnomer.

bhearsum (commented):
This looks quite plausible to me; it would be good to see a try run to make sure that everything that can be run there is tested? (I think you also mentioned that you need to pick up a `mozapkpublisher` bump?)

Eijebong:
Try push https://firefox-ci-tc.services.mozilla.com/tasks/groups/VZvDS4hXQ0WEAMxpEVTLeA

bhearsum (approved):

PR #1222, opened 2025-05-29
feat(beetmover): add support for uploading payload data as a file

bhearsum:
The motivation here is to make uploading the experiment configuration for translations easier (see https://github.com/mozilla/translations/issues/466). This config comes in as action task input, and is never published as a task artifact in the form that we want it. Rather than create an unnecessary file (most likely through a special task that simply creates that file out of parameters), let's add support for writing out files that come into beetmover as data.

I'm not _super_ happy about requiring base64 encoded data, but I'm also not sure of a better alternative. (At one point I allowed JSON objects to given instead of a base64 encoded string - but that only serves a much narrower set of use cases.)


bhearsum:
An example usage of this is over in https://firefox-ci-tc.services.mozilla.com/tasks/RH8TCETfQz29bLK0_WhprQ

hneiva (approved):
As a feature and code LGTM
I'm a bit worried about ppl using this to upload things right off of decision tasks, slowing down taskgraph even more

bhearsum:
> As a feature and code LGTM I'm a bit worried about ppl using this to upload things right off of decision tasks, slowing down taskgraph even more

Well...that's exactly the intended use case here :laughing:. It's hard to imagine anything large being uploaded through these means...but we can cross that bridge if we come to it.

bhearsum:
> As a feature and code LGTM I'm a bit worried about ppl using this to upload things right off of decision tasks, slowing down taskgraph even more

I was thinking about this a bit more, and I think it's less of a concern than I imagined. Uploading things directly via the decision task by this mechanism should be no worse than if the decision task were to create another task to do so, as long as the data in question is already available to the decision task. In either case, we'd be transmitting the data to a downstream task definition.

This does, of course, make it a bit easier for someone to do extra work (eg: fetch some remote data) that is not already done as part of the decision task and have that feed directly into a downstream beetmover task. This scenario seems fairly unlikely to me though, and probably more difficult than putting that in some other task and having that feed into beetmover.

PR #1224, opened 2025-05-30
fix: use gcp prod for dev/fake-prod vpn signing

bhearsum:
...and add a note about why we don't use a VPN-looking keyid.

Additional context on that that doesn't belong in the comments is that prior to #1099, we didn't have _any_ entry for apk signing in dev and fake-prod. In a more ideal world we might consider using a single, generically named, key for all dep apk signing. I'm not sure it's worth pursuing that at the moment.


jcristau (approved):
Thank you.

PR #1227, opened 2025-06-03
Replace jose with pyjwt

### LINKED ISSUE #1104: addonscript: move away from python-jose
See https://github.com/mpdavis/python-jose/issues/340 for example.

Eijebong:
jose is very sporadically maintained and we already have a dependency on pyjwt through github3-py.

Fixes #1104

### DIFF
--- a/addonscript/requirements/base.in
+++ b/addonscript/requirements/base.in
@@ -1,6 +1,6 @@
 # pyup: ignore file
 aiohttp
 async-timeout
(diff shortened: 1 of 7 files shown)


bhearsum (approved):
I see this has coverage, but it might be worth a run through try nonetheless. Your call.

jcristau (approved):
Thanks!

PR #1229, opened 2025-06-05
Make sure the versions of python we use are matching eveywhere

Eijebong:
The tests were running python "3.8" which was pulling 3.8.20 but we're keeping this around for old mac signers which are running 3.8.3. I switched the image to use that instead

With that out of the way, both docker images are using a fully qualified python version so I switched the pin script to use those instead of letting uv pick the lowest possible version that matched. This shouldn't change much but is more correct

I added a full repin to get the noise from switching out of the way. It has the benefits of unpinning aiohttp (see #1228 for why it was pinned).


jcristau (approved):
:+1:

PR #1231, opened 2025-06-13
landoscript diffs should have sorted filenames

SoundBot:
This addresses this bugzilla ticket: https://bugzilla.mozilla.org/show_bug.cgi?id=1967637:

> A typical diff is sorted by filename/path. The diffs that [landoscript](https://github.com/mozilla-releng/scriptworker-scripts/tree/master/landoscript) generates are not. This can plausibly happen in any action that calls diff_contents more than once, but I've certainly seen it in the android_l10n and merge_day actions.
> 
> It's an extremely minor thing, but it can be confusing to someone that's not aware that it may happen.

### DIFF
--- a/landoscript/src/landoscript/actions/merge_day.py
+++ b/landoscript/src/landoscript/actions/merge_day.py
@@ -150,7 +150,9 @@ async def run(
     # process replacements, regex-replacements, and update clobber file
     replacements = merge_info.replacements
     regex_replacements = merge_info.regex_replacements
-    diffs = []
+
+    files_to_diff = []
+
     if replacements or regex_replacements:
         log.info("Performing replacements and regex_replacements")
         needed_files = []
@@ -166,8 +168,7 @@ async def run(
         for fn in orig_contents:
             if orig_contents[fn] is None:
                 raise LandoscriptError(f"Couldn't find file '{fn}' in repository!")
-
-            diffs.append(diff_contents(str(orig_contents[fn]), new_contents[fn], fn))
+            files_to_diff.append((fn, str(orig_contents[fn]), new_contents[fn]))
 
     log.info("Touching clobber file")
     orig_clobber_file = (await github_client.get_files("CLOBBER", bump_branch))["CLOBBER"]
@@ -175,7 +176,14 @@ async def run(
         raise LandoscriptError("Couldn't find CLOBBER file in repository!")
 
     new_clobber_file = get_new_clobber_file(orig_clobber_file)
-    diffs.append(diff_contents(orig_clobber_file, new_clobber_file, "CLOBBER"))
+    files_to_diff.append(("CLOBBER", orig_clobber_file, new_clobber_file))
+
+    files_to_diff.sort(key=lambda x: x[0])
+
+    # Generate diffs in sorted order
+    diffs = []
+    for file_path, orig_file, new_file in files_to_diff:
+        diffs.append(diff_contents(orig_file, new_file, file_path))
 
     diff = "\n".join(diffs)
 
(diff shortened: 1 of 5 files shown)


landoscript/src/landoscript/actions/merge_day.py
```diff
@@ -166,16 +168,25 @@ async def run(
...
     orig_clobber_file = (await github_client.get_files("CLOBBER", bump_branch))["CLOBBER"]
     if orig_clobber_file is None:
         raise LandoscriptError("Couldn't find CLOBBER file in repository!")
 
     new_clobber_file = get_new_clobber_file(orig_clobber_file)
-    diff += diff_contents(orig_clobber_file, new_clobber_file, "CLOBBER")
+    files_to_diff.append(("CLOBBER", orig_clobber_file, new_clobber_file))
+
+    if not files_to_diff:
+        return []
```
bhearsum: This is impossible to hit in reality at the moment (because the `CLOBBER` file is always in this list), but if we need to return early here, we must return `actions`, because there may be others added above. (This is different than in `android_l10n_sync` of course, where there's only one potential action.)

It would probably be cleaner to write this as a `if files_to_diff` block anyways, and leave the only `return` statement at the end.

bhearsum (changes requested):
Thank you for putting this together! It looks largely fine modulo the few comments below.

Please also update https://github.com/mozilla-releng/scriptworker-scripts/blob/master/landoscript/src/landoscript/actions/android_l10n_import.py and https://github.com/mozilla-releng/scriptworker-scripts/blob/master/landoscript/src/landoscript/actions/version_bump.py to handle this correctly - thank you!

Also, while reviewing this I discovered an issue with the tests. That's being fixed in https://github.com/mozilla-releng/scriptworker-scripts/pull/1232. Once it lands you'll need to rebase your work on top of it to make sure it is tested properly.

landoscript/src/landoscript/actions/merge_day.py
SoundBot: makes sense, I removed that.

bhearsum:
> @bhearsum I made these changes, and I'm waiting on #1232

Thanks! What I see looks good. I'll hold off on formally approving until that PR merges, and you can rebase on top of it.

bhearsum:
> > @bhearsum I made these changes, and I'm waiting on #1232
> 
> Thanks! What I see looks good. I'll hold off on formally approving until that PR merges, and you can rebase on top of it.

This has merged now; you should be clear to refresh this work.

landoscript/tests/conftest.py
```diff
@@ -217,6 +217,19 @@ def assert_add_commit_response(action, commit_msg_strings, initial_values, expec
...
     diffs = action["diff"].split("\ndiff")
 
+    # Extract file paths from diffs and verify they are sorted
+    file_paths = []
+    for diff in diffs:
+        if not diff:
+            continue
+
+        path_line = diff.split("\n")[0]
+        if path_line.startswith("diff --git"):
```
bhearsum: Unfortunately, this misses all but the first file. Due to the way the diffs are split, we lose the `diff` in the in the subsequent sections.

You can probably fix this by adjusting `diffs` after the splitting happens.

bhearsum (changes requested):

landoscript/tests/conftest.py
SoundBot: You are correct! I made changes. 

Unrelated to that, I found and fixed some cases where the sorting order was incorrect

bhearsum (approved):
Nice work, thank you for working through all of this!

bhearsum:
> P.S. Do contributions to this repository get an acknowledgment in the FF release notes? I'm talking about community contributions [here](https://www.mozilla.org/en-US/firefox/137.0/releasenotes/?utm_source=firefox-browser&utm_medium=firefox-desktop&utm_campaign=about-dialog)

I'll look into this; I think it's possible for this type of contribution.

PR #1232, opened 2025-06-13
fix(landoscript): ensure a newline exists between diff sections

bhearsum:
I believe this broke with https://github.com/mozilla-releng/scriptworker-scripts/pull/1203, where `diff_contents` (correctly) stopped always having a newline character at the end of a diff, but callers were not updated to handle it.

Unfortunately, this was masked by a bug in the tests where we split on `diff\n` instead of `\ndiff` - resulting in the diffed files not being split correctly.

### DIFF
--- a/landoscript/src/landoscript/actions/android_l10n_import.py
+++ b/landoscript/src/landoscript/actions/android_l10n_import.py
@@ -90,7 +90,7 @@ async def run(
         log.info(f"fetching original files from l10n repository: {dst_files}")
         orig_files = await github_client.get_files(dst_files, branch=to_branch)
 
-        diff = ""
+        diffs = []
         for l10n_file in l10n_files:
             if l10n_file.dst_name not in orig_files:
                 log.warning(f"WEIRD: {l10n_file.dst_name} not in orig_files, continuing anyways...")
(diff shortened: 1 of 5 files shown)


bhearsum:
Various tests of this have been done on https://treeherder.mozilla.org/jobs?repo=try&revision=b03bedac74ca99784b3ce0b855c5fbf7578fcef8 and https://treeherder.mozilla.org/jobs?repo=try&revision=8261e81eb89b378a5566d8b73a5c8977ea542642 (`early-to-late-beta` is failing due to https://github.com/mozilla-releng/mozilla-taskgraph/pull/106, not anything to do with this patch).

Eijebong (approved):
The patch makes sense, however while reviewing this, I found https://github.com/mozilla-releng/scriptworker-scripts/pull/1232/files#diff-11dcea3f6305f036dc56d6df14db892a02d6291a537bf52c6d5bff6143eb276bR247 which looks wrong, the second condition will always be true, probably should be: `if f"-{before}" in diff and f"+{after}" in diff:`?

bhearsum:
> The patch makes sense, however while reviewing this, I found https://github.com/mozilla-releng/scriptworker-scripts/pull/1232/files#diff-11dcea3f6305f036dc56d6df14db892a02d6291a537bf52c6d5bff6143eb276bR247 which looks wrong, the second condition will always be true, probably should be: `if f"-{before}" in diff and f"+{after}" in diff:`?

Yes, indeed!

PR #1235, opened 2025-06-20
fix: review that CLOBBER file is in diff in merge day tests

bhearsum:
This is something I noticed was missing while reviewing the final form of #1231.

### DIFF
--- a/landoscript/tests/test_merge_day.py
+++ b/landoscript/tests/test_merge_day.py
@@ -3,6 +3,7 @@ import pytest
 from pytest_scriptworker_client import get_files_payload
 from simple_github.client import GITHUB_GRAPHQL_ENDPOINT
 
+from landoscript.actions import merge_day
 from landoscript.script import async_main
 
 from .conftest import (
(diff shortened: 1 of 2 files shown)


landoscript/tests/test_merge_day.py
```diff
@@ -238,20 +244,24 @@ async def test_success_bump_esr(aioresponses, github_installation_responses, con
...
+
     def assert_func(req):
         assert_merge_response(
             context.config["artifact_dir"],
             req,
             expected_actions,
             initial_values,
             expected_bumps,
+            {"CLOBBER": "Merge day clobber 2025-03-03"},
+            {"CLOBBER": "Merge day clobber 2025-06-1"},
```
jcristau: shouldn't that be 2025-06-10?

bhearsum: yes, sigh...and now I get to figure out why the test didn't fail....(other tests did when I messed with the date..)

bhearsum: ah; it's because this substring match still works which is unfortunate...

jcristau (approved):

PR #1236, opened 2025-06-24
Support hardened signing in vpn_behavior and add sign_pkg_behavior

oskirby:
Over in the VPN land, we have been trying to update the signing behavior so that we can get better control over the application bundle which binaries need entitlements or not. This led us to try and switch to the `sign_hardened_behavior` for generating a signed application bundle and it mostly seems to have worked great.

Unfortunately, where it fell down for us was in the generation of the macOS installer package. The Mozilla VPN installer package needs some additional manipulation before notarization that isn't present in the Gecko behavior. We tried to work around this with a repackage task, but that leaves the package's internals unsigned so I think we need to fix it by making a modification to `iscript` instead.

What we try to do in this PR is add support for the hardened signer. We do this by checking for the existence of a `hardened-sign-config` in the task payload. If found, we simply call `sign_hardened_behavior(config, task, create_pkg=False)` to sign the application bundle and then use the existing packaging and notarization proceedures for the VPN.

The old VPN signing behavior is retained in the `vpn_legacy_behavior()` method, and gets invoked when `hardened-sign-config` is absent, which should support backwards compatibility for tasks that don't specify the hardened signing config.

While I was at it, I also attempted to add a new `sign_pkg_behavior()` which I think should also solve the problem by running `productsign` directly on the repackaged installer.


iscript/src/iscript/macvpn.py
```diff
@@ -305,7 +296,40 @@ async def vpn_behavior(config, task, notarize=True):
...
+    Raises:
+        IScriptError: on fatal error.
+
+    """
+
+    # Sign the application bundle(s)
+    if "hardened-sign-config" in task["payload"]:
+        top_apps = await sign_hardened_behavior(config, task, create_pkg=False)
+    else:
+        top_apps = await vpn_legacy_behavior(config, task)
```
hneiva, line 322: I don't think we need to support the legacy behaviour, IMO it should just use hardened signing + create the pkg the way we need it.

iscript/src/iscript/macvpn.py
```diff
@@ -305,7 +296,40 @@ async def vpn_behavior(config, task, notarize=True):
...
+
+
+async def vpn_behavior(config, task, notarize=True):
+    """Notarize vpn app.
+
+    Workflow:
+    . Sign all inner apps
+    . Sign main app
+    . Create pkg
+    . Notarize and staple pkg (optional)
```
hneiva: We're notarizing in a different task. As it stands right now, iscript notarizations are deprecated.

oskirby: This pleases me greatly to know!

iscript/src/iscript/macvpn.py
oskirby: Because the `hardened-sign-config` comes from the task payload, there may be a period of time where tasks get submitted both with and without the updated signing method depending on when PRs get rebased to pick up the change. Therefore I felt it a little smarter to keep support for the legacy behavior.

hneiva (approved):
LGTM!!

PR #1237, opened 2025-06-25
Move all scripts to uv

Eijebong:
This switches all dependency management to use uv instead of having `requirements.txt` for each and every scriptworker.

While it's got a lot of upsides (no more "oops we need some extra deps for mozbuild installed manually for consumers", no more "here's a 15k PR because it's the same 500 lines repeated 30 times when updating dependencies", no more "I need to add scriptworker-client as a dependency to my scriptworker, I'll be back in 24h", no more pip compile), it's also not perfect.

The main drawbacks are the minimum python version is shared between all workers. Which means that we're setting the global minimum python version is 3.8 because of iscript. If you look closely at the lockfile you'll see that it matters for example for mozapkpublisher which has a completely different set of dependencies between the version compatible with 3.8 and the one that isn't. Thankfully, this is lockfile noise only as it will install the highest possible version it can.

The other drawback is that we now cannot have dependencies that contradict eachother between scripts. No "foo > 5" and "foo < 5" at the same time. This shouldn't be an issue and if it ever becomes one there are apparently ways with dependency groups to work around the issue.

#### Try pushes

For bitrisescript:
https://firefox-ci-tc.services.mozilla.com/tasks/groups/R68Muz7HQE6ISikv0B_9uw

For the rest: https://treeherder.mozilla.org/jobs?repo=try&revision=a6ab54fd80a337f337419fd52f100a67a6fc4a4d

(TagBuild failures are because landoscript cannot deal with those on staging without some help since the migration to git, but the script itself failed, so it works™. Had to force some tasks along but all scriptworkers did their job)


bhearsum:
I haven't really looked at the code yet, but I have a few upfront questions:

> The main drawbacks are the minimum python version is shared between all workers. Which means that we're setting the global minimum python version is 3.8 because of iscript.

What sort of impact does this have on our ability to use modern typing? I'm guessing this is worked around by the fact that the individual scripts have their own pyproject.yaml's, most of which are marked as >=3.11, and that's what we use for testing and runtime?

> The other drawback is that we now cannot have dependencies that contradict eachother between scripts. No "foo > 5" and "foo < 5" at the same time. This shouldn't be an issue and if it ever becomes one there are apparently ways with dependency groups to work around the issue.

This will almost certainly come up in the future. Is there a way we can simulate this now, or otherwise ensure that we know we have a straightforward way to workaround it? (I wouldn't want to get into a situation where we need an urgent fix, and have to figure it out on the spot.)

And a totally naive one: can you explain why we need the top-level `pyproject.yaml`? Is this just because of how we choose to do some our CI? Is it to avoid all the dependency repetition? Something else? (Writing this down as a comment in that file might be useful.)

hneiva:
FWIW: the new mac signers are on python 3.11 - and somewhat easier to upgrade python version.

I don't think we should worry too much about the old signers anymore.

Eijebong:
> What sort of impact does this have on our ability to use modern typing? I'm guessing this is worked around by the fact that the individual scripts have their own pyproject.toml's, most of which are marked as >=3.11, and that's what we use for testing and runtime?

This shouldn't be an issue, it only affects dependency resolution because we're forcing both our tests and runners to have a specific python version. So while the pyproject.toml's cannot specify that it wants python 3.11, it is implicitly the case because our docker images say so. It's not great but it's that or we're back to one lockfile per project. (or maybe 2 workspaces but I don't think that's possible without duping a lot of stuff)

> And a totally naive one: can you explain why we need the top-level pyproject.toml

That's so we can declare all projects to be in a single workspace. This is mostly to allow us to have a single lockfile (otherwise we'd have one lockfile per worker). Also it turned out that there are tests at the root (py-init311) so it also takes care of those dependencies as well as the ones for readthedocs.

> This will almost certainly come up in the future. Is there a way we can simulate this now, or otherwise ensure that we know we have a straightforward way to workaround it?

I will have a look at this next week. It's a good point that we'd rather want to figure it out now than during some emergency but the fact that I only found one conflict that got resolved I don't know how long ago and was on a test related (pytest-mock) makes me think that it's somewhat unlikely to happen.

bhearsum:
Ah; with mac signing moving to 3.11, maybe we can wait until we can drop 3.8 to merge this, or perhaps create a new branch to deploy to the 3.8 mac signers if they need to be updated before they are EOL'ed?

bhearsum:
Thanks for all of this.

> > What sort of impact does this have on our ability to use modern typing? I'm guessing this is worked around by the fact that the individual scripts have their own pyproject.toml's, most of which are marked as >=3.11, and that's what we use for testing and runtime?
> 
> This shouldn't be an issue, it only affects dependency resolution because we're forcing both our tests and runners to have a specific python version. So while the pyproject.toml's cannot specify that it wants python 3.11, it is implicitly the case because our docker images say so. It's not great but it's that or we're back to one lockfile per project. (or maybe 2 workspaces but I don't think that's possible without duping a lot of stuff)

Seeing as 3.8 is going away soon, it certainly doesn't seem worth adding lockfiles would have much benefit.

> 
> > And a totally naive one: can you explain why we need the top-level pyproject.toml
> 
> That's so we can declare all projects to be in a single workspace. This is mostly to allow us to have a single lockfile (otherwise we'd have one lockfile per worker). Also it turned out that there are tests at the root (py-init311) so it also takes care of those dependencies as well as the ones for readthedocs.

Right, right, I forgot about the root-level tests. (In some ideal world it seems like we would not have anything at the root level, and all of that stuff would be in `scriptworker_client`, but that is _certainly_ not something to solve here.)

> 
> > This will almost certainly come up in the future. Is there a way we can simulate this now, or otherwise ensure that we know we have a straightforward way to workaround it?
> 
> I will have a look at this next week. It's a good point that we'd rather want to figure it out now than during some emergency but the fact that I only found one conflict that got resolved I don't know how long ago and was on a test related (pytest-mock) makes me think that it's somewhat unlikely to happen.

I guess in the worst case scenario, we could introduce a project-specific lockfile if needed in a pinch? (Which I presume would be very straightforward...)

Eijebong:
> I guess in the worst case scenario, we could introduce a project-specific lockfile if needed in a pinch? (Which I presume would be very straightforward...)

Yeah, actually getting a project out of the workspace and specifying dependencies within the workspace to by path should be enough. I'll give *that* a try as it feels like it'll be something like `+2/-2` (and then +2000 because lockfile dupe but in a pinch we can live with that for a bit)

ahal:
re: typing, you can also do `from __future__ import annotations` from older Pythons and the modern syntax will be allowed.

bhearsum (approved):

Eijebong:
https://github.com/mozilla-releng/scriptworker-scripts/commit/c5e9a992bf17ba0b8bc0efb6737010a3e0af56a6 this is what's needed to get a script out of the workspace. Pushed that on dev, ran a promotion for firefox-ios https://firefox-ci-tc.services.mozilla.com/tasks/groups/cQ05ieXlQ8S0qVk9T21EzA which went good. So pretty easy to do in a hurry if needed.

PR #1238, opened 2025-06-25
fix: don't re-use lando status from failed landoscript tasks

bhearsum:
This allows reruns to work as sheriffs expected them to.


jcristau (approved):
Nice.  I thought you'd have to get the lando job's status, but using the previous task run's status instead seems like it should work fine and be simpler.

PR #1239, opened 2025-06-30
Traverse the github pushlogs in the right direction when looking for a shippable revision

Eijebong:
I completetly missed that when I wrote this but github returns commits from old to new and we need to traverse them in reverse order to avoid trying to ship the latest available candidate.

The test I added here is based on real data that the github API returned for the one release where we detected that.


jcristau (approved):
Thanks.

PR #1241, opened 2025-07-02
Remove duplicated dockerfiles

### LINKED ISSUE #1094: get rid of duplicate dockerfiles
The Dockerfiles we use to build and deploy scripts with are in `taskcluster/docker`. We still have other versions, eg: https://github.com/mozilla-releng/scriptworker-scripts/blob/master/pushapkscript/Dockerfile that are entirely separate, and end up getting out of date and causing confusion. We should get rid of these ones.

This may involve adjusting tests and/or other consumers of them to use `taskgraph build-image` instead of `docker build`.

Eijebong:
I removed their .test counterpart in #1237 and I'm pretty sure that those Dockerfiles are also unused. Since they're more confusing than anything else, let's remove them and see if something ends up screaming

Fixes #1094


bhearsum (approved):
Based on my (somewhat inadequate) comment in https://github.com/mozilla-releng/scriptworker-scripts/issues/1094, this seems like it should be fine.

If CI passes here, and in a deployment to a `dev` branch, I can't imagine anything else that would use these.

Thank you!

Eijebong:
Well actually I just realized that they were broken anyway from the uv move 😂 So we know they're not used

PR #1245, opened 2025-07-02
Remove all docker-compose files

Eijebong:
Those were relying on Dockerfiles that were broken by #1237 and removed in #1241

I'm not even sure what the purpose of this was given the hard part about running those scripts locally is getting credentials/environment variables correctly which none of this was doing.

And it's easy enough to pull an image from CI if required if one needs to do that for some reason


bhearsum (approved):
The earliest mention of these I can find is in https://github.com/mozilla-releng/beetmoverscript/pull/234, which contains no justification for them.

jcristau (approved):

PR #1247, opened 2025-07-04
Bug 1858573 - Add library constraints settings in iscript

hneiva:


iscript/src/iscript/hardened_sign.py
```diff
@@ -35,16 +35,20 @@
...
-        entitlement_urls.add(cfg["entitlements"])
+    # Get unique entitlement/libconstraint urls
+    resource_urls = set()
+
+    # Collect urls from config
+    for resource_key in ("entitlements", "libconstraints"):
+        for cfg in hardened_sign_config:
+            if not cfg.get(resource_key, None):
+                continue
+            resource_urls.add(cfg[resource_key])
```
bhearsum, line 46: I realize this is only adding on an additional URL, but I'm interested to know why we feed URLs here instead of relying on upstream artifacts, as we do for other inputs. Downloading these at runtime in this manner means these artifacts are not part of chain of trust verification. Eventually, it will also mean that we're downloading raw files from Github, which will almost certainly end up getting rate limited.

This isn't something that needs to be fixed as part of this work, but unless there's a very strong reason we need to do it like this, please file something for converting this to use upstream artifacts.

iscript/src/iscript/hardened_sign.py
```diff
@@ -35,16 +35,20 @@
...
+
     # Async download and set url -> file location mapping
     url_map = {}
     futures = []
-    for url in entitlement_urls:
+    for url in resource_urls:
         filename = url.split("/")[-1]
```
bhearsum, line 53: AFAICT, it's possible (albeit unlikely) that we could overwrite files in this loop. Please add a check to make sure that the filenames will be distinct, to avoid this potentially confusing footgun.

bhearsum (changes requested):

iscript/src/iscript/hardened_sign.py
hneiva: Nice catch! Fixed

iscript/src/iscript/hardened_sign.py
hneiva: Filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1976717

bhearsum (approved):

PR #1248, opened 2025-07-07
Bug 1975771 - Make sure scriptworkers get the memo about having to stop

Eijebong:
If k8s tells a worker to stop before the worker has setup its signal handlers, said worker would never receive the USR1 one and continue picking jobs despite being on a timer. If a task is still running when the timer expires, pre-stop would exit and k8s would end up sending a sigterm to the task which results in `Automation Error: python exited with signal -15`.

This changes the pre-stop behavior to continue sending the signal while it's waiting. This is safe to do as the receiving side is idempotent.

This goes with https://github.com/mozilla-releng/scriptworker/pull/705


Eijebong:
Tested this by pushing the bitrise worker to dev and triggering a staging release for firefox iOS. No spam in logs, the worker received the signal (although it had started within 15s so this didn't verify that the bug is indeed fixed) but everything worked fine so this at least doesn't break everything.

docker.d/pre-stop.sh
```diff
@@ -30,13 +30,11 @@ POLL_INTERVAL=5
...
 started=$(date +%s)
 
-echo "Sending SIGUSR1 signal to process id $SCRIPTWORKER_PID"
 
-kill -s USR1 $SCRIPTWORKER_PID
 
 while true; do
-    echo "checking status..."
-    kill -0 $SCRIPTWORKER_PID
+    echo "Sending SIGUSR1 signal to process id $SCRIPTWORKER_PID"
```
jcristau: nit:
```suggestion
    echo "Sending SIGUSR1 to process id $SCRIPTWORKER_PID"
```

Also, I'm wondering how useful it is to be repeating that line every 5 seconds?

jcristau (approved):

docker.d/pre-stop.sh
Eijebong: True, I left the line because it was spamming before anyway but I guess there's not much value in it. The pre-stop hook logs are all emitted at the same time when it fails anyway (and the other logs continue to get output at the same time) so everything is confusing anyway.

Eijebong: Moved it out of the loop and removed the line that spammed about "waiting 5s" too

PR #1250, opened 2025-07-07
Update local testing docs for signingscript

Eijebong:
This is based on my experience running a
`shippable-l10n-signing-linux64` task locally. While it doesn't cover all use cases it's better than what was there before


bhearsum (approved):
Thank you!

PR #1253, opened 2025-07-08
Remove unused `deps` from tox.ini

jcristau:
Also sync the ruff version between pre-commit and uv.lock to avoid split brain.


Eijebong (approved):
We'll 100% forget but I guess CI will cry when that happens so 🤷

PR #1254, opened 2025-07-08
Respect the uv.lock file when installing packages in docker container

### LINKED ISSUE #1251: Container builds don't respect uv.lock
We're not copying `uv.lock` when building containers so it's not doing its job... This is the cause for https://bugzilla.mozilla.org/show_bug.cgi?id=1974870 because the lockfile isn't included in the containers, the push to prod that updated said file didn't retrigger an image build so it reused images with old dependencies (including mozapkpublisher)

Eijebong:
This works by copying the workspace definition and the uv.lock file within the container and then syncing only the package that is interesting to us ($SCRIPT_NAME). This also gets rid of the manual installation of scriptworker_client which is now unnecessary and should've been a red flag in the first place but I completely missed it, it will now only be installed when a package requires it, and because it is contained in the workspace, it will just work (tm). We could probably avoid copying scriptworker_client in the container for most scripts but the gain is not worth splitting the image definition between scripts with and without the client lib installed.

Fixes #1251

### DIFF
--- a/taskcluster/docker/base/Dockerfile
+++ b/taskcluster/docker/base/Dockerfile
@@ -17,20 +17,19 @@ ADD --chown=app:app topsrcdir/configloader /app/configloader
 ADD --chown=app:app topsrcdir/docker.d /app/docker.d
 # %include .dockerignore
 ADD --chown=app:app topsrcdir/.dockerignore /app/.dockerignore
+# %include uv.lock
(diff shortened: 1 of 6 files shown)


jcristau:
Presumably we can diff the contents of the images before/after this change?

Eijebong:
Diffs look correct now AFAICT

jcristau (approved):

PR #1257, opened 2025-07-16
Drop support for python 3.8

Eijebong:
As mentioned before, this also drops notarization_poller as it's not used anymore and was deployed on old mac signers for historical reasons. Let's not do that anymore and clean things up since I don't want to add support for python 3.11 for it


bhearsum (approved):

bhearsum (changes requested):
oops, looks like iscript has an issue?

Eijebong:
Oh oops, I didn't see that, will fix

bhearsum (approved):

PR #1258, opened 2025-07-17
Always use a python binary from the python image instead of using uv managed ones

Eijebong:
This feels more appropriate and is more in line with what shipit does

### DIFF
--- a/taskcluster/kinds/docker-image/kind.yml
+++ b/taskcluster/kinds/docker-image/kind.yml
@@ -15,6 +15,7 @@ tasks:
         definition: base-test
         args:
             PYTHON_VERSION: "3.11.9"
+            UV_VERSION: "0.7.15"
 
     # Used by push-image kind
     skopeo: {}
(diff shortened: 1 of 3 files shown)


taskcluster/kinds/docker-image/kind.yml
```diff
@@ -15,13 +15,15 @@ tasks:
...
             PYTHON_VERSION: "3.11.9"
+            UV_VERSION: "0.7.15"
 
     # Used by push-image kind
     skopeo: {}
 
     base:
         args:
             PYTHON_VERSION: "3.11.9"
+            UV_VERSION: "0.7.15"
```
hneiva, line 26: nit: I think we should just pin to `"{major}.{minor}"` instead of patch. This way we automatically get any bug fixes with it.

hneiva (approved):

taskcluster/kinds/docker-image/kind.yml
jcristau: The counterpoint is that there's value in a reproducible build process.

PR #1261, opened 2025-07-22
fix: don't return earlier lando status for merge conflicts

bhearsum:
Merge conflicts end up resolving a run with the `intermittent-task` exception. We want to keep using this exception, because automatic retries are a feature here. However, we need to make sure that when we resume from this run that we start fresh and ignore the earlier lando attempt. To do this, we need to enrich the status artifact we add with a distinct failure reason.
Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1978590

### DIFF
--- a/landoscript/src/landoscript/script.py
+++ b/landoscript/src/landoscript/script.py
@@ -10,6 +10,7 @@ from scriptworker.utils import calculate_sleep_time
 import taskcluster
 from landoscript import lando, treestatus
 from landoscript.actions import android_l10n_import, android_l10n_sync, l10n_bump, merge_day, tag, version_bump
+from landoscript.errors import LandoscriptError, MergeConflictError
 from scriptworker_client.github import extract_github_repo_owner_and_name
 from scriptworker_client.github_client import GithubClient
 from scriptworker_client.utils import retry_async
@@ -123,7 +124,7 @@ async def process_actions(session, context, owner, repo, public_artifact_dir, br
     return lando_actions
 
 
-async def get_status_url_from_earlier_run(session: aiohttp.ClientSession) -> str | None:
+async def get_status_url_from_earlier_run(session: aiohttp.ClientSession) -> lando.LandoStatus | None:
     root_url = os.environ.get("TASKCLUSTER_ROOT_URL")
     task_id = os.environ.get("TASK_ID")
     run_id = os.environ.get("RUN_ID")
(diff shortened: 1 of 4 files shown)


bhearsum:
I'm hoping to prove this out in a staging release, but it's very tricky to force the right conditions to make it happen. I'm going to give it another few tries, but in the meantime this is ready for review.

bhearsum:
I managed to get things lined up in https://firefox-ci-tc.services.mozilla.com/tasks/JEi3A0q3Qz-0Z7rCmXrw-Q/runs/0, where the first run got a merge conflict and the second run ignored the earlier lando job. The second run did fail, but I believe that was due to state issues on the staging lando workers.

landoscript/src/landoscript/script.py
```diff
@@ -150,7 +151,7 @@ async def get_status_url_from_earlier_run(session: aiohttp.ClientSession) -> str
                 return
 
             # `getArtifact` has built in retries, and automatically raises on failure
-            artifactInfo = queue.getArtifact(task_id, run_id, "public/build/lando-status.txt")
+            artifactInfo = queue.getArtifact(task_id, run_id, "public/build/lando-status.json")
```
jcristau, line 154: nit: there's a possible window where the earlier run predates this new code and doesn't have a lando-status.json artifact.  This may not be worth worrying about though.

landoscript/src/landoscript/script.py
```diff
@@ -218,10 +220,27 @@ async def async_main(context):
...
+                        elif isinstance(e, LandoscriptError):
+                            status["failure_reason"] = "landoscript_error"
+                        else:
+                            status["failure_reason"] = "unknown"
+
+                        # re-write status to ensure any updated error is written out
+                        with open(os.path.join(public_artifact_dir, "lando-status.json"), "w+") as f:
+                            f.write(json.dumps(status))
+
+                        raise e
```
jcristau:
```suggestion
                        raise
```

jcristau (approved):
Thanks!

landoscript/src/landoscript/script.py
bhearsum: I guess theoretically that's possible, but as long as this change is rolled out when no lando jobs are running, that seems like it would be exceedingly unlikely? (I guess it's possible we could rerun a lando job hours or days later...but that seems even less likely.)

PR #1262, opened 2025-07-23
Add scriptworker as an extra optional dependency of iscript

Eijebong:


hneiva (approved):

bhearsum:
What's the rationale here? (I don't necessarily disagree, but there's no part of scriptworker that iscript depends on directly...)

hneiva:
This is what we're trying to do via puppet:
```
  exec { "install ${scriptworker_base} iscript":
    command     => 'uv sync --locked --inexact --package iscript --extra scriptworker',
    cwd         => $scriptworker_scripts_clone_dir,
    environment => {
      'VIRTUAL_ENV' => "${scriptworker_base}/.venv",
    },
    user        => $user,
    group       => $group,
    refreshonly => true,
    subscribe   => $ss_deps,
    require     => $ss_deps,
  }
```
This avoids having to maintain both scriptworker and scriptworker-scripts repo locally in the signers.
:)

Eijebong:
Yeah, this way we benefit from the lockfile from scriptworker-scripts to pin scriptworker dependencies. And it simplifies installation a lot. And since it's an optional dependency, unless you actually request it you're not getting it.

PR #1263, opened 2025-07-23
feat: push docker images to GAR repository

bhearsum:
This is to support the migration to Mozcloud. We'll do this in parallel with pushes to dockerhub to avoid one pipeline busting the other.

I've dropped support for the archive tags here, because as far as I know we've never made use of them.

### DIFF
--- a/taskcluster/docker/skopeo/push_image_gar.sh
+++ b/taskcluster/docker/skopeo/push_image_gar.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+set -e
+# Set pipefail so curl failures are caught before the pipe to jq
+set -o pipefail  # This will fail on sh / only works on bash
+
+test $APP
+test $DOCKER_TAG
+test $DOCKER_REPO
+test $MOZ_FETCHES_DIR
+test $TASKCLUSTER_ROOT_URL
+test $TASK_ID
+test $VCS_HEAD_REPOSITORY
+test $VCS_HEAD_REV
+
+echo "=== Generating dockercfg ==="
+PASSWORD_URL=http://taskcluster/secrets/v1/secret/project/releng/scriptworker-scripts/deploy
+mkdir -m 700 $HOME/.docker
+# curl --fail forces curl to return a non-zero exit code if the response isn't HTTP 200 (i.e.: HTTP 403 Unauthorized)
+curl --fail -v $PASSWORD_URL | jq '.secret.gar.dockercfg' > $HOME/.docker/config.json
+chmod 600 $HOME/.docker/config.json
+
+cd $MOZ_FETCHES_DIR
+unzstd image.tar.zst
+
+echo "=== Inserting version.json into image ==="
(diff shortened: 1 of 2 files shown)


taskcluster/docker/skopeo/push_image_gar.sh
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
```
jcristau, line 1: Any reason to not just reuse the existing push_image.sh script, making DOCKER_ARCHIVE_TAG optional and passing the path to the secret as an env variable from the kind?

taskcluster/kinds/push-image-gar/kind.yml
```diff
@@ -0,0 +1,74 @@
...
+            - worker.env
+            - run-on-git-branches
+    dependencies:
+        image: docker-image-{name}
+    worker:
+        taskcluster-proxy: true
+        docker-image: {in-tree: skopeo}
+        max-run-time: 1800
+        env:
+            VCS_HEAD_REPOSITORY: https://github.com/mozilla-releng/scriptworker-scripts
```
jcristau: nit: this should come from the decision task instead of being hardcoded here (the existing kind has the same issue though)

taskcluster/kinds/push-image-gar/kind.yml
```diff
@@ -0,0 +1,74 @@
...
+            TASKCLUSTER_ROOT_URL: "$TASKCLUSTER_ROOT_URL"
```
jcristau: nit: the worker already sets the root url, no need to have it in the payload (again, the existing kind has the same bug)

taskcluster/kinds/push-image-gar/kind.yml
```diff
@@ -0,0 +1,74 @@
...
+            APP: "{name}"
+            DOCKER_REPO: "{name}"
+            VCS_HEAD_REV: "{head_rev}"
+            DOCKER_TAG: "{docker_tag}"
+    run-on-tasks-for: [github-push]
+    run-on-git-branches:
+        # Task-specific branches are added in transform
```
jcristau: this comment is no longer true?

jcristau (commented):
This looks pretty good, thanks!

taskcluster/docker/skopeo/push_image_gar.sh
bhearsum: There will be more changes than this later. In particular, we're going to be starting with google service account credentials, and will need to use them to generate docker creds. (Obviously we could still do this in the same script...)

Once I know exactly how different things will be I'll consider this, though.

taskcluster/kinds/push-image-gar/kind.yml
bhearsum: Good call; let's make that improvement.

taskcluster/kinds/push-image-gar/kind.yml
bhearsum: Indeed!

PR #1265, opened 2025-07-29
complete hack: add support for a signingscript docker-compose.yml

bhearsum:
this is 100% not intended for review or landing at this time...i'm just experimenting with a docker compose for signingscript, and want to see if the changes i've made even build in taskcluster


bhearsum:
this is a bad idea!

PR #1269, opened 2025-08-05
iscript: Allow signing with an explicit codesign identifier

oskirby:
## Description

We have a bug in the VPN project that we introduced when attempting to split the client binary and the daemon. The problem occurs because macOS's service database relies on the designated requirement of the daemon binary to prevent tampering or replacement of the daemon.

By splitting the daemon into its own binary, we changed its codesign identifier and hence it no longer matches the designated requirements of the old daemon and operating system refuses to launch it. One suggestion from the apple developer forums is to explicitly set the signing identifier to match the main application during signing. I have tested this out with some local builds and it seems to work just fine, so this PR attempts to make this option available using the `iscript` hardened signer behaviour.

## How this change works
The `mac_sign_hardened` can accept a new optional value in the `config` dictionary named `identifier`. When present, the arguments `--identifier <value>` are added to the `codesign` command invocation, otherwise the `codesign` arguments should be unaffected when this value is missing.

## References
Apple developer forums: [Upgrading an SMAppService daemon and changing the plist ](https://developer.apple.com/forums/thread/795022)
JIRA Issue: [VPN-7191](https://mozilla-hub.atlassian.net/browse/VPN-7191)

### DIFF
--- a/iscript/src/iscript/data/i_task_schema.json
+++ b/iscript/src/iscript/data/i_task_schema.json
@@ -131,6 +131,9 @@
                   "libconstraints": {
                     "type": "string"
                   },
+                  "identifier": {
+                    "type": "string"
+                  },
                   "globs": {
                     "type": "array",
                     "minItems": 1,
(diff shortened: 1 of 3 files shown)


oskirby:
After some thinking, I don't think that forcing the signing identifer is really the solution to the VPN bug. The problem being that it's too little too late and we have already released the buggy version to some users. We are going to address the problem another way.

PR #1270, opened 2025-08-06
feat(iscript) Bug 1976717 - Use upstream artifacts for signing resources

hneiva:


hneiva:
Tested on local mac signer

iscript/tests/test_hardened_sign.py
```diff
@@ -47,6 +47,22 @@ async def test_download_signing_resources(mocker):
     await hs.download_signing_resources(hs_config, Path("fakefolder"))
 
 
+def test_get_upstream_signing_resources(tmpdir):
```
bhearsum: This test seems to test this specific function just fine. I don't see any coverage that would ensure that the new `sign_hardened_behavior` is tested. If we have existing tests for that, some updates seem like they would be warranted. There are two notable changes there:
1) remote resources are fetched per app
2) `sign_config_files` may come from upstream and remote files
2) upstream resources and remote resources are combined, and must get chosen correctly.

iscript/src/iscript/hardened_sign.py
```diff
@@ -28,11 +28,41 @@
...
+    # Map out paths to their absolute locations
+    signing_resources = {}
+    for resource_path in resource_paths:
+        # Resolve the path relative to the work directory
+        abs_path = Path(get_artifact_path(task_id, resource_path, work_dir=work_dir))
+        if not abs_path.exists():
+            raise IScriptError(f"Signing resource {abs_path} does not exist")
+        if not abs_path.is_file():
+            raise IScriptError(f"Signing resource {abs_path} is not a file")
+        signing_resources[resource_path] = abs_path.resolve()
```
bhearsum, line 62: If I'm reading things correctly, `resource_path` ends up being something like `public/build/entitlements.xml` (at least, it does in the new test you've written). This eventually gets passed as `file_map` to `build_sign_command`, which [expects keys like `entitlements`](https://github.com/mozilla-releng/scriptworker-scripts/blob/2b602fea768a0f83eef61560a407b4f2e76bf374/iscript/src/iscript/hardened_sign.py#L144).

It seems like either the new test is wrong, or there's a real bug here. In either case, having a more end to end test of this functionality seems like it would be helpful.

bhearsum (commented):
Good idea to support both external and upstream signing resources -- it should make migrating to this safer, and it avoids the need to backport everywhere.

iscript/src/iscript/mac.py
```diff
@@ -478,9 +479,18 @@ def get_app_paths(config, task):
     all_paths = []
     for upstream_artifact_info in task["payload"]["upstreamArtifacts"]:
         for subpath in upstream_artifact_info["paths"]:
-            orig_path = get_artifact_path(upstream_artifact_info["taskId"], subpath, work_dir=config["work_dir"])
             formats = upstream_artifact_info["formats"]
-            app = App(orig_path=orig_path, formats=formats, artifact_prefix=_get_artifact_prefix(subpath))
+            if not formats:
+                # Signing resources don't have a signing format
+                continue
```
ahal, line 485: Isn't it better for this case to raise rather than silently skip?

ahal (approved):
Lgtm, though I'm a bit confused about why we're allowing empty formats

iscript/tests/test_hardened_sign.py
hneiva: Added more tests on `test_sign_hardened_behavior`

iscript/src/iscript/hardened_sign.py
hneiva: `signing_resources` in this function would be something like:
```python
{"public/build/security/browser_entitlement.xml": "/some/absolute/path/123/browser_entitlement.xml"}
```
The final use for it is where you linked: 
```python
cmd.append(file_map[config["entitlements"]])  # file_map here is equivalent to signing_resouces
```
Where in the above example `config["entitlements"]` should be `public/build/security/browser_entitlement.xml`
Then `file_map[config["entitlements"]]` should point to the absolute path of the file

iscript/src/iscript/mac.py
hneiva: By default, iscript used to sign all items in `upstreamArtifacts`.
With this change, items with an empty `formats` field will be skipped and not signed.
This allows signing resources (entitlements, libconstraints, etc) to be included in `upstreamArtifacts` for downloading from the build task without having iscript sign them.

bhearsum (approved):

PR #1273, opened 2025-08-26
pushmsixscript: wait for PreProcessing or CommitFailed status (bug 1985285)

jcristau:
It seems we've now started receiving PendingCommit status, and updates are getting lost somewhere.

### DIFF
--- a/pushmsixscript/src/pushmsixscript/microsoft_store.py
+++ b/pushmsixscript/src/pushmsixscript/microsoft_store.py
@@ -273,7 +273,7 @@ def _wait_for_commit_completion(config, channel, session, submission_id, headers
     log.info(response_json.get("status"))
     attempts = 1
     # TODO: Confirm what the status is for a complete submission!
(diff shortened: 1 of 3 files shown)


pushmsixscript/src/pushmsixscript/microsoft_store.py
```diff
@@ -273,7 +273,7 @@ def _wait_for_commit_completion(config, channel, session, submission_id, headers
     log.info(response_json.get("status"))
     attempts = 1
     # TODO: Confirm what the status is for a complete submission!
```
Eijebong, line 275: Nit: Feels weird to remove the TODO in tests but not here

Eijebong (approved):

pushmsixscript/src/pushmsixscript/microsoft_store.py
jcristau: I figured having both was redundant, but given the vagueness of the docs I'm still not convinced we have this right...

PR #1274, opened 2025-08-27
Update dependencies

ahal:


pyproject.toml
```diff
@@ -14,7 +14,6 @@ members = [
     "githubscript",
     "iscript",
     "landoscript",
-    "notarization_poller",
```
Eijebong, line 17: Oops, is that something I forgot when I removed it? Surprised nothing was complaining about it

Eijebong (approved):

PR #1276, opened 2025-08-28
feat(pushmsix): Make scopes dynamic for firefox/comm

hneiva:
+ also adds esr as an allowed channel (tb will use it)


pushmsixscript/docker.d/init_worker.sh
```diff
@@ -16,6 +16,7 @@ export TOKEN_RESOURCE=https://manage.devcenter.microsoft.com
 export STORE_URL=https://manage.devcenter.microsoft.com/v1.0/my/applications/
 export MOCK_APPLICATION_ID=MOCK-APP-ID
 export RELEASE_ROLLOUT_PERCENTAGE=25.0
+export MSIX_SCOPES_PREFIX=project:releng:microsoftstore:
```
jcristau: Move this to the `firefox` case below

pushmsixscript/docker.d/worker.yml
```diff
@@ -12,6 +12,7 @@ application_ids:
     beta: { "$eval": "BETA_APPLICATION_ID" }
     release: { "$eval": "RELEASE_APPLICATION_ID" }
   else: {}
+msix_scopes_prefix: { "$eval": "MSIX_SCOPES_PREFIX" }
```
jcristau: The other scripts use `taskcluster_scope_prefix: { "$eval": "TASKCLUSTER_SCOPE_PREFIX" }`.
I'd suggest we do the same here.

jcristau (approved):

PR #1277, opened 2025-08-29
fix(landoscript): Replace ignore_config with dataclass (bug 1968299)

elkal98:
This required adjusting the consumer as well


bhearsum:
In reviewing this, I realized that due to the dynamic nature of the keys, that we can't enhance the typing here at all. This patch moves things around a bit, but ultimately doesn't provide any better guarantees than the current code. Unfortunately, I'll have to close this (and the bug).

Sorry for your wasted time :(. If you'd like to look at a different issue in this repository, https://bugzilla.mozilla.org/show_bug.cgi?id=1968302 may be a good choice.

PR #1279, opened 2025-09-04
signingscript: drop support for old authenticode certs

jcristau:
The corresponding private keys were held in AWS and not exportable, so now that autograph moved to GCP they can't be used anymore.


bhearsum (approved):
I was initially surprised not to see removals for the prod blocks...but it looks like we never had relevant formats in there? It looks like the formats you're removing never should've been added to those blocks, seeing as they're GCP blocks...

jcristau:
> I was initially surprised not to see removals for the prod blocks...but it looks like we never had relevant formats in there? It looks like the formats you're removing never should've been added to those blocks, seeing as they're GCP blocks...

In the dep case there aren't non-exportable keys so gcp vs aws were the same, I think.

PR #1281, opened 2025-09-09
chore: update dependencies

ahal:


uv.lock
```diff
@@ -1545,14 +1571,14 @@ wheels = [
 
 [[package]]
 name = "httplib2"
-version = "0.22.0"
+version = "0.30.0"
```
Eijebong, line 1574: That's a fun bump. They went from 0.22 -> 0.30 with nothing in between

Eijebong: Something to keep in mind for sure if wee see issues. The changes were a lot of removals and it looks like it at least broke proxy support (which we don't use).

Eijebong (approved):

uv.lock
ahal: Thanks for looking into it.

PR #1282, opened 2025-09-10
GithubClient.get_file_listing now supports multiple file paths at once

Mohmn:


scriptworker_client/tests/test_github_client.py
```diff
@@ -284,12 +284,14 @@ async def test_get_branch_head_oid_branch_not_found(aioresponses, github_client)
...
 async def test_get_repository_files(aioresponses, github_client):
     branch = "main"
-    expected = [
-        "file1",
-        "file2",
-        "a/b/bfile1",
-        "a/b/c/d/e/deepfile1",
-    ]
+    expected = {
+        "": [
```
bhearsum: We shouldn't be changing the return value to fix this; just what's done inside of the function.

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -192,7 +192,7 @@ async def get_branch_head_oid(self, branch: str) -> str:
 
         return repo["object"]["oid"]
 
-    async def get_file_listing(self, path: str = "", branch: Optional[str] = None, depth_per_query=10) -> List[str]:
+    async def get_file_listing(self, paths: Union[str, List[str]] = "", branch: Optional[str] = None, depth_per_query=10) -> Dict[str, List[str]]:
```
bhearsum: We shouldn't be changing the return value to fix this; just what's done inside of the function.

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -252,20 +242,42 @@ async def get_file_listing(self, path: str = "", branch: Optional[str] = None, d
...
 
-        # Process the returing entries
-        entries = resp["repository"]["object"].get("entries")
-        files, refetches = self._process_file_listings(entries, prefix=Path(path))
-        # Any subtrees that were not fully traversed will be returned in `refetches`
-        # We need to refetch data starting at each of these subtrees to ensure we
-        # don't miss anything.
-        for refetch in refetches:
-            files.extend(await self.get_file_listing(str(refetch), branch, depth_per_query))
+        results: Dict[str, List[str]] = {}
```
bhearsum: If I'm reading correctly, we'd avoid the return type change if this returned the combined values of all entries of this dict. Preferably, we would sort them before returning as well.

scriptworker_client/tests/test_github_client.py
```diff
@@ -9,7 +9,7 @@
 from gql.transport.exceptions import TransportQueryError
 from simple_github.client import GITHUB_GRAPHQL_ENDPOINT
 from yarl import URL
-
+from graphql import strip_ignored_characters
```
bhearsum, line 12: TIL - this looks a bit nicer than the `dedent` we're currently using. Please add it to `dev-dependencies` in `pyproject.toml` to make sure it's explicitly declared.

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -252,20 +242,42 @@ async def get_file_listing(self, path: str = "", branch: Optional[str] = None, d
         for _ in range(depth_per_query - 1):
             file_expr = recursive_expr.substitute(obj_expr=file_expr)
 
-        str_query = query.substitute(owner=self.owner, repo=self.repo, branch=branch, path=path, file_expr=file_expr)
+        # Build multiple path queries with aliases
```
bhearsum, line 249: The aliases are needed because `path` may have characters that aren't allowed in keys, I presume? If so, please add a comment to that effect so we remember why in the future.

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -252,20 +242,42 @@ async def get_file_listing(self, path: str = "", branch: Optional[str] = None, d
...
-        for refetch in refetches:
-            files.extend(await self.get_file_listing(str(refetch), branch, depth_per_query))
+        results: Dict[str, List[str]] = {}
+        repo_data = resp["repository"]
+
+        for i, path in enumerate(paths):
+            alias = f"path{i}"
+            entries = repo_data[alias].get("entries")
+            files, refetches = self._process_file_listings(entries, prefix=Path(path))
+            for refetch in refetches:
```
bhearsum: Now that `get_file_listings` supports multiple paths at once, we don't need to call it once per refetch, we can make one call that includes all of them.

bhearsum (changes requested):
This is a good start, and most of the way there! There's some notes inline below with some necessary changes, but they should be fairly straightforward.

Also, as you probably realized already, you'll need similar test changes in `landoscript` to fix the test failures there.

scriptworker_client/tests/test_github_client.py
Mohmn: done, reverted back to org

scriptworker_client/src/scriptworker_client/github_client.py
Mohmn: Reverted dict usage; switched back to list

scriptworker_client/src/scriptworker_client/github_client.py
Mohmn: oops, how did i miss this

scriptworker_client/src/scriptworker_client/github_client.py
Mohmn: added comment

Mohmn:
> This is a good start, and most of the way there! There's some notes inline below with some necessary changes, but they should be fairly straightforward.
> 
> Also, as you probably realized already, you'll need similar test changes in `landoscript` to fix the test failures there.

I'm getting a KeyError in the `Lando script`, even though I haven't made any changes to the get_files function
im a litter confused here
Screenshot from 2025-09-12 02-51-40

bhearsum:
> > This is a good start, and most of the way there! There's some notes inline below with some necessary changes, but they should be fairly straightforward.
> > Also, as you probably realized already, you'll need similar test changes in `landoscript` to fix the test failures there.
> 
> I'm getting a KeyError in the `Lando script`, even though I haven't made any changes to the get_files function im a litter confused here Screenshot from 2025-09-12 02-51-40

This might be because there's now a different number of graphql responses happening. Tests [like this](https://github.com/mozilla-releng/scriptworker-scripts/blob/7e1a19dfeb37f012c08c4611d7dc762d5ab2f889/landoscript/tests/test_android_l10n_import.py#L421-L499) are [setting up two responses](https://github.com/mozilla-releng/scriptworker-scripts/blob/7e1a19dfeb37f012c08c4611d7dc762d5ab2f889/landoscript/tests/test_android_l10n_import.py#L513); if everything is returned in one response with your change, you'll need to adjust that to mock only one response.

If this is what's happening, `get_files` ends up seeing the second `get_file_listings` mocked response.

PR #1283, opened 2025-09-14
feat: GithubClient.get_file_listing now supports multiple file paths at once

Mohmn:


Mohmn:
Created a new PR because I accidentally pushed my GitHub app’s PEM key in the previous one.

bhearsum:
> Created a new PR because I accidentally pushed my GitHub app’s PEM key in the previous one.

Oops. You should make sure you revoke that; it may have been downloaded already, or still available somewhere even if it's not visible in the PR.

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -252,18 +246,51 @@ async def get_file_listing(self, path: str = "", branch: Optional[str] = None, d
...
 
         # Process the returing entries
-        entries = resp["repository"]["object"].get("entries")
-        files, refetches = self._process_file_listings(entries, prefix=Path(path))
-        # Any subtrees that were not fully traversed will be returned in `refetches`
-        # We need to refetch data starting at each of these subtrees to ensure we
-        # don't miss anything.
-        for refetch in refetches:
-            files.extend(await self.get_file_listing(str(refetch), branch, depth_per_query))
+        # # Any subtrees that were not fully traversed will be returned in `refetches`
```
bhearsum: nit: please get rid of the extra ` #` here.

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -210,19 +210,13 @@ async def get_file_listing(self, path: str = "", branch: Optional[str] = None, d
...
-              repository(owner: "$owner", name: "$repo") {
-                object(expression: "$branch:$path") {
-                  $file_expr
-                }
-              }
-            }
-                """
-            )
-        )
+        if isinstance(paths, list):
```
bhearsum: Was this block added to handle nested calls where there may not be any subentries? If so, can we avoid making the nested call in that case instead?

It would be preferable to let an exception bubble up for any outside caller that doesn't provide any paths to fetch listings for; calling without a path to fetch for is not a valid invocation.

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -252,18 +246,51 @@ async def get_file_listing(self, path: str = "", branch: Optional[str] = None, d
...
+            node = repo_data.get(alias)
+            if not node:
+                continue
+            entries = node.get("entries")
+            files_, refetches_ = self._process_file_listings(entries, prefix=Path(path))
+            refetches.extend([str(reftch) for reftch in refetches_])
+            files.extend(files_)
+
+        if refetches:
+            refetch_flies = await self.get_file_listing(refetches, branch, depth_per_query)
```
bhearsum: nit: `refetch_files` instead of `refetch_flies`.

Mohmn:
> > Created a new PR because I accidentally pushed my GitHub app’s PEM key in the previous one.
> 
> Oops. You should make sure you revoke that; it may have been downloaded already, or still available somewhere even if it's not visible in the PR.

Yes I already revoked it

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -192,7 +192,7 @@ async def get_branch_head_oid(self, branch: str) -> str:
 
         return repo["object"]["oid"]
 
-    async def get_file_listing(self, path: str = "", branch: Optional[str] = None, depth_per_query=10) -> List[str]:
+    async def get_file_listing(self, paths: Union[str, List[str]] = "", branch: Optional[str] = None, depth_per_query=10) -> List[str]:
```
bhearsum: I tested this patch in [a real world job](https://treeherder.mozilla.org/jobs?repo=try&revision=7f77bf9dac8e9214be8d113398109f7565960a08&selectedTaskRun=Ql0bArOIS1iLRUXyGy03Kw.0). It ended up failing due to the graphql query timing out. Moving the `depth_per_query` down to `5` fixed that. Let's make that the new default, since this function will often do more work now.

bhearsum (changes requested):
This looks wonderful overall! My test of it in a real world job with `depth_per_query` lowered [worked fine](https://treeherder.mozilla.org/jobs?repo=try&revision=7f77bf9dac8e9214be8d113398109f7565960a08&selectedTaskRun=Ql0bArOIS1iLRUXyGy03Kw.1), and seems to have sped up the querying in this job by ~20%.

There's a few notes inline with a few adjustments that are needed, but they're all minor things.

scriptworker_client/src/scriptworker_client/github_client.py
Mohmn: cool :fast_forward:

scriptworker_client/src/scriptworker_client/github_client.py
Mohmn: raised expection for the outside called

bhearsum (changes requested):
This looks really good, and addresses everything that I commented on previously.

Unfortunately, with all of the `refetches` being passed at once we've uncovered a new scaling issue that causes a `Timeout on validation of query` error to be thrown in nested requests (full log available at https://firefoxci.taskcluster-artifacts.net/ZWd99bW1T7i45rN3IQZeQQ/5/public/logs/live_backing.log).

As best I can tell, this is happening because we're requesting too many nested entries - even with `depth_per_query` set to 1.

My suggestion to try to fix this is to add a new `paths_per_query` parameter, and add any `paths` greater than this to `refetches`. This should handle the initial `paths` being greater than that number, and deal with any nested calls as well.

Mohmn:
> This looks really good, and addresses everything that I commented on previously.
> 
> Unfortunately, with all of the `refetches` being passed at once we've uncovered a new scaling issue that causes a `Timeout on validation of query` error to be thrown in nested requests (full log available at https://firefoxci.taskcluster-artifacts.net/ZWd99bW1T7i45rN3IQZeQQ/5/public/logs/live_backing.log).
> 
> As best I can tell, this is happening because we're requesting too many nested entries - even with `depth_per_query` set to 1.
> 
> My suggestion to try to fix this is to add a new `paths_per_query` parameter, and add any `paths` greater than this to `refetches`. This should handle the initial `paths` being greater than that number, and deal with any nested calls as well.

I’m planning to send  requests with smaller payload ```(8 paths at a time)``` and run ```N(at max 4-5)``` of them concurrently.
I’m also considering replacing ```recursion with a BFS queue```, which would make it easier to run concurrent code .  how does it sound to you ?

bhearsum:
> > This looks really good, and addresses everything that I commented on previously.
> > Unfortunately, with all of the `refetches` being passed at once we've uncovered a new scaling issue that causes a `Timeout on validation of query` error to be thrown in nested requests (full log available at https://firefoxci.taskcluster-artifacts.net/ZWd99bW1T7i45rN3IQZeQQ/5/public/logs/live_backing.log).
> > As best I can tell, this is happening because we're requesting too many nested entries - even with `depth_per_query` set to 1.
> > My suggestion to try to fix this is to add a new `paths_per_query` parameter, and add any `paths` greater than this to `refetches`. This should handle the initial `paths` being greater than that number, and deal with any nested calls as well.
> 
> I’m planning to send requests with smaller payload `(8 paths at a time)` and run `N(at max 4-5)` of them concurrently. I’m also considering replacing `recursion with a BFS queue`, which would make it easier to run concurrent code . how does it sound to you ?

I think concurrency here would actually be worse; we're already hitting timeouts, and adding concurrency will run through tokens quicker, and possibly trip other DoS measures from Github. (To the best of my knowledge, this function is entirely serial at the moment, despite being `async`.)

Mohmn:
> > > This looks really good, and addresses everything that I commented on previously.
> > > Unfortunately, with all of the `refetches` being passed at once we've uncovered a new scaling issue that causes a `Timeout on validation of query` error to be thrown in nested requests (full log available at https://firefoxci.taskcluster-artifacts.net/ZWd99bW1T7i45rN3IQZeQQ/5/public/logs/live_backing.log).
> > > As best I can tell, this is happening because we're requesting too many nested entries - even with `depth_per_query` set to 1.
> > > My suggestion to try to fix this is to add a new `paths_per_query` parameter, and add any `paths` greater than this to `refetches`. This should handle the initial `paths` being greater than that number, and deal with any nested calls as well.
> > 
> > 
> > I’m planning to send requests with smaller payload `(8 paths at a time)` and run `N(at max 4-5)` of them concurrently. I’m also considering replacing `recursion with a BFS queue`, which would make it easier to run concurrent code . how does it sound to you ?
> 
> I think concurrency here would actually be worse; we're already hitting timeouts, and adding concurrency will run through tokens quicker, and possibly trip other DoS measures from Github. (To the best of my knowledge, this function is entirely serial at the moment, despite being `async`.)

My bad  I didn’t think about the token limit.  
I added   ```paths_per_query param``` as per your suggestion

bhearsum (approved):
Thank you! I'm probably going to tweak the default value for `paths_per_query`, before merging after some experiment, but this otherwise looks good. I appreciate you reworking and adjusting this!

PR #1284, opened 2025-09-15
feat(pushmsix):Add ESR_APPLICATION_ID for Thunderbird

stephenking314:


pushmsixscript/docker.d/worker.yml
```diff
@@ -11,6 +11,7 @@ application_ids:
   then:
     beta: { "$eval": "BETA_APPLICATION_ID" }
     release: { "$eval": "RELEASE_APPLICATION_ID" }
+    esr: { "$eval": "ESR_APPLICATION_ID" }
```
hneiva:
```suggestion
    esr:
      $if: 'COT_PRODUCT == "thunderbird"
      then: { "$eval": "ESR_APPLICATION_ID" }
      else: null    
```

I think we just need to prevent the key from being calculated for firefox.

hneiva (changes requested):

pushmsixscript/docker.d/init_worker.sh
```diff
@@ -25,7 +25,8 @@ case $COT_PRODUCT in
     ;;
   thunderbird)
     export BETA_APPLICATION_ID=9PFQWVMSS45P
-    export RELEASE_APPLICATION_ID=9PM5VM1S3VMQ
+    export RELEASE_APPLICATION_ID=9MX7TGVGQFCL
```
hneiva, line 28: I just noticed the `RELEASE_APPLICATION_ID` was also changed, is this intentional?

stephenking314: Yeah, this was intentional.   Our existing listing was always actually ESR.   We decided to keep that listing as ESR, label as such, and create a new listing for the standard release channel.

hneiva (approved):
LGTM

PR #1287, opened 2025-09-25
fix: lower `paths_per_query` to 1 in github client

bhearsum:
We're seeing timeouts in android-l10n jobs due to the increased query size caused by https://github.com/mozilla-releng/scriptworker-scripts/pull/1283; this restores the queries to their original size without backing out the feature.

It also fixes a bug where `paths_per_query` was not being forwarded in recursive calls.

### DIFF
--- a/landoscript/src/landoscript/actions/android_l10n_import.py
+++ b/landoscript/src/landoscript/actions/android_l10n_import.py
@@ -64,7 +64,7 @@ async def run(
             if "**" in contents["paths"][0]["reference"]:
                 # localized file paths contain globs; we need that directory
                 # structure to determine the files we need to fetch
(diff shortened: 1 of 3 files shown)


landoscript/src/landoscript/actions/android_l10n_import.py
```diff
@@ -64,7 +64,7 @@ async def run(
             if "**" in contents["paths"][0]["reference"]:
                 # localized file paths contain globs; we need that directory
                 # structure to determine the files we need to fetch
-                force_paths = await l10n_github_client.get_file_listing(str(src_file_prefix), depth_per_query=4, paths_per_query=100)
```
bhearsum, line 67: This wasn't the only cause of the problem, but it wasn't helping. This was my mistake (I manually fiddled with this value in some Try tests ahead of updating and merging the patch.)

ahal (approved):

PR #1292, opened 2025-10-19
Bug 1968280 - Draft: Landoscript should fetch file modes for newly added or files being removed. 

elkal98:
Hi, I'm getting the below error: 

`src/scriptworker_client/github_client.py:176: error: Incompatible types in assignment (expression has type "None", target has type "Union[str, dict[str, Optional[str]]]")  [assignment]`

I think its happening when I set `ret[name] = None`. I tried to change the return type to include `None`, but that ended up creating errors in other places that use this function indirectly. So, can you please let me know how to resolve this or if I need to do any other changes.


landoscript/src/landoscript/util/diffs.py
```diff
@@ -14,7 +14,7 @@ def diff_contents(orig, modified, file):
         # orig does not exist yet; ie: it will be added
         orig_contents = ""
         fromfile = "/dev/null"
-        add_remove_metadata = "new file mode 100644\n"
+        add_remove_metadata = f"new file mode {mode or 100644}\n"
```
bhearsum, line 17: Rather than having a default here, this function should raise an error if `mode` is not set when a file is being newly added or removed.

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -93,7 +93,9 @@ async def _execute():
 
         await retry_async(_execute, attempts=3, retry_exceptions=(TransportQueryError,), sleeptime_kwargs={"delay_factor": 0})
 
-    async def get_files(self, files: Union[str, List[str]], branch: Optional[str] = None, files_per_request: int = 200) -> Dict[str, Union[str, None]]:
+    async def get_files(
+        self, files: Union[str, List[str]], branch: Optional[str] = None, files_per_request: int = 200, mode: Optional[str] = None
```
bhearsum, line 97: This is not quite right - we want a boolean argument called something like `fetch_mode` that defaults to `False`. When True it should:
* Change the query to fetch file modes
* Return file modes in the return value

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -165,7 +175,10 @@ async def get_files(self, files: Union[str, List[str]], branch: Optional[str] =
                 if v is None:
                     ret[name] = None
                 else:
-                    ret[name] = v["text"]
+                    if mode:
+                        ret[name] = {"mode": v.get("mode"), "text": v.get("text")}
+                    else:
+                        ret[name] = v.get("text")
```
bhearsum, line 181: Let's update the value for each file to always be a dict; that will make things easier for callers, and make the function more comprehensible (especially the type signatures).

landoscript/src/landoscript/actions/android_l10n_import.py
```diff
@@ -47,7 +47,7 @@ async def run(
     async with GithubClient(github_config, l10n_owner, l10n_repo) as l10n_github_client:
         toml_files = [info.toml_path for info in android_l10n_import_info.toml_info]
         # we always take the tip of the default branch when importing new strings
-        toml_contents = await l10n_github_client.get_files(toml_files)
+        toml_contents = await l10n_github_client.get_files(toml_files, mode=None)
```
bhearsum, line 50: After you've updated `get_files` as noted below, the `l10n_github_client.get_files` calls in this function and `android_l10n_sync.py` will need to be updated to set `fetch_mode=True`.

You will also need changes to actually pass the modes along to `diff_contents` for each file.

landoscript/src/landoscript/actions/l10n_bump.py
```diff
@@ -66,7 +66,7 @@ async def run(
             files = [bump_config.path, *platform_config_files]
             try:
                 log.info(f"fetching bump files from github: {files}")
-                orig_files = await github_client.get_files(files, branch)
+                orig_files = await github_client.get_files(files, branch, mode=None)
```
bhearsum, line 69: This specific change should not be needed (see my comment elsewhere about the interface of `get_files`. You will need to changes to this and all other `get_files` calls to deal with the fact that it will now be returning dicts for each file, rather than strings.

bhearsum (changes requested):
This is a fine start. Please take a closer read of https://bugzilla.mozilla.org/show_bug.cgi?id=1968280#c0 to make sure you're making all of those changes, most of which I've highlighted in the comments here.

> src/scriptworker_client/github_client.py:176: error: Incompatible types in assignment (expression has type "None", target has type "Union[str, dict[str, Optional[str]]]")  [assignment]

This will probably be fixed as a side effect of my requested changes to the `get_files` signature.

PR #1295, opened 2025-10-22
WIP: explicitly specify location of non-pypi dependencies

bhearsum:
Incomplete patch for now; just doing a quick test for CI first.

### DIFF
--- a/balrogscript/pyproject.toml
+++ b/balrogscript/pyproject.toml
@@ -41,6 +41,9 @@ dev = [
     "pytest-mock",
 ]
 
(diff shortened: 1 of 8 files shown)


bhearsum:
This was related to https://bugzilla.mozilla.org/show_bug.cgi?id=1995826, which showed that `pip` would try to install `scriptworker-client` from pypi. We don't use plain `pip` to do any installation, nor would it be affected by this patch. If we do something to guard against `pip` doing the wrong thing we can open a different PR.

PR #1299, opened 2025-11-13
Bug 1999653 - Add YUM repo config to beetmover for firefox

Eijebong:


beetmoverscript/docker.d/worker.yml
```diff
@@ -93,6 +102,15 @@ clouds:
                 name: 'mozilla'
                 location: 'us'
                 project: 'moz-fx-productdelivery-pr-38b5'
+            product_yum_repos:
+              devedition:
+                name: 'mozilla-rpm'
+                location: 'us'
+                project: 'moz-fx-productdelivery-no-38b5'
```
jcristau: Seems wrong for prod to push to the stage repo, so if this is intended it needs some kind of explanation.

jcristau (changes requested):

beetmoverscript/docker.d/worker.yml
Eijebong: Nope, just me copy pasting, not realizing that no/pr were also different...

Eijebong: Fixed. Thanks for catching that and saving me the headache later

jcristau (approved):

PR #1300, opened 2025-11-21
feat: support optional overwrites in translations uploads

bhearsum:
This was requested because we have [a new upload task in translations](https://github.com/mozilla/translations/pull/1285) whose entire purpose is uploading some data to a consistent path, and a similar use case coming soon.


bhearsum:
This got tested in https://firefox-ci-tc.services.mozilla.com/tasks/bC7kHEipQuakYaChrT9cvg

jcristau (approved):

PR #1312, opened 2025-12-12
Implement a mark-as-merged action for shipitscript (bug 1890753)

Eijebong:
The API is getting implemented as part of https://github.com/mozilla-releng/shipit/pull/1801

### DIFF
--- a/shipitscript/src/shipitscript/script.py
+++ b/shipitscript/src/shipitscript/script.py
@@ -35,6 +35,14 @@ def mark_as_shipped_action(context):
     ship_actions.mark_as_shipped_v2(context.ship_it_instance_config, release_name)
 
 
+def mark_as_merged_action(context):
+    """Action to mark a merge automation as complete"""
+    automation_id = context.task["payload"]["automation_id"]
+
+    log.info("Marking the merge automation as complete ...")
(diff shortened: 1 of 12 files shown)


shipitscript/src/shipitscript/script.py
```diff
@@ -109,6 +117,7 @@ def update_product_channel_version_action(context):
 # ACTION_MAP {{{1
 ACTION_MAP = {
     "mark-as-shipped": mark_as_shipped_action,
+    "mark-as-merged": mark_as_merged_action,
```
ahal, line 120: Did you consider making a single action called "mark-completed" or something that can handle both? Or is that not really viable due to how releases vs merges are handled in shipit?

ahal (approved):

ahal (changes requested):
Ah didn't mean to approve due to the tox failures, but looks good barring that.

shipitscript/src/shipitscript/script.py
Eijebong: It could be possible I guess, although it might get a bit crafty because we don't want to be able to call mark-as-shipped from merge automation for example. And since we rely on scopes for the action instead of the payload, I don't think it's worth adding a second scope onto everything.

ahal (approved):

PR #1314, opened 2025-12-16
ci: user proper worker-type for b-linux

ahal:


abhishekmadan30 (approved):

jcristau (changes requested):
I don't think that pool exists.

PR #1315, opened 2025-12-16
ci(push-image): don't set TASKCLUSTER_ROOT_URL in task env

ahal:


abhishekmadan30 (approved):

bhearsum:
Is this because it wasn't necessary, or because it was causing a problem?

jcristau:
@bhearsum it was causing failures after the switch from docker-worker to d2g pools, because while docker-worker overrides this env variable from the payload (among others: https://github.com/taskcluster/taskcluster/blob/62dc77f1f9116a28d662c08b788575a97d75217e/workers/docker-worker/src/task.js#L352-L368), with d2g we'd get the bogus value and fetch-content would fail with e.g.
> Download failed: unknown url type: '$TASKCLUSTER_ROOT_URL/api/queue/v1/task/QeQkbv00SjGPYoa6w0Gzvw/artifacts/public/image.tar.zst'

PR #1331, opened 2026-01-30
feat(iscript) Bug 1812480 - Change file ownership before creating pkgs

hneiva:


jcristau (changes requested):
This doesn't seem reasonable to me.  Can you expand on what problem this is trying to solve?

PR #1332, opened 2026-02-02
Pin taskcluster >= 92 in landoscript

Eijebong:
This essentially undoes 13aa62cdb28c2eb1ef5342117c92daecd7b7b955. The change that made the async client got reverted in taskcluster 92 so 91 is the only version we need to be careful about.
The pin is technically not necessary but it doesn't cost anything it'll prevent 91 from being selected by the resolved if there's some freak accident where it becomes the only resolvable version. 🤷

### DIFF
--- a/landoscript/pyproject.toml
+++ b/landoscript/pyproject.toml
@@ -19,7 +19,7 @@ dependencies = [
     "moz.l10n",
     "scriptworker",
     "scriptworker-client",
-    "taskcluster<91",
+    "taskcluster>=92",
     "yarl",
 ]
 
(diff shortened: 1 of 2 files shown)


bhearsum (approved):
Thanks!

landoscript/pyproject.toml
```diff
@@ -19,7 +19,7 @@ dependencies = [
     "moz.l10n",
     "scriptworker",
     "scriptworker-client",
-    "taskcluster<91",
+    "taskcluster>91",
```
jcristau: Shouldn't it be `>=92`?

jcristau (changes requested):

landoscript/pyproject.toml
Eijebong: Yes, yes it should... >91 can still pick up 91.0.1...

jcristau (approved):

PR #1334, opened 2026-02-02
Feat(iscript): Bug 2014010 - Retry codesign command

hneiva:
Apple timestamp servers are being flaky and failing ~40% of the time. Hoping retrying the requests fixes it.


ahal (approved):
Thanks, looks reasonable!

Should there be some kind of back off?

hneiva:
the default is a gradual backoff, which is roughly:
5 attempts @ approx: 6, 14, 26, 48, 96 seconds

PR #1337, opened 2026-02-05
Update landoscript merge_day.py to accomodate Thunderbird

dandarnell:
Thunderbird does not have a CLOBBER file, so this PR bypasses that requirement for the merge_day action when run on Thunderbird repos.


bhearsum (changes requested):
Let's make this cleaner and more explicit by making clobber file operations configurable at the payload level rather than hardcoding this magic string in this action. I expect we could add something like `update_clobber_file` to `MergeInfo` that defaults to `true` (to preserve existing behaviour) and override it in your task configuration.

If you need any additional pointers here just let me know.

dandarnell:
> Let's make this cleaner and more explicit by making clobber file operations configurable at the payload level rather than hardcoding this magic string in this action. I expect we could add something like `update_clobber_file` to `MergeInfo` that defaults to `true` (to preserve existing behaviour) and override it in your task configuration.
> 
> If you need any additional pointers here just let me know.

Good point. I've updated the patch to update the CLOBBER file based on `update_clobber_file` in `MergeInfo`, which defaults to `true`.

bhearsum (changes requested):
The [merge_day_payload part of the schema](https://github.com/mozilla-releng/scriptworker-scripts/blob/179752f05e1e1fdb3690634aab17790a3b160a91/landoscript/src/landoscript/data/landoscript_task_schema.json#L5) should get updated for this. (In fact, I think you'll end up with a `MALFORMED_PAYLOAD` if you try to set this without updating it.)

If you want to give this a run through on try feel free to stop by #firefox-ci on Matrix and have someone push it to the dev workers.

I also suggest you add a test to cover this behaviour, but seeing as it's a Thunderbird-specific path I'll leave that up to you and not block on it.

dandarnell:
The schema is now updated.

bhearsum (approved):

PR #1338, opened 2026-02-08
Bug2015223: Adds Thunderbird Release AppID

stephenking314:
Initially existing stable channel will continue to be used for ESR.
 - "app_id" is changed to a dict, "app-ids" to facilitate looking up the app_id by channel.
 - Thunderbird release channel will use org.mozilla.ThunderbirdRelease

Another change down the line will change over the default stable channel "org.mozilla.Thunderbird" to release channel and ESR will use "org.mozilla.ThunderbirdESR".
This is being done in a phased approach to avoid force-switching users to a different release channel.


pushflatpakscript/docker.d/init_worker.sh
```diff
@@ -25,6 +25,12 @@ case $ENV in
     export REPO_TOKEN_STABLE_PATH=$CONFIG_DIR/stable_token.txt
     echo $REPO_TOKEN_BETA | base64 -d > $REPO_TOKEN_BETA_PATH
     echo $REPO_TOKEN_STABLE | base64 -d > $REPO_TOKEN_STABLE_PATH
+
+    if [[$COT_PRODUCT == 'thunderbird']]; then
```
jcristau:
```suggestion
    if [ $COT_PRODUCT = thunderbird ]; then
```

pushflatpakscript/docker.d/worker.yml
```diff
@@ -1,7 +1,13 @@
 work_dir: { "$eval": "WORK_DIR" }
 artifact_dir: { "$eval": "ARTIFACTS_DIR" }
 verbose: { "$eval": "VERBOSE == 'true'" }
-app_id: { "$eval": "APP_ID" }
+app_ids:
+  beta: {"$eval": "APP_ID_BETA" }
+  stable: {"$eval": "APP_ID_STABLE"}
+  esr:
+    $if: 'COT_PRODUCT == "thunderbird"'
```
jcristau: suggestion: instead of hardcoding thunderbird, maybe use `$if: defined("APP_ID_ESR")`, and similar below

pushflatpakscript/src/pushflatpakscript/flathub.py
```diff
@@ -142,18 +142,18 @@ def files(tar, path):
...
     # Consolidate ostree refs into list of Flatpak IDs available in repo
     flatpak_refs = [ref.split("/")[1] for ref in flatpak_refs if ref.startswith("app/")]
 
     # Create a list, if any, of all unexpected Flatpak IDs present in repo
-    invalid_refs = set(flatpak_refs) - {context.config["app_id"]}
+    invalid_refs = set(flatpak_refs) - {context.config["app_ids"]}
 
-    if context.config["app_id"] not in flatpak_refs:
-        raise TaskVerificationError(f"Supplied app ID ({context.config['app_id']}) is not present in Flatpak!")
+    if context.config["app_ids"][channel] not in flatpak_refs:
```
jcristau: Need to check that `channel in context.config["app_ids"]`, since the channel is supplied by the payload?

pushflatpakscript/src/pushflatpakscript/task.py
```diff
@@ -1,6 +1,6 @@
 from scriptworker.exceptions import TaskVerificationError
 
-_CHANNELS_AUTHORIZED_TO_REACH_FLATHUB = ("beta", "stable")
+_CHANNELS_AUTHORIZED_TO_REACH_FLATHUB = ("beta", "stable", "esr")
```
jcristau: Isn't this effectively `config["app_ids"].keys()`?

jcristau (changes requested):

pushflatpakscript/src/pushflatpakscript/task.py
stephenking314: It is, though there may be some value to defining the allowed channels separately as constant, given that it is used for both functions in this file.

pushflatpakscript/docker.d/init_worker.sh
stephenking314: I made this change, but I'm curious, why do we want to prefer the POSIX `[` aka `test` here as opposed to the bash expression `[ ... ]`?  (I know I was missing required whitespace either way)

jcristau: Personally I tend to avoid bashisms unless they add value, which AFAIK `[[` doesn't?

stephenking314: Here I don't think it makes a difference.  In the general sense it can make things easier. `[` allows variable splitting which in turn requires variables to be quoted if that not the intention.   Maybe the larger advantage for string comparison is that `[[` allows pattern matching and regular expressions.  You can also put logical operators inside the expression, but that comes down to preference either way

JKB0DY:
you've got a mix of different whitespace styles, i do not now your style guidelines but,  for a easyer reading i would recomend to chose one of them and stay with it.
(whitespaces represented as '-')
1. {-"content"-}
2. {"content"-}
3. {"content"}

see pushflatpakscript/docker.d/worker.yml

stephenking314:
Closing in favor of: https://github.com/mozilla-releng/scriptworker-scripts/pull/1387

PR #1341, opened 2026-02-09
feat(treescript): Bug 2015581 Add github create branch action

hneiva:


treescript/src/treescript/github/versionmanip.py
```diff
@@ -57,7 +61,11 @@ def get_version(contents: str, version_cls: Type[BaseVersion]) -> BaseVersion:
     Returns:
         BaseVersion: The parsed version object.
     """
-    lines = [line for line in contents.splitlines() if line and not line.startswith("#")]
+    if not is_xcconfig:
+        lines = [line for line in contents.splitlines() if line and not line.startswith("#")]
+    else:
+        # version.xcconfig files should only have a single line with APP_VERSION = 111
+        lines = [line.lstrip("APP_VERSION = ") for line in contents.splitlines()]
```
Eijebong: `lstrip` here works kinda by miracle and the fact that we don't use any of these characters in the version name. `removeprefix` would be much more appropriate

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -43,6 +43,57 @@ async def __aenter__(self):
...
+        variables = {
+            "input": {
+                "repositoryId": repo_id,
+                "name": f"refs/heads/{branch_name}",
+                "oid": source_oid,
+            }
+        }
+
+        verb = "Would create" if dry_run else "Creating"
+        log.debug(f"{verb} {branch_name} on repo {repo}[{repo_id}] at {source_branch}@{source_oid}")
```
Eijebong:
```suggestion
        log.debug(f"{verb} {branch_name} on repo {self.repo}[{repo_id}] at {source_branch}@{source_oid}")
```

Otherwise you're logging the full response from graphql?

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -43,6 +43,57 @@ async def __aenter__(self):
     async def __aexit__(self, *excinfo):
         await self.close()
 
+    async def create_branch(self, branch_name: str, from_branch: Optional[str] = None, dry_run: bool = False) -> None:
+        """Create a new branch in the repository.
+
+        Args:
+            branch (str): The name of the new branch to create.
```
Eijebong:
```suggestion
            branch_name (str): The name of the new branch to create.
```

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -43,6 +43,57 @@ async def __aenter__(self):
...
+            from_branch (str): The branch to create the new branch from. Uses the
+                repository's default branch if unspecified (optional).
```
Eijebong, line 52: Missing docs on `dry_run`, not that it really needs any but that's inconsistent

treescript/src/treescript/util/task.py
```diff
@@ -154,6 +154,26 @@ def get_version_bump_info(task):
     return version_info
 
 
+# get_version_bump_info {{{1
```
Eijebong: 👀 wrong copy/paste

treescript/src/treescript/github/branch.py
```diff
@@ -0,0 +1,35 @@
...
+"""Treescript branch methods."""
+
+from typing import Dict
+
+from scriptworker_client.github_client import GithubClient
+
+from treescript.util.task import get_branch, get_create_branch_info, should_push
+
+
+def get_branch_name(task: Dict) -> str:
```
Eijebong, line 12:
```suggestion
def get_branch_name(task: Dict) -> Optional[str]:
```

Don't think it can be `None` but then you should use `["branch_name"]`. We don't use anything to actually read those type annotations but if we ever do it'd be nice to not have to fix this

treescript/src/treescript/github/branch.py
```diff
@@ -0,0 +1,35 @@
...
+    """Get the source branch name from a task's create_branch_info.
```
Eijebong: This is the opposite of what this function does isn't it? It's the target branch name, not the source? IMO worth renaming the function to `get_target_branch_name` also to clear out the confusion with `get_branch`

treescript/src/treescript/github/branch.py
```diff
@@ -0,0 +1,35 @@
...
+
+    Args:
+        task (Dict): The task definition containing create_branch_info.
+
+    Returns:
+        str: The name of the branch to create from, or None if not specified.
```
Eijebong: Same here, this says the opposite of what it does

scriptworker_client/src/scriptworker_client/github_client.py
```diff
@@ -43,6 +43,57 @@ async def __aenter__(self):
...
+                  oid
+                }
+              }
+            }""")
+        )
+        str_info_query = info_query.substitute(owner=self.owner, repo=self.repo, branch=source_branch)
+        repo = (await self._client.execute(str_info_query))["repository"]
+
+        if repo.get("object") is None:
+            raise UnknownBranchError(f"branch '{source_branch}' not found in repo!")
```
Eijebong, line 72: I wonder if we should also check if the branch already exists and bail early? I'm seeing the "branch creation passed but version bump got 500s" already...

Although there's an argument for saying that if the branch already exists it could've been created by mistake by a human and shouldn't be used and thus failing here is the right thing to do...

Eijebong (changes requested):
A bunch of minor things. Also I would've preferred if this had been split in 2 commits (I'm not even sure submitting the whole thing was intentional), the version bump in xcconfig and the branch creation since they're completely unrelated in code. Neither the commit nor the bug even mentions that version bump semantic change.

treescript/src/treescript/github/branch.py
```diff
@@ -0,0 +1,41 @@
...
+        task (Dict): The task definition containing create_branch_info.
+
+    Returns:
+        str: The name of the target branch.
+
+    Raises:
+        TreeScriptError: If branch_name is not specified in the task.
+    """
+    create_branch_info = get_create_branch_info(task)
+    if "branch_name" not in create_branch_info:
```
Eijebong, line 25: Nit: I don't think this can happen? I'd expect schema validation to fail it early

```
"create_branch_info": {
    "type": "object",
    "properties": {
        "branch_name": {
            "type": "string"
        }
    },
    "required": [
        "branch_name"
    ]
},
```

Eijebong (approved):

treescript/src/treescript/github/branch.py
hneiva: `create_branch_info` can be None

Eijebong: But then we have another problem?

```
>>> "branch_name" not in None
Traceback (most recent call last):
  File "<python-input-0>", line 1, in <module>
    "branch_name" not in None
TypeError: argument of type 'NoneType' is not a container or iterable
```

Eijebong: Also it can't be None at that point

```
create_branch_info = task.get("payload", {}).get("create_branch_info")
if not create_branch_info:
    raise TaskVerificationError("Requested create branch but no create_branch_info in payload")
```

PR #1347, opened 2026-02-10
feat(treescript): Bug 2015875 Support ios xcconfig version files

hneiva:


treescript/src/treescript/github/versionmanip.py
```diff
@@ -16,7 +16,11 @@
 log = logging.getLogger(__name__)
 
 
-ALLOWED_BUMP_FILES = ("version.txt",)
+ALLOWED_BUMP_FILES = (
```
Eijebong, line 19: I think for now this is fine but if we want to add more file types here it might be worth transforming this into a dict like this:

```py
BUMP_FILES_METHODS = {
    "version.txt": bump_version_txt,
    "...xcconfig": bump_xcconfig,
}
```

which would avoid having to pass extra args to `get_version` and clarify the path to version bumps a bit.

treescript/src/treescript/github/versionmanip.py
```diff
@@ -57,7 +61,11 @@ def get_version(contents: str, version_cls: Type[BaseVersion]) -> BaseVersion:
     Returns:
         BaseVersion: The parsed version object.
     """
-    lines = [line for line in contents.splitlines() if line and not line.startswith("#")]
+    if not is_xcconfig:
+        lines = [line for line in contents.splitlines() if line and not line.startswith("#")]
+    else:
+        # version.xcconfig files should only have a single line with APP_VERSION = 111
+        lines = [line.removeprefix("APP_VERSION = ") for line in contents.splitlines()]
```
Eijebong, line 68: Bit scared about not filtering empty lines here but I guess as long as we're the only ones modifying the file it's probably going to be fine

Eijebong (approved):

PR #1360, opened 2026-02-24
Handle more ok and failure states in pushmsix submissions

Eijebong:
According to microsoft's [StoreBroker] (which matches microsoft's [documentation] on the existing status lists), a happy path looks like this for an application:

`PendingCommit` -> `CommitStarted` -> `PreProcessing` -> `Certification` -> `Release` -> `PendingPublication` -> `Publishing` -> `Published`.

`Commit`, `PreProcessing`, `Publish`, `Certification` and `Release` all have a corresponding `Failed` state.

This handles every state after `PreProcessing` as being terminal and OK, and every failure state as being terminal and obviously NOK.

Hopefully this reduces the amount of silent failed uploads we're seeing to the MSIX store as well as the amount of false failure.

Data shows that half the failures since the start of the year were because the submission ended up in `Certification` without passing by `PreProcessing` (well, it passed by it between poll attempts). And a third of the remaining ones was because of lack of state cleanup between runs, where the state got messed up because of the previous errors. The remaining errors are actual timeouts and we should maybe consider increasing the timeout if it happens again (at least now we should be able to detect other failures and rule them out).

[StoreBroker]: https://github.com/microsoft/StoreBroker/blob/47d06c77d80b6a47d1b66a78d63d1609934d7104/Documentation/USAGE.md#status-progression
[documentation]: https://learn.microsoft.com/en-us/windows/uwp/monetize/get-status-for-an-app-submission

### DIFF
--- a/pushmsixscript/src/pushmsixscript/microsoft_store.py
+++ b/pushmsixscript/src/pushmsixscript/microsoft_store.py
@@ -267,13 +267,19 @@ def _get_submission_status(config, channel, session, submission_id, headers):
     return response.json()
 
 
+FAILURE_STATUSES = ("CommitFailed", "PreProcessingFailed", "CertificationFailed", "ReleaseFailed", "PublishFailed", "Canceled")
+# Statuses from the microsoft store that indicate the submission has passed the
+# commit phase and is out of our control
+OK_STATUSES = ("PreProcessing", "Certification", "Release", "PendingPublication", "Publishing", "Published")
+TERMINAL_STATUSES = FAILURE_STATUSES + OK_STATUSES
+
+
 def _wait_for_commit_completion(config, channel, session, submission_id, headers):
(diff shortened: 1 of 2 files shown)


pushmsixscript/src/pushmsixscript/microsoft_store.py
```diff
@@ -267,13 +267,19 @@ def _get_submission_status(config, channel, session, submission_id, headers):
     return response.json()
 
 
+FAILURE_STATUSES = ("CommitFailed", "PreProcessingFailed", "CertificationFailed", "ReleaseFailed", "PublishFailed", "Canceled")
+# Statuses from the microsoft stores that indicate the submission has passed the
```
jcristau:
```suggestion
# Statuses from the microsoft store that indicate the submission has passed the
```

jcristau (approved):

PR #1378, opened 2026-03-03
signingscript: verify gpg signatures

jcristau:
Verify the signature we get back from autograph against the public key.


signingscript/src/signingscript/script.py
```diff
@@ -69,6 +70,7 @@ async def async_main(context):
                 copy_to_dir(os.path.join(work_dir, source), context.config["artifact_dir"], target=source)
             if {"autograph_gpg", "gcp_prod_autograph_gpg", "stage_autograph_gpg"}.intersection(set(path_dict["formats"])):
                 copy_to_dir(context.config["gpg_pubkey"], context.config["artifact_dir"], target="public/build/KEY")
+                await set_up_gpg_keyring(context)
```
Eijebong: Isn't this too late? `output_files = await sign(` on line 67 calls:

```
signing_func = _get_signing_function_from_format(fmt)
output = await signing_func(context, output, fmt, **kwargs)
```

And for gpg, `sign_gpg_with_autograph` which then calls `verify_gpg`. But you're setting the keyring after signing occured so it'll always fail?

signingscript/src/signingscript/script.py
```diff
@@ -82,6 +84,17 @@ async def async_main(context):
     log.info("Done!")
 
 
+async def set_up_gpg_keyring(context):
+    with open(context.config["gpg_pubkey"], "rb") as pubkey, open(os.path.join(context.config["work_dir"], "trustedkeys.gpg"), "wb") as keyring:
+        p = await asyncio.create_subprocess_exec("gpg", "--dearmor", stdin=pubkey, stdout=keyring)
+        try:
+            ret = await asyncio.wait_for(p.wait(), timeout=2)
+        except TimeoutError:
+            ret = -1
```
Eijebong: Might want to also kill `p` instead of leaking it? Although I guess if should die on its own when the task fails? 🤔

signingscript/src/signingscript/script.py
```diff
@@ -82,6 +84,17 @@ async def async_main(context):
...
+        if ret != 0:
+            raise Exception("Could not make gpg trusted keyring")
```
Eijebong: Any reason to not raise a `SigningScriptError` like the other exceptions around?

signingscript/src/signingscript/script.py
jcristau: Yes, absolutely.  Thanks.

signingscript/src/signingscript/script.py
jcristau: For some reason `async_main` seems to raise bare `Exception`s, and I followed that.  I'll switch it though.

Eijebong (approved):

PR #1382, opened 2026-03-05
signingscript: unpin winsign

jcristau:
I don't think we need to pin a particular version in pyproject.toml, we can let the regular uv.lock updates handle it along with other dependencies.


Eijebong (approved):
Agreed

PR #1383, opened 2026-03-05
Update docker images from debian 12 to debian 13

jcristau:
Bump python 3.11-based docker images from 3.11.9 to 3.11.15, which also updates the underlying OS.

This required some fixes to msix-packaging (for compatibility with the newer glibc/libstdc++/ICU) and winsign (for compatibility with the newer osslsigncode).

### DIFF
--- a/taskcluster/kinds/docker-image/kind.yml
+++ b/taskcluster/kinds/docker-image/kind.yml
@@ -11,8 +11,8 @@ transforms:
     - taskgraph.transforms.task:transforms
 
 meta:
-    - &py311 "3.11.9"
-    - &py314 "3.14.3"
+    - &py311 "3.11.15-trixie"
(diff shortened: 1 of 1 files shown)


taskcluster/kinds/docker-image/kind.yml
```diff
@@ -11,7 +11,7 @@ transforms:
     - taskgraph.transforms.task:transforms
 
 meta:
-    - &py311 "3.11.9"
+    - &py311 "3.11.15"
```
Eijebong: I'm not sure how I feel about patch python versions silently changing the underlying debian version... It might be worth pinning that more precisely?

jcristau: Fair; updated.

Eijebong (approved):

PR #1386, opened 2026-03-11
pushflatpakscript: make scopes include app id in addition to channel

jcristau:
- refactor token_locations config to map channel+app_id to a token
- expect `{scope_prefix}:{channel}:{app_id}` instead of `{scope_prefix}:{channel}`, but fall back to the pre-existing/default app id for backwards compat


jcristau:
Superseded by #1387, closing.

PR #1395, opened 2026-03-17
Slim down docker images

jcristau:
Switch to the "slim" images variants, and use `--no-install-recommends`
when installing packages.

Install extra packages we need:
- the test images all need git, to clone scriptworker-scripts itself
- the pushflatpakscript and pushapkscript Dockerfiles call curl
- the githubscript tests expect the mimetypes module to map apk to
  application/vnd.android.package-archive, which requires
  the media-types package to populate /etc/mime.types
- pushflatpakscript needs a compiler and a couple of libraries to install
  flat-manager-client's dependencies
- signingscript uses wget to download rcodesign, and needs git, make and
  libicu-dev for the msix-packaging build

Also make build_msix_packaging.sh not ignore errors.

### DIFF
--- a/taskcluster/kinds/docker-image/kind.yml
+++ b/taskcluster/kinds/docker-image/kind.yml
@@ -11,8 +11,8 @@ transforms:
     - taskgraph.transforms.task:transforms
 
 meta:
-    - &py311 "3.11.15-trixie"
-    - &py314 "3.14.3-trixie"
+    - &py311 "3.11.15-slim-trixie"
+    - &py314 "3.14.3-slim-trixie"
     - &uv_version "0.7.15"
 
 tasks:
@@ -60,10 +60,7 @@ tasks:
         args:
             SCRIPT_NAME: bouncerscript
     githubscript:
-        definition: script
         parent: base
-        args:
-            SCRIPT_NAME: githubscript
     landoscript:
         definition: script
         parent: base
@@ -91,6 +88,13 @@ tasks:
         parent: base
 
     # Testing images
+    githubscript-test-py311:
+        definition: base-test
+        args:
+            PYTHON_VERSION: *py311
+            # tests expect mimetypes.guess_type to map *.apk to application/vnd.android.package-archive
+            APT_PACKAGES: media-types
+
     pushapkscript-test-py311:
         definition: base-test
         args:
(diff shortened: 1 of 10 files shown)


taskcluster/kinds/docker-image/kind.yml
```diff
@@ -91,6 +91,14 @@ tasks:
         parent: base
 
     # Testing images
+    githubscript-test-py311:
+        definition: base-test
+        args:
+            PYTHON_VERSION: *py311
+            # tests expect mimetypes.guess_type to map *.apk to application/vnd.android.package-archive
+            # XXX should we also install that on the prod githubscript image?
```
Eijebong: I feel like we should answer this before merging?

jcristau: I'm not sure.  The worst thing that can happen is an artifact is attached to a github release without a content-type, which doesn't feel like all that terrible.

jcristau: Added now, slightly reduces the risk of bustage.

Eijebong: I think that it's better because:

```

def upload_asset(self, content_type, name, asset, label=None):
    headers = {"Content-Type": content_type}
```

and I'm not sure how github is going to behave when we send `None` as our content type

Eijebong (approved):

jcristau:
image.tar.zst sizes before/after, and % win (combined with #1399 for pushapkscript):
```
addonscript	339398673	62714936	81
githubscript	341787711	80204231	76
pushapkscript	713589311	253397336	64
pushflatpakscript	363139457	176915742	51
shipitscript	339394128	62728645	81
pushmsixscript	342153909	65477968	80
bouncerscript	339578747	62920007	81
balrogscript	342060862	65447223	80
treescript	381083230	121396342	68
bitrisescript	341778231	65121597	80
landoscript	341970735	65293796	80
beetmoverscript	360831328	84196660	76
signingscript	1269177085	967182602	23
```

Source:
```
taskcluster group list -c L7eDPPHFTdGdv9Rc1i598Q | \
  sed -n '/docker-image-.*script / { s/docker-image-//;s/ completed//; p }' | \
  while read taskid script; do \
    before=$(curl -Is https://firefoxci.taskcluster-artifacts.net/$taskid/0/public/image.tar.zst | awk '/^content-length:/ {print int($2)}'); \
    after=$(curl -ILs https://firefox-ci-tc.services.mozilla.com/api/index/v1/task/scriptworker.cache.level-3.docker-images.v2.
$script.latest/artifacts/public%2Fimage.tar.zst | awk '/^content-length:/ { size =$2 } END { print int(size) }' ); \
    printf '%s\t%d\t%d\t%d\n' $script $before $after $(( (before - after) * 100 / before )) ; \
  done
```

PR #1399, opened 2026-03-18
pushapkscript: install default-jre-headless instead of default-jdk

jcristau:
AIUI we need to run keytool and bundletool, neither of which requires the full jdk.


bhearsum (approved):

jcristau:
Before: image.tar.zst is 680M
After: image.tar.zst is 567M

PR #1403, opened 2026-03-19
fix: use full python install on skopeo image

bhearsum:
This is needed to run `run-task`. Presumably regressed by https://github.com/mozilla-releng/scriptworker-scripts/pull/1395.

See https://firefox-ci-tc.services.mozilla.com/tasks/WszCJA9bSQSt46xX7PtBkQ/runs/0 and https://firefox-ci-tc.services.mozilla.com/tasks/RQDNW2TUT221noIPdXKC_A for push tasks with and without these change.

### DIFF
--- a/taskcluster/docker/skopeo/Dockerfile
+++ b/taskcluster/docker/skopeo/Dockerfile
@@ -34,7 +34,7 @@ FROM debian:trixie@sha256:55a15a112b42be10bfc8092fcc40b6748dc236f7ef46a358d9392b
 ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update -qq \
     && apt-get dist-upgrade -y \
-    && apt-get install --no-install-recommends -y jq zstd python3-minimal curl \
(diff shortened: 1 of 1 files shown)


taskcluster/docker/skopeo/Dockerfile
```diff
@@ -34,7 +34,9 @@ FROM debian:trixie@sha256:55a15a112b42be10bfc8092fcc40b6748dc236f7ef46a358d9392b
 ENV DEBIAN_FRONTEND=noninteractive
 RUN apt-get update -qq \
     && apt-get dist-upgrade -y \
-    && apt-get install --no-install-recommends -y jq zstd python3-minimal curl \
+    # python3 used instead of python3-minimal because we need shutil from the
+    # stdlib for `run-task`
```
jcristau: nit: not sure we need to keep the comment

jcristau (approved):
Thanks for the fix!

PR #1405, opened 2026-03-23
Ensure /etc/mime.types is present in all images

Eijebong:
02a0296b4889f1bbd810363685077b3ce1389e1c slimmed down docker images and installed `mime-types` in the githubscript image. While testing beetmoverscript I found out that it too needs mime types or it fails with `ScriptWorkerTaskException: Unable to discover valid mime-type for path (....deb), mimetypes.guess_type() returned None`.

Instead of making beetmoverscript another special child that requires its package and instead of registering those types, just bite the bullet and add `media-types` to all images. The package uses 26kB, I think we'll survive the size difference. This allows us to make githubscript a normal child again also and will prevent other unforeseen issues...

(I did try to register the .rpm/.deb types but the base test image isn't slim and has `media-types` installed which means tests wouldn't make much sense as we couldn't verify that registration actually worked and was correct for types it already knows)


bhearsum (approved):
> The package uses 26kB, I think we'll survive the size difference.

agreed

PR #1406, opened 2026-03-24
pushflatpak: REPO_TOKEN_TB_* variables are optional

jcristau:
We don't have tokens for the new apps yet, don't crash on startup.


bhearsum (commented):
would it preferable to back this out in the meantime instead of marking them as optional? (then we wouldn't have to remember to mark them as required again later...)

bhearsum (approved):

Eijebong (approved):

jcristau:
> would it preferable to back this out in the meantime instead of marking them as optional? (then we wouldn't have to remember to mark them as required again later...)

I'd rather keep them, if only to reserve/document the variable names since we have to pass this stuff from sops → cloudops-infra → here and it's all error prone :)

PR #1412, opened 2026-03-26
Bug 2006684 - add builddecisionscript

jcristau:
Port the build-decision code to run as a scriptworker task.

The following changes have been made:
- use `taskgraph.util.keyed_by` instead of a local copy
- input is taken from fields in the task payload instead of either command-line arguments or the environment
- we don't have a taskcluster proxy; instead we use temporary credentials obtained/maintained by scriptworker to create tasks / trigger hooks / fetch secrets
- the CLI is modified to match other scriptworkers

Unlike other scriptworkers, we skip chain of trust verification.


jcristau:
This depends on https://github.com/mozilla-releng/scriptworker/pull/798 to get temporary tc credentials with the task's scopes.

bhearsum (approved):
For the builddecision code itself, I mostly reviewed by comparing fxci-config's version to this, which looked sane.

> Unlike other scriptworkers, we skip chain of trust verification.

This is presumably because there's nothing upstream to verify?

(This also makes me wonder: should CoT be verifying any aspect of build decision? I've always found it a bit weird that decision tasks and build-decision tasks don't have a formal link...)

jcristau:
> For the builddecision code itself, I mostly reviewed by comparing fxci-config's version to this, which looked sane.
> 
> > Unlike other scriptworkers, we skip chain of trust verification.
> 
> This is presumably because there's nothing upstream to verify?
> 
Indeed.

> (This also makes me wonder: should CoT be verifying any aspect of build decision? I've always found it a bit weird that decision tasks and build-decision tasks don't have a formal link...)

I don't think so?  We rebuild the decision task's definition based on the repo contents, and there's not necessarily a build-decision task (particularly for github), so we can't really go past that.

hneiva:
@jcristau should we merge this and start testing?

jcristau:
@hneiva not yet, I've been testing in dev, which doesn't need this merged

Issue #912, opened 2024-02-09
addonscript: don't retry on 4xx errors


jcristau:
While it makes sense to retry http requests on timeout or 5xx, 4xx is unlikely to be temporary, so retrying just delays the inevitable.

PR #1413, opened 2026-03-26
Bug 2025695 - landoscript: extend version_bump to handle JSON manifests

JohanLorenzo:
[Bug 2025695](https://bugzilla.mozilla.org/show_bug.cgi?id=2025695)

## Summary

- Extends `version_bump` to handle JSON manifests: adds `browser/extensions/newtab/manifest.json` and `browser/extensions/webcompat/manifest.json` to `ALLOWED_BUMP_FILES`, uses `endswith("/manifest.json")` to select the JSON-specific version reader, and uses `str.replace` for the actual replacement (same as all other files)
- Adds `parse_manifest_version()` as an independently unit-testable pure function
- Makes the action idempotent: warns and skips (instead of raising) when the replacement produces no change

## Merge order

1. mozilla-releng/fxci-config [#912](https://github.com/mozilla-releng/fxci-config/pull/912) — merged ✓
2. mozilla-releng/fxci-config [#933](https://github.com/mozilla-releng/fxci-config/pull/933) — merged ✓ (scope name update)
3. mozilla-services/cloudops-infra [#6835](https://github.com/mozilla-services/cloudops-infra/pull/6835) (dev) — merged ✓
4. mozilla-services/cloudops-infra [#6848](https://github.com/mozilla-services/cloudops-infra/pull/6848) (chart secret) — merged ✓
5. k8s-sops: dev worker secrets — done ✓
6. mozilla-releng/k8s-autoscale [#264](https://github.com/mozilla-releng/k8s-autoscale/pull/264) (dev) — merged ✓
7. **This PR** — merged ✓
8. mozilla-services/cloudops-infra [#6838](https://github.com/mozilla-services/cloudops-infra/pull/6838) (prod) — merged ✓
9. k8s-sops: prod worker secrets — manual step
10. mozilla-releng/k8s-autoscale [#266](https://github.com/mozilla-releng/k8s-autoscale/pull/266) (prod) — merged ✓
11. mozilla-releng/fxci-config [#979](https://github.com/mozilla-releng/fxci-config/pull/979) — fix production lando action scope — merged ✓
12. mozilla-extensions/xpi-manifest [#271](https://github.com/mozilla-extensions/xpi-manifest/pull/271) — merged ✓

## Verification

- Unit tests: 104 pass (`uv run pytest` in `landoscript/`)
- Staging end-to-end test via local landoscript (CoT disabled): fetched `manifest.json` from `staging-firefox`, bumped `150.1.0 → 150.2.0`, submitted to staging Lando, got `LANDED` status
- Staging XPI release via Ship-It staging: version-bump task [BRV3TeLXT8S1wONrxgN7_g](https://firefox-ci-tc.services.mozilla.com/tasks/BRV3TeLXT8S1wONrxgN7_g) completed successfully on `xpi-1-lando-dev` worker, landing [this commit](https://github.com/mozilla-releng/staging-firefox/commit/cf5f4fda76c7a226dda97f61bd973d42ba63cfff) on staging-firefox:
  ```
  fetching bump files from github
  browser/extensions/newtab/manifest.json: successfully bumped!
  -  "version": "151.3.0",
  +  "version": "151.4.0",
  Not a dry run...submitting lando actions
  submitting POST request to https://stage.lando.nonprod.webservices.mozgcp.net/api/repo/staging-firefox-main
  success! got 200 response with 'LANDED' status
  ```

### DIFF
--- a/landoscript/src/landoscript/actions/version_bump.py
+++ b/landoscript/src/landoscript/actions/version_bump.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import os.path
 import typing
@@ -17,10 +18,11 @@ from landoscript.util.version import find_what_version_parser_to_use
 
 log = logging.getLogger(__name__)
 
-# A list of files that this action is allowed to operate on.
 ALLOWED_BUMP_FILES = (
     "browser/config/version.txt",
     "browser/config/version_display.txt",
+    "browser/extensions/newtab/manifest.json",
+    "browser/extensions/webcompat/manifest.json",
     "config/milestone.txt",
     "mobile/android/version.txt",
     "mail/config/version.txt",
@@ -32,6 +34,12 @@ ALLOWED_BUMP_FILES = (
 class VersionBumpInfo:
     next_version: str
     files: list[str]
+
+
+def parse_manifest_version(orig_contents: str) -> BaseVersion:
+    """Extract and parse the version field from a JSON manifest."""
+    manifest = json.loads(orig_contents)
+    return BaseVersion.parse(manifest["version"])
 
 
 async def run(
@@ -119,6 +127,11 @@ async def run(
 
 
 def get_cur_and_next_version(filename, orig_contents, next_version, munge_next_version):
+    if filename.endswith(".json"):
+        cur = parse_manifest_version(orig_contents)
+        next_ = BaseVersion.parse(next_version)
+        return cur, next_
+
     VersionClass: BaseVersion = find_what_version_parser_to_use(filename)
     lines = [line for line in orig_contents.splitlines() if line and not line.startswith("#")]
     cur = VersionClass.parse(lines[-1])
--- a/landoscript/src/landoscript/actions/version_bump.py
+++ b/landoscript/src/landoscript/actions/version_bump.py
@@ -95,7 +95,8 @@ async def run(
 
             modified = orig.replace(str(cur), str(next_))
             if orig == modified:
-                raise LandoscriptError("file not modified, this should be impossible")
+                log.warning(f"{file}: version replacement produced no change, skipping")
+                continue
 
             log.info(f"{file}: successfully bumped! new contents are:")
             log_file_contents(modified)
--- a/landoscript/src/landoscript/actions/version_bump.py
+++ b/landoscript/src/landoscript/actions/version_bump.py
@@ -1,3 +1,4 @@
+import json
 import logging
 import os.path
 import typing
@@ -17,10 +18,11 @@ from landoscript.util.version import find_what_version_parser_to_use
 
 log = logging.getLogger(__name__)
 
-# A list of files that this action is allowed to operate on.
 ALLOWED_BUMP_FILES = (
     "browser/config/version.txt",
     "browser/config/version_display.txt",
+    "browser/extensions/newtab/manifest.json",
+    "browser/extensions/webcompat/manifest.json",
     "config/milestone.txt",
     "mobile/android/version.txt",
     "mail/config/version.txt",
@@ -32,6 +34,12 @@ ALLOWED_BUMP_FILES = (
 class VersionBumpInfo:
     next_version: str
     files: list[str]
+
+
+def parse_manifest_version(orig_contents: str) -> BaseVersion:
+    """Extract and parse the version field from a JSON manifest."""
+    manifest = json.loads(orig_contents)
+    return BaseVersion.parse(manifest["version"])
 
 
 async def run(
@@ -119,6 +127,11 @@ async def run(
 
 
 def get_cur_and_next_version(filename, orig_contents, next_version, munge_next_version):
+    if filename.endswith("/manifest.json"):
+        cur = parse_manifest_version(orig_contents)
+        next_ = BaseVersion.parse(next_version)
+        return cur, next_
+
     VersionClass: BaseVersion = find_what_version_parser_to_use(filename)
     lines = [line for line in orig_contents.splitlines() if line and not line.startswith("#")]
     cur = VersionClass.parse(lines[-1])
--- a/landoscript/tests/test_version_bump.py
+++ b/landoscript/tests/test_version_bump.py
@@ -6,6 +6,7 @@ from pytest_scriptworker_client import get_files_payload
 from landoscript.errors import LandoscriptError
 from landoscript.script import async_main
 from landoscript.actions.version_bump import ALLOWED_BUMP_FILES
+import json as _json
 from landoscript.util.version import _VERSION_CLASS_PER_BEGINNING_OF_PATH
 
 from .conftest import (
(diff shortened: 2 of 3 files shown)


jcristau:
I don't love that this is an entirely new ad-hoc action.  Couldn't we re-use the existing version_bump and tag actions?

JohanLorenzo:
> I don't love that this is an entirely new ad-hoc action. Couldn't we re-use the existing version_bump and tag actions?

Discussed over Element. We agreed on reusing `version_bump` now and, in a follow-up, safeguard files via scopes. This way, we avoid to scope-creep this set of PRs.

landoscript/src/landoscript/actions/version_bump.py
```diff
@@ -17,23 +18,51 @@
 
 log = logging.getLogger(__name__)
 
-# A list of files that this action is allowed to operate on.
+# Files that use plain-text version strings (one version per file).
```
bhearsum: This comment is not true, I think? (Maybe it was in an earlier iteration of the code?

landoscript/src/landoscript/actions/version_bump.py
```diff
@@ -17,23 +18,51 @@
...
     files: list[str]
 
 
+def parse_manifest_version(orig_contents: str) -> BaseVersion:
+    """Extract and parse the version field from a JSON manifest."""
+    manifest = json.loads(orig_contents)
+    return BaseVersion.parse(manifest["version"])
+
+
+def apply_manifest_version_bump(orig: str, next_: BaseVersion) -> str:
```
bhearsum: Why isn't the current replacement method good enough for this file? I'm not seeing anything obvious in https://searchfox.org/firefox-main/source/browser/extensions/newtab/manifest.json that suggests it wouldn't work.

(I understand the motivation to have a more refined method here; I'm not sure it's worth the complication in this case though.)

bhearsum (commented):
I don't necessarily disagree that it's good to have this in the same file, but I can't help but notice that at least the current implementation has two notable departures from the main code path (a special way to create "next" versions and a special way to do the replacement). This is likely to lead to maintenance and comprehensibility problems later, especially if/when we add more cases like this.

Bubbling these sorts of branches up as high as possible would be one way to improve the situation. Rearchitecturing/refactoring to generalize the idea of filename-specific bumping methods might be another, eg: perhaps associating filenames with bump version generation and bump methods.

landoscript/src/landoscript/actions/version_bump.py
```diff
@@ -17,23 +18,51 @@
 
 log = logging.getLogger(__name__)
 
-# A list of files that this action is allowed to operate on.
+# Files that use plain-text version strings (one version per file).
 ALLOWED_BUMP_FILES = (
     "browser/config/version.txt",
     "browser/config/version_display.txt",
+    "browser/extensions/newtab/manifest.json",
```
jcristau, line 24: Per the bug we also want to add webcompat.

landoscript/src/landoscript/actions/version_bump.py
```diff
@@ -17,23 +18,51 @@
...
     "config/milestone.txt",
     "mobile/android/version.txt",
     "mail/config/version.txt",
     "mail/config/version_display.txt",
 )
 
+# Subset of ALLOWED_BUMP_FILES that are JSON manifests requiring
+# manifest-specific version extraction and replacement logic.
+# Only files explicitly listed here will trigger that code path.
+JSON_MANIFEST_BUMP_FILES = frozenset(
```
jcristau: Isn't that anything that looks like *.json?

jcristau (commented):
Agree with Ben on the extra code seeming unnecessary.

landoscript/src/landoscript/actions/version_bump.py
JohanLorenzo: Fixed by removing the inaccurate comment.

landoscript/src/landoscript/actions/version_bump.py
JohanLorenzo: Okay, I switched to `str.replace` for all files.

landoscript/src/landoscript/actions/version_bump.py
JohanLorenzo: Added.

landoscript/src/landoscript/actions/version_bump.py
JohanLorenzo: You right, done.

JohanLorenzo:
Thanks for the reviews! How does this simpler approach look?

landoscript/src/landoscript/actions/version_bump.py
```diff
@@ -34,6 +36,12 @@ class VersionBumpInfo:
     files: list[str]
 
 
+def parse_manifest_version(orig_contents: str) -> BaseVersion:
```
bhearsum, line 39: If we're not using a parser to write there's no good reason to use it to read either IMO. If we care about the risk of updating the wrong substring in the file the previously reviewed version of this is preferable. If we're using simple string replacement, this is unnecessary.

bhearsum (changes requested):

landoscript/src/landoscript/actions/version_bump.py
JohanLorenzo: I'm sorry, I'm not sure I follow. `get_cur_and_next_version()` parses both the current and the next version no matter if it's a plain text or not. So the existing implementation is already parses the data to read but not to write. 

The extra layer here just extracts the content of `manifest["version"]` for this check to happen: 

https://github.com/mozilla-releng/scriptworker-scripts/blob/01e618e73177c339b137e58ed4039b45b0562220/landoscript/src/landoscript/actions/version_bump.py#L80-L86

I'm not sure what I should change here. Could you expand on it @bhearsum?

bhearsum: You're right, my bad. This is slightly different and more involved than what we do currently, but not categorically different.

landoscript/src/landoscript/actions/version_bump.py
```diff
@@ -119,6 +128,11 @@ def extract_path(diff_text):
 
 
 def get_cur_and_next_version(filename, orig_contents, next_version, munge_next_version):
+    if filename.endswith(".json"):
+        cur = parse_manifest_version(orig_contents)
```
bhearsum, line 132: I might suggest a more detailed check on what we're parsing here rather than just looking for ".json", but this is also fine for now.

bhearsum (approved):

landoscript/src/landoscript/actions/version_bump.py
JohanLorenzo: Good point. Done by looking for `/manifest.json` instead.

PR #1424, opened 2026-04-15
Bug 1950203 - fix(signingscript): correct openh264 plugin format name in notarization setup check

hneiva:
The format name `apple_notarize_openh264_plugin` didn't match the actual format key `apple_notarization_openh264_plugin` in the signing format map, causing `setup_apple_notarization_credentials` to never run for openh264 tasks and resulting in an AttributeError on `context.apple_credentials_path`.


bhearsum (approved):
I see in https://phabricator.services.mozilla.com/D293693 that this is correct.

PR #1430, opened 2026-04-24
fix(signingscript): Bug 2033267 - clear stale authenticode outfile before retrying winsign

hneiva:
When winsign.sign.sign_file fails transiently (e.g. DigiCert CRL hiccup during post-sign verification), osslsigncode has already written <orig_path>-new to disk. Subsequent retries then fail with "Overwriting an existing file is not supported.", masking the original error and turning transient failures into permanent task failures. _winsign_helper now unlinks the outfile before each sign_file invocation so retries start clean.

### DIFF
--- a/signingscript/src/signingscript/sign.py
+++ b/signingscript/src/signingscript/sign.py
@@ -11,6 +11,7 @@ import json
 import logging
 import lzma
 import os
+import pathlib
 import re
 import resource
 import shutil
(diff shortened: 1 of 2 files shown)


hneiva:
Example failure that this patch fixes:
https://firefox-ci-tc.services.mozilla.com/tasks/I4YMborpSZGKntgWqffcvQ/runs/0/logs/public/logs/live_backing.log#L127-211

ahal (approved):

PR #1438, opened 2026-05-06
chore: add support for temporary mozcloud production environment

bhearsum:
These workers will be used to verify that balrogscript is working correctly with the new MozCloud production environment on a project branch; it will never be used by Nightly et. al. This configuration will be removed once testing is complete.


jcristau (approved):
nit: looks like `AUTH0_AUDIENCE` could be set in the toplevel `case $ENV` block instead of repeated for each cot product.

PR #1443, opened 2026-05-08
chore: adjust renovate cron to run twice a month

ahal:
I plan to make the template default to monthly, but for things like scriptworker-scripts that deploy bi-weekly, I want it to be more frequent.

So overwrite the schedule to be twice a month as there isn't a good way to express bi-weekly in cron syntax. I think it's ok that they will now run on arbitrary days in relation to the deploys. This fully de-couples bumping dependencies from releases, so it's one less thing to do on release days.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,5 +2,6 @@
   "$schema": "https://docs.renovatebot.com/renovate-schema.json",
   "extends": [
     "github>mozilla-releng/reps:renovate-preset.json5"
-  ]
+  ],
+  "schedule": ["* 2 1,15 * *"]
 }


bhearsum (approved):
Do you know how this will interact with the overridden schedules for `github-actions` and `pre-commit` in the template? Is the intention to leave those monthly, or will they (ideally) follow the scheduled defined here?

(I think either is fine, as neither of those are terribly important updates to take, but I wanted to double check the intent here.)

ahal:
My intention was to leave them monthly as they only impact CI and not the product.. But tbh I'm not sure if it's going to work or not!

PR #1447, opened 2026-05-20
fix(landoscript): log HTTP error response body before raising (bug 2040934)

JohanLorenzo:
[Bug 2040934](https://bugzilla.mozilla.org/show_bug.cgi?id=2040934)

### DIFF
--- a/landoscript/src/landoscript/lando.py
+++ b/landoscript/src/landoscript/lando.py
@@ -57,17 +57,21 @@ async def submit(
     url = f"{lando_api}/repo/{lando_repo}"
     json = {"actions": actions}
 
-    async with timeout(30):
-        log.info(f"submitting POST request to {url}")
+    async def _post():
+        resp = await session.post(url, json=json, headers=_get_auth_headers(lando_token))
+        if resp.status >= 400:
+            body = await resp.text()
+            log.error(f"POST to {url} failed with {resp.status}: {body}")
+            if resp.status < 500:
+                raise LandoscriptError(f"POST to {url} failed with {resp.status}: {body}")
+        resp.raise_for_status()
+        return resp
 
+    log.info(f"submitting POST request to {url}")
+
+    async with timeout(30):
         submit_resp = await retry_async(
-            session.post,
-            args=(url,),
-            kwargs={
-                "json": json,
-                "raise_for_status": True,
-                "headers": _get_auth_headers(lando_token),
-            },
+            _post,
             attempts=10,
             retry_exceptions=ClientResponseError,
             sleeptime_callback=sleeptime_callback or calculate_sleep_time,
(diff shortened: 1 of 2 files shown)


landoscript/tests/test_script.py
```diff
@@ -365,6 +366,36 @@ async def test_ignore_closed_tree_properly_errors(aioresponses, github_installat
...
+        "version_bump_info": {
+            "files": ["browser/config/version.txt"],
+            "next_version": "135.0",
+        },
+    }
+    initial_values = {"browser/config/version.txt": "134.0"}
+    submit_uri, _, _, scopes = setup_test(aioresponses, github_installation_responses, context, payload, ["version_bump"])
+    setup_github_graphql_responses(aioresponses, get_files_payload(initial_values))
+
+    error_body = '{"details": "User foo@example.com is not allowed to use this API for repo repo_name."}'
```
JohanLorenzo, line 386: Message taken from https://github.com/mozilla-conduit/lando/blob/e85c497263805501a820a77fe09fc3ed55278a9f/src/lando/headless_api/api.py#L432-L437. I used this new test for a red/green TDD approach.

landoscript/src/landoscript/lando.py
```diff
@@ -57,17 +57,19 @@ async def submit(
...
     json = {"actions": actions}
 
-    async with timeout(30):
-        log.info(f"submitting POST request to {url}")
+    async def _post():
+        resp = await session.post(url, json=json, headers=_get_auth_headers(lando_token))
+        if resp.status >= 400:
+            body = await resp.text()
+            log.error(f"POST to {url} failed with {resp.status}: {body}")
+            resp.raise_for_status()
```
jcristau: nit: consider moving the raise_for_status call outside the conditional, just in case

landoscript/src/landoscript/lando.py
```diff
@@ -57,17 +57,19 @@ async def submit(
     url = f"{lando_api}/repo/{lando_repo}"
     json = {"actions": actions}
 
-    async with timeout(30):
-        log.info(f"submitting POST request to {url}")
+    async def _post():
+        resp = await session.post(url, json=json, headers=_get_auth_headers(lando_token))
+        if resp.status >= 400:
+            body = await resp.text()
+            log.error(f"POST to {url} failed with {resp.status}: {body}")
```
jcristau, line 64: AFAICT this means we'll log the same thing up to 10 times in the retry loop?
An improvement might be to not retry on 4xx (only 5xx and things like timeouts or network errors).

landoscript/src/landoscript/lando.py
JohanLorenzo: Good point. Done.

landoscript/src/landoscript/lando.py
JohanLorenzo: Agreed. Done.

jcristau (approved):

PR #1449, opened 2026-05-26
feat(shipitscript): add support for creating new nightly releases

bhearsum:
This is not quite ready for review, but goes hand in hand with https://github.com/mozilla-releng/shipit/pull/1972.

### DIFF
--- a/shipitscript/src/shipitscript/data/create_new_nightly_release_task_schema.json
+++ b/shipitscript/src/shipitscript/data/create_new_nightly_release_task_schema.json
@@ -1,0 +1,50 @@
+{
+    "title": "Taskcluster ShipIt create new nightly release task minimal schema",
+    "type": "object",
+    "properties": {
+        "dependencies": {
+            "type": "array",
+            "minItems": 1,
+            "uniqueItems": true,
+            "items": {
+                "type": "string"
+            }
+        },
+        "scopes": {
+            "type": "array",
+            "minItems": 2,
+            "uniqueItems": true,
+            "items": {
+                "type": "string"
+            }
+        },
+        "payload": {
+            "type": "object",
+            "properties": {
+                "product": {
+                  "type": "string"
+                },
+                "channel": {
+                  "type": "string"
+                },
+                "buildid": {
+                  "type": "string"
+                },
+                "version": {
+                  "type": "string"
+                },
+                "locales": {
+                  "type": "array",
+                  "minItems": 1,
+                  "uniqueItems": true,
+                  "items": {
+                      "type": "string"
+                  }
+                }
+            },
+            "required": ["product", "channel", "buildid", "version", "locales"],
+            "additionalProperties": false
+        }
+    },
+    "required": ["dependencies", "scopes", "payload"]
+}
--- a/shipitscript/src/shipitscript/script.py
+++ b/shipitscript/src/shipitscript/script.py
@@ -114,12 +114,42 @@ def update_product_channel_version_action(context):
         ship_actions.update_product_channel_version(shipit_config, product, channel, version)
 
 
+def create_new_nightly_release_action(context):
+    payload = context.task["payload"]
+    shipit_config = context.ship_it_instance_config
+    product = payload["product"]
+    channel = payload["channel"]
+    version = payload["version"]
+    buildid = payload["buildid"]
+    locales = payload["locales"]
+    nightly = ship_actions.get_nightly_metadata(shipit_config, product, channel, buildid)
+    if nightly:
+        if len(nightly) != 1:
+            raise TaskVerificationError("Found multiple nightlies for the same buildid?")
+
+        nightly = nightly[0]
+        if nightly["version"] != version:
+            log.error(f"Nightly already exists; version does not match: {nightly['version']}")
+            sys.exit(1)
+
+        submitted_locales = set(locales)
+        existing_locales = set(nightly["locales"])
+        if mismatch := submitted_locales.symmetric_difference(existing_locales):
+            log.error(f"Nightly already exists; missing/added locales: {mismatch}")
+            sys.exit(1)
+
+        log.info("Nightly already exists with matching data; doing nothing.")
+    else:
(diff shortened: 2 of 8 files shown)


bhearsum:
Tested over in https://firefox-ci-tc.services.mozilla.com/tasks/LXpe1oFERfSwNqDW9TQbMA

shipitscript/src/shipitscript/script.py
```diff
@@ -114,12 +114,32 @@ def update_product_channel_version_action(context):
...
+    product = payload["product"]
+    channel = payload["channel"]
+    version = payload["version"]
+    buildid = payload["buildid"]
+    locales = payload["locales"]
+    nightly = ship_actions.get_nightly_metadata(shipit_config, product, channel, buildid)
+    if nightly:
+        assert len(nightly) == 1, "Found multiple nightlies for the same buildid?"
+        nightly = nightly[0]
+        if nightly["version"] != version or nightly["locales"] != locales:
```
Eijebong: Bit scared about the unsorted `locales` comparison here. It might be a nothingburger but if either the task or shipit doesn't sort locales it'll end up failing and the error isn't great in that case. Could do something along the line of:

```py
requested_locales = sorted(locales)
existing_locales = sorted(nightly["locales"])
if nightly["version"] != version:
    log.error(f"Nightly already exists but version doesn't match. Requested: {version}, existing: {nightly['version']}")
    sys.exit(1)
if existing_locales != requested_locales:
    log.error(f"Nightly already exists but locales don't match. Requested: {requested_locales}, existing: {existing_locales}")
    sys.exit(1)

```

Although that might be overprotective, I don't know how often this will actually fail

shipitscript/tests/test_script.py
```diff
@@ -94,6 +94,49 @@ async def test_async_main(context, monkeypatch, task, raises):
         await script.async_main(context)
 
 
+@pytest.mark.parametrize(
+    "existing_nightly,expect_create,expect_exit",
+    (
+        # no existing nightly: create one
+        (None, True, False),
+        # existing nightly matches version + locales: still calls create
+        ([{"version": "150.0a1", "locales": ["en-US", "de"]}], False, False),
```
Eijebong, line 103: `# [...]: still calls create`, look inside, `expect_create=False`

shipitscript/src/shipitscript/script.py
```diff
@@ -114,12 +114,32 @@ def update_product_channel_version_action(context):
...
+    payload = context.task["payload"]
+    shipit_config = context.ship_it_instance_config
+    product = payload["product"]
+    channel = payload["channel"]
+    version = payload["version"]
+    buildid = payload["buildid"]
+    locales = payload["locales"]
+    nightly = ship_actions.get_nightly_metadata(shipit_config, product, channel, buildid)
+    if nightly:
+        assert len(nightly) == 1, "Found multiple nightlies for the same buildid?"
```
Eijebong: Should this raise a `TaskVerificationError`? 🤔
I doubt it'll ever happen but if we're running with `python -O` in prod, assertions are stripped so this wouldn't actually fail

shipitscript/src/shipitscript/script.py
```diff
@@ -114,12 +114,32 @@ def update_product_channel_version_action(context):
...
+        nightly = nightly[0]
+        if nightly["version"] != version or nightly["locales"] != locales:
+            log.error("Nightly already exists, but version and/or locales don't match!")
+            sys.exit(1)
```
Eijebong, line 139: Might be worth adding a log line after that `if` when nightlies did match. Something along the line of "nightly already there and matches the request, doing nothing".

shipitscript/tests/test_ship_actions.py
```diff
@@ -48,3 +48,53 @@ def test_mark_as_merged(monkeypatch, timeout, expected_timeout):
         taskcluster_client_id="some-id", taskcluster_access_token="some-token", api_root="http://some.ship-it.tld/api/root", timeout=expected_timeout
     )
     release_instance_mock.complete_merge_automation.assert_called_with(123, headers={"X-Forwarded-Proto": "https", "X-Forwarded-Port": "80"})
+
+
+@pytest.mark.parametrize("timeout, expected_timeout", ((1, 1), ("10", 10), (None, 60)))
+def test_get_nightly_metadata(monkeypatch, timeout, expected_timeout):
+    ReleaseClassMock = MagicMock()
+    release_instance_mock = MagicMock()
+    nightly_metadata = {"version": "150.0a1", "locales": ["en-US", "de"]}
```
Eijebong: This should be a list? `nightly_metadata = [{"version": "150.0a1", "locales": ["en-US", "de"]}]`

It works now because `get_nightly_metadata` doesn't actually parse anything except for the fact that it's JSON and passes the output through unchanged.

shipitscript/src/shipitscript/script.py
bhearsum: Yeah, this should be guarded against.

shipitscript/src/shipitscript/script.py
bhearsum: TIL about python -O and asserts. Yes, let's use an exception instead.

shipitscript/tests/test_ship_actions.py
bhearsum: Yeah, that's right. (I'm going to fix this, but also remove the useless assert...)

shipitscript/tests/test_script.py
bhearsum: sigh, yes

shipitscript/src/shipitscript/data/create_new_nightly_release_task_schema.json
```diff
@@ -0,0 +1,51 @@
...
+                  }
+                }
+            },
+            "required": ["product", "channel", "buildid", "version", "locales"],
+            "additionalProperties": false
+        }
+    },
+    "required": ["dependencies", "scopes", "payload"]
+}
+
```
Eijebong: Nit: ✂️

```suggestion
```

Eijebong (approved):

PR #1458, opened 2026-06-01
Bug 2037124 - Add screenshots for Nightly flatpak

saschanaz:


pushflatpakscript/media/screenshots/nightly/image2.png
Eijebong, line 1: I know this is an updated version of the same wikipedia page as release but I personally think we should change that screenshot to something else... That flag has a very, let's say unfortunate, color palette, shape and symbolism combined...

saschanaz: I'm open to change it to something else. https://en.wikipedia.org/wiki/Firefox maybe?

saschanaz: Done, but then I realized that I can't do this with local nightly as it has a different logo... I'll figure that out.

Edit: that done too

saschanaz:
image3 is intentionally not about:welcome because Nightly uses the firstrun page without the usual onboarding. https://searchfox.org/firefox-main/rev/e28b34ab33dbf49364999070168cbb7e11e8e5bd/browser/branding/nightly/pref/firefox-branding.js#8

jcristau (approved):

PR #1459, opened 2026-06-01
fix(bug 2043228): allow renovate PRs to be created at any hour

bhearsum:
From [the documentation](https://docs.renovatebot.com/key-concepts/scheduling/#global-schedule-vs-specific-schedule)

> Renovate can only update a dependency if both of these conditions are true:
>
> * The Renovate program is running (on your hardware, or on Mend's hardware)
> * The schedule set in the Renovate config file(s) allows Renovate to look for updates for that dependency

The documentation additionally describes that Mend controls when the Renovate program is running.

With this in mind, I believe the recent changes to restrict Renovate to only running at 2am UTC have resulted in fewer PRs being created. Now that we've made the change to restrict to one day a month, having them opened at any hour shouldn't be a big deal.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,5 +3,5 @@
   "extends": [
     "github>mozilla-releng/reps:renovate-preset.json5"
   ],
-  "schedule": ["* 2 1,15 * *"]
+  "schedule": ["* * 1,15 * *"]
 }


Eijebong:
Doesn't this mean that we could get multiple PRs on the same day?

Edit: Meant to comment this on the reps one (https://github.com/mozilla-releng/reps/pull/92), got bamboozled by the same PR being opened in all repos that required it and me not checking. But discussing it here works too.

bhearsum:
> Doesn't this mean that we could get multiple PRs on the same day?
> 
> Edit: Meant to comment this on the reps one ([mozilla-releng/reps#92](https://github.com/mozilla-releng/reps/pull/92)), got bamboozled by the same PR being opened in all repos that required it and me not checking. But discussing it here works too.

As I understand it, yeah, it could. The alternative is set a wider hour range (eg: 0-6 like before), but that still runs the risk of multiple PRs a day (and also missing PRs entirely). My opinion is that now that we're restricting the day of the month that running at any hour ought to be fine, even if we occasionally get more than one PR for the same thing in the same repo on the two days out of the month that it runs.

jcristau (approved):

PR #1471, opened 2026-06-15
feat: support providing to_revision and from_revision in merge day payloads

bhearsum:
This will allow us to make these tasks more idempotent, and avoid the issue described in https://bugzilla.mozilla.org/show_bug.cgi?id=2047586.

In the short term, we need to fall back to using branch refs. Once I update all of the callers/task creators, we can drop these fallbacks and make `to_revision` required (and `from_revision` when appropriate) to help keep things safer in the future.

Once this is in production I'll need to make some changes to `mozilla-taskgraph`, Gecko, and xpi-manifest to get everything using the new payload format, then I can follow-up here to make the new `revision` parameters required and ensure we always operate safely.

### DIFF
--- a/landoscript/src/landoscript/actions/merge_day.py
+++ b/landoscript/src/landoscript/actions/merge_day.py
@@ -35,7 +35,10 @@ class VersionFile:
 class MergeInfo:
     to_branch: str
     fetch_version_from: str
+    # TODO: to_revision should be required after all callers are updated to use it
+    to_revision: str = ""
     from_branch: str = ""
+    from_revision: str = ""
     base_tag: str = ""
     end_tag: str = ""
     merge_old_head: bool = False
@@ -44,6 +47,9 @@ class MergeInfo:
     version_files: list[VersionFile] = field(default_factory=list)
     replacements: list[list[str]] = field(default_factory=list)
     regex_replacements: list[list[str]] = field(default_factory=list)
+
+    # TODO: add __post_init__ to require `from_revision` when `from_branch` is
+    # present after all callers are already doing so
 
     @classmethod
     def from_payload_data(cls, payload_data) -> Self:
@@ -59,6 +65,8 @@ async def run(
 ) -> list[LandoAction]:
     to_branch = merge_info.to_branch
     from_branch = merge_info.from_branch
+    to_revision = merge_info.to_revision
+    from_revision = merge_info.from_revision
     end_tag = merge_info.end_tag
     base_tag = merge_info.base_tag
     merge_old_head = merge_info.merge_old_head
@@ -67,9 +75,23 @@ async def run(
     actions = []
 
     log.info("Starting merge day operations!")
-    to_version = await get_version(github_client, version_file, to_branch)
+    if not to_revision:
+        # TODO: remove this fallback once `to_revision` is required
+        to_revision = await github_client.get_branch_head_oid(to_branch)
+
+    to_version = await get_version(github_client, version_file, to_revision)
+
     log.info(f"to_version is: {to_version}")
     if end_tag:
+        # `end_revision` is purposely made to be whatever the current tip of `to_branch`
+        # is when a landoscript task runs (as opposed to, eg: using the `to_revision`).
+        # This guards against a task being created, something being pushed to the
+        # `to_branch`, and then the `END` tag ending up on an outdated revision.
+        # Even in a case where multiple versions of the same task are scheduled
+        # and race against one another the tag cannot end up in the wrong place so
+        # long as the `version` in it was fetched from a concrete revision. (If a race
+        # does take place, the second task will attempt to move a tag, which will
+        # fail the task.)
         end_revision = await github_client.get_branch_head_oid(to_branch)
         # End tag specifically uses the `to_version` _before_ we bump it
         # (because we're declaring its current version as "done")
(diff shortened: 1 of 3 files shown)


bhearsum:
> whatever is fetching the versions, like ship it, would need to be sure not to fetch from the tip of a branch

It occurs to me that we might be able to pass the revisions instead or in addition to the branches. We actually already look up the tip of the branch revs for other reasons, which is another source of potential inconsistency, so this might be good to do regardless.

If we took this route, I think we could continue to look up versions safely in landoscript.

bhearsum:
I updated the patch to take revisions instead of versions; I think this will be a much better approach. I still need to do some more real world testing on Try before this is ready for review.

bhearsum:
This got a test on Try with the `main-to-beta` and `bump-main` scenarios. It's not really easy to specifically arrange the same circumstances we saw in production, but I ran the same action with the same input multiple times to try to force as much racing as possible. For `main-to-beta` I used:
```
force-dry-run: false
behavior: main-to-beta
to-revision: c2eb8775b1e5aa7039527c4151842f3f7fed733e
from-revision: 0d246fab5df475bca94501bd5b120f383b91dedf
```

And for `bump-main` I used:
```
force-dry-run: false
behavior: main-to-beta
to-revision: 0d246fab5df475bca94501bd5b120f383b91dedf
```

In the first case, the first `main-to-beta` task succeeded and the others failed (after some log inspection, because the git tag already existed on the Lando workers).

In the second case, the first `bump-main` succeeded and the subsequent ones failed with merge conflicts errors. This is because they're basing the patches on the now-outdated revision in the action input, but Lando is trying to apply them to `main` (which has been bumped). This ends up causing a MergeConflictError in Lando, which causes reruns (that then fail in the same way). This is probably not ideal, but still preferable to double bumping. (I'm tempted to say we should remove the automatic rerun for merge conflict errors...but that requires more thought before doing it.)

This was all done on [this try push](https://treeherder.mozilla.org/jobs?repo=try&revision=b8a963313f6e362e9073e0038052a1269952de8d&searchStr=merge).

landoscript/src/landoscript/actions/merge_day.py
```diff
@@ -84,10 +100,17 @@ async def run(
...
+        if from_revision:
+            bump_version = await get_version(github_client, version_file, from_revision)
+            end_revision = from_revision
+        else:
+            # TODO: remove this fallback once `from_revision` is being passed whenever
+            # `from_branch is
+            bump_version = await get_version(github_client, version_file, from_branch)
+            end_revision = await github_client.get_branch_head_oid(from_branch)
+
+        bump_revision = from_revision
```
jcristau: from_revision can be None here, wouldn't that be a problem?

jcristau (changes requested):
Would be nice to have a description of what the parameters mean / are used for, maybe in a description field in the schema?

landoscript/src/landoscript/actions/merge_day.py
bhearsum: Right, indeed - this was supposed to be using `end_revision` :facepalm:.

landoscript/src/landoscript/actions/merge_day.py
```diff
@@ -67,10 +75,18 @@ async def run(
...
+        # TODO: remove this fallback once `to_revision` is required
+        to_version = await get_version(github_client, version_file, to_branch)
     log.info(f"to_version is: {to_version}")
     if end_tag:
-        end_revision = await github_client.get_branch_head_oid(to_branch)
+        if to_revision:
+            end_revision = to_revision
+        else:
+            # TODO: remove this fallback once `to_revision` is required
+            end_revision = await github_client.get_branch_head_oid(to_branch)
```
jcristau: I'm not sure about the change for the _END tag.  If we're "merging" branch A to branch B, and we're racing with some other push to branch B, it seems to me the _END tag should still point at whatever was head of B just before the merge, not when the task payload was built?
(Admittedly on merge day branch B is supposed to be quiet, but I could imagine an l10n bump or similar automated commit slipping in, even if manual uplifts aren't happening.)

bhearsum: Hrm. So my thinking here was that if we had multiple tasks running by accident again, task A would merge `from_branch` to `to_branch`, and then task B would end up tagging that new tip with `_END`, which would be wrong. However, task A would already have tagged in that situation, so task B would fail anyways.

So, I think we can safely keep the tagging logic as-is as long as we're fetching versions from a fixed revision. Any races would result in attempting to move a tag, which would fail. (This may become untrue at some point in the future if Lando doesn't fail when a tag already exists...but I don't think we need to plan for that hypothetical right now.)

landoscript/src/landoscript/actions/merge_day.py
```diff
@@ -108,11 +140,11 @@ async def run(
             raise TaskVerificationError("'from_branch' is required when merge_old_head is True or version_files are present")
 
         bump_version = to_version
-        bump_branch = to_branch
+        bump_revision = to_revision
```
jcristau, line 144: to_revision is optional, don't we need a fallback here?  (is this covered by the "legacy" tests?)

landoscript/src/landoscript/data/landoscript_task_schema.json
```diff
@@ -97,6 +97,14 @@
                         "beta"
                     ]
                 },
+                "from_revision": {
+                    "type": "string",
+                    "description": "The revision on the from_branch to fetch its current version from, tag with `_BASE` tags, and use as the target revision for merge actions."
+                },
+                "to_revision": {
+                    "type": "string",
+                    "description": "The revision on the to_branch to fetch its current version from and tag with `_END` tags."
```
jcristau: the part about the _END tags is wrong now?

jcristau (changes requested):
I think this is close, just a couple of bits to double check.

landoscript/src/landoscript/actions/merge_day.py
bhearsum: Good catch, thanks. This code path is covered by tests, but unfortunately the mocked graphql responses are very dumb, and don't account for the inputs (they just return a sequence of different responses). (This is mostly because the graphql endpoint is generic, so there's extra hoops to jump through if you want mocked responses accounting for inputs.)

landoscript/src/landoscript/data/landoscript_task_schema.json
bhearsum: Right - it was when I wrote it, but not after addressing other comments. ty.

landoscript/src/landoscript/actions/merge_day.py
```diff
@@ -67,9 +75,23 @@ async def run(
     actions = []
 
     log.info("Starting merge day operations!")
-    to_version = await get_version(github_client, version_file, to_branch)
+    if to_revision:
+        to_version = await get_version(github_client, version_file, to_revision)
+    else:
+        # TODO: remove this fallback once `to_revision` is required
+        to_revision = await github_client.get_branch_head_oid(to_branch)
+        to_version = await get_version(github_client, version_file, to_branch)
```
jcristau: nit:
```suggestion
    if not to_revision:
        # TODO: remove this fallback once `to_revision` is required
        to_revision = await github_client.get_branch_head_oid(to_branch)
    to_version = await get_version(github_client, version_file, to_revision)
```

jcristau (approved):

landoscript/src/landoscript/actions/merge_day.py
bhearsum: Good call.

PR #1475, opened 2026-06-25
feat!: add support to publish to huawei store

abhishekmadan30:


pushapkscript/tests/integration/test_integration_script.py
```diff
@@ -420,5 +436,40 @@ def test_main_with_samsung_store(self, push_apk):
...
+            track=None,
+            expected_package_names=["org.mozilla.fenix"],
+            store="huawei",
+            rollout_percentage=None,
+            dry_run=False,
+            contact_server=True,
+            skip_check_multiple_locales=True,
+            skip_check_ordered_version_codes=False,
+            skip_check_same_locales=True,
+            skip_checks_fennec=True,
```
JohanLorenzo, line 461: Doesn't block the merge: What are the reasons for skipping the Fennec checks? For reference, this is what they do: 

https://github.com/mozilla-releng/mozapkpublisher/blob/1dd73f8157075719dca9dfe4116a533ad3680cb4/mozapkpublisher/common/apk/checker.py#L23-L31

This flag was introduced in https://github.com/mozilla-releng/mozapkpublisher/pull/156 to support more products than Firefox itself. Firefox for Android has its quirks compare to simpler apps.

pushapkscript/pyproject.toml
```diff
@@ -13,6 +13,7 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
 ]
 dependencies = [
+    # TODO: bump to >=12.0.0 once mozapkpublisher 12.0.0 (huawei store) is on PyPI
```
JohanLorenzo, line 16: Question: what was your experience between mozapkpublisher and pushapkscript? Did the split make it easier to test changes locally? If you were to implement a worker that publishes to Android stores, would you split these 2 packages?

I don't have any preferences in the answer, I just want to gauge if this split is still worth it today.

JohanLorenzo (approved):
LGTM! 👍 As rightfully documented in `pushapkscript/pyproject.toml`, this depends on https://github.com/mozilla-releng/mozapkpublisher/pull/325. I'm approving it because I don't believe this PR is going to change a lot so no need for my approval again to get it landed.

One more question: how did the testing go beyond the unit tests in this PR?

pushapkscript/tests/integration/test_integration_script.py
abhishekmadan30: Samsung doesn't run them either, so I just matched it: test_integration_script.py#L429 (https://github.com/mozilla-releng/scriptworker-scripts/blob/4e5ca343de5669c7ad492e0fd6f819ec96d551ba/pushapkscript/tests/integration/test_integration_script.py#L429)

PR #1481, opened 2026-07-02
bouncerscript: retry requests to bounceradmin on 5xx

jcristau:


bhearsum (changes requested):
It looks like this currently retries ClientError (really...why?) and ServerTimeoutError. If I'm reading https://docs.aiohttp.org/en/latest/web_exceptions.html correctly, we'll want to retry on HTTPServerError.

jcristau:
> It looks like this currently retries ClientError (really...why?) and ServerTimeoutError. If I'm reading https://docs.aiohttp.org/en/latest/web_exceptions.html correctly, we'll want to retry on HTTPServerError.

These are server-side exceptions (aiohttp does both).  The client side is https://docs.aiohttp.org/en/stable/client_reference.html#client-exceptions, in our case ClientResponseError, a subclass of ClientError.

bhearsum (approved):
ah my bad, ty

PR #1482, opened 2026-07-14
chore: Update dependencies

hneiva:


uv.lock
```diff
@@ -4116,23 +4131,23 @@ resolution-markers = [
...
-    { name = "requests" },
-    { name = "roman-numerals" },
-    { name = "snowballstemmer" },
-    { name = "sphinxcontrib-applehelp" },
-    { name = "sphinxcontrib-devhelp" },
-    { name = "sphinxcontrib-htmlhelp" },
-    { name = "sphinxcontrib-jsmath" },
-    { name = "sphinxcontrib-qthelp" },
-    { name = "sphinxcontrib-serializinghtml" },
+    { name = "alabaster", marker = "python_full_version >= '3.12'" },
```
bhearsum, line 4134: This is probably fine, but I don't really understand why it's here? I presume this patch was generated with `uv lock -U` or such?

bhearsum (approved):

uv.lock
hneiva: Yup. I just ran `./maintenance/pin.sh`

PR #1491, opened 2026-07-17
bug 2055645: support direct git tags in landoscript

bhearsum:
Landoscript supports requesting tags to be added to repositories. The [underlying tag action](https://github.com/mozilla-releng/scriptworker-scripts/blob/1cb7c72fcc9e5523c9c917fc8fa3a0ca620d4b2d/landoscript/src/landoscript/actions/tag.py) supports hg and git repositories, but git support is [only wired in when used by the merge day action](https://github.com/mozilla-releng/scriptworker-scripts/blob/1cb7c72fcc9e5523c9c917fc8fa3a0ca620d4b2d/landoscript/src/landoscript/actions/merge_day.py#L100)). The toplevel tag action support [hardcodes hg](https://github.com/mozilla-releng/scriptworker-scripts/blob/1cb7c72fcc9e5523c9c917fc8fa3a0ca620d4b2d/landoscript/src/landoscript/script.py#L93).

In the immediate term, this means we can't make early-tagging and version-bump tasks work on Try (because there's no hg-to-git mapping we can look up there). In the medium term, this will be needed when we start firing release promotion from git events, and won't have an hg revision to feed to these tasks.

I'm not _super_ happy with controlling hg vs. git based on the presence of hg_repo_url...but seeing as hg support as a finite lifespan, it's probably fine.

### DIFF
--- a/landoscript/tests/test_tag.py
+++ b/landoscript/tests/test_tag.py
@@ -59,30 +59,47 @@ async def test_success(aioresponses, github_installation_responses, context, tag
 
 
 @pytest.mark.asyncio
-async def test_no_tags(aioresponses, github_installation_responses, context):
+@pytest.mark.parametrize(
+    "tag_info,dry_run",
+    (
+        pytest.param(
+            {
+                "revision": "ghijkl654321",
+                "tags": ["BUILD1"],
+            },
+            True,
+            id="dry_run",
+        ),
+        pytest.param(
+            {
+                "revision": "ghijkl654321",
+                "tags": ["BUILD1"],
+            },
+            False,
+            id="one_tag",
+        ),
+        pytest.param(
+            {
+                "revision": "ghijkl654321",
+                "tags": ["BUILD1", "RELEASE"],
+            },
+            False,
+            id="multiple_tags",
+        ),
+    ),
+)
+async def test_success_git(aioresponses, github_installation_responses, context, tag_info, dry_run):
     payload = {
         "actions": ["tag"],
         "lando_repo": "repo_name",
-        "tag_info": {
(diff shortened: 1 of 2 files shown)


bhearsum:
This was tested on Try for both [version bump](https://firefox-ci-tc.services.mozilla.com/tasks/TmX3dA_GTcKCsSAZZ9GtLg) and [early tagging](https://firefox-ci-tc.services.mozilla.com/tasks/VMiANxGZQ2WMEXtc5GylSQ) tasks, which added the appropriate tags on [this commit](https://github.com/mozilla-releng/staging-firefox/commit/b4475b554061f50db396f90182929f3389b4961d).

landoscript/tests/test_tag.py
```diff
@@ -59,30 +59,61 @@ def assert_func(req):
...
+
+    def assert_func(req):
+        assert_tag_response(req, tag_info, tag_info["revision"])
+
+    await run_test(aioresponses, github_installation_responses, context, payload, ["tag"], not dry_run, assert_func)
 
 
 @pytest.mark.asyncio
-async def test_hg_repo_url(aioresponses, github_installation_responses, context):
+async def test_no_tags(aioresponses, github_installation_responses, context):
```
jcristau: Presumably in a real task this would error out earlier when validating against the schema?

jcristau (approved):

landoscript/tests/test_tag.py
bhearsum: Yes, that's a good point. It doesn't here because `scriptworker` isn't in the mix, which is what is doing the schema validation (ugh). I'm going to remove this test given that...

PR #1492, opened 2026-07-21
ci: add missing UV_VERSION arg to test images

jcristau:


jcristau:
Per https://bugzilla.mozilla.org/show_bug.cgi?id=2047074#c13, kaniko would previously have resolved `FROM ghcr.io/astral-sh/uv:` (empty version) to the `latest` tag.  With the update in image_builder 6.2.0 (#1488), it now errors out, which uncovered this bug.

Eijebong (approved):
Fun

PR #1493, opened 2026-07-22
feat(iscript): remove notarization support

hneiva:
Notarization has moved to signingscript, so remove all notarization behaviors, helper functions, exceptions, config keys, and task-schema entries from iscript, leaving only the signing and pkg-building behaviors.


hneiva:
Note to future me:
`arrow` package is now unused and can be removed
I didn't want to muddle this PR more than it needs.

bhearsum (approved):
This looks reasonable; ideally you'll sanity check on dev before this rolls out to prod though?

hneiva:
Ran it on my local testing mac mini signer. WFM 😃

PR #1495, opened 2026-07-27
chore(iscript): Fix python dependency list

hneiva:
- Add requests (was being imported by another package, but used explicitly)
- Removes aiohttp, arrow, macholib, taskcluster (unused)

### DIFF
--- a/iscript/pyproject.toml
+++ b/iscript/pyproject.toml
@@ -13,16 +13,16 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
 ]
 dependencies = [
-    "aiohttp",
-    "arrow>=1.0",
     "attrs",
-    "macholib",
     "mozbuild",
     "pexpect",
+    "requests",
     "requests-hawk",
     "scriptworker-client",
-    "taskcluster",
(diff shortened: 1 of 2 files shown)


iscript/pyproject.toml
```diff
@@ -13,16 +13,16 @@ classifiers = [
     "Programming Language :: Python :: 3.11",
 ]
 dependencies = [
-    "aiohttp",
-    "arrow>=1.0",
     "attrs",
-    "macholib",
     "mozbuild",
```
bhearsum, line 17: Are you sure `mozbuild` is still needed here? I don't see any direct usage of it, nor any indirect usage of it through `scriptworker_client`.

hneiva: There's a `from mozpack import mozjar` in autograph.py 
And `mozpack` is provided by `mozbuild` 🤷

bhearsum: Oh; I was expecting the mozbuild imports to be...well...prefixed by mozbuild :(

bhearsum (approved):

PR #1496, opened 2026-07-28
Bug 2058413 - update k8s scriptworkers to python 3.14

jcristau:


bhearsum (commented):
Is there a reason that [pyproject.toml](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/pyproject.toml), [the toplevel tox.ini](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/tox.ini), and [script-specific tox.ini](https://github.com/mozilla-releng/scriptworker-scripts/blob/master/balrogscript/tox.ini) files shouldn't be updated as well? Possibly also [the check kind](https://github.com/mozilla-releng/scriptworker-scripts/blob/4a3f76e232c710cfc826d11e32d36859d3ea83c4/taskcluster/kinds/check/kind.yml#L22)?

(Presumably scriptworker_client and iscript still need 3.11 though?)

jcristau:
@bhearsum updated tox.ini and the check kind.  I don't think we can update pyproject.toml just yet since it's global and we still have some tasks using 3.11?

bhearsum:
> @bhearsum updated tox.ini and the check kind. I don't think we can update pyproject.toml just yet since it's global and we still have some tasks using 3.11?

Yeah, good point (I didn't realize that was just the `ruff` `target-version` when I glanced at it...).

scriptworker_client/tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py311,mypy
```
bhearsum, line 2: We should probably maintain 3.11 testing here seeing as `iscript` depends on it.

bhearsum (changes requested):

scriptworker_client/tox.ini
jcristau: envlist is just what gets run by default when you don't pass `-e` to tox.  `tox -e py311` still works and remains in CI.

bhearsum: ah, right, ty

bhearsum (approved):

bhearsum:
Thanks for doing this!

PR #1500, opened 2026-08-07
bug 2061475: switch adhoc to new gpg signing key; add rpm signing to adhoc-signing

bhearsum:
I am _explicitly_ not adding a new format here, as we sometimes do in the past. Because we're planning to revoke the previous key it's better to stop signing with it ASAP.

Note that RPM signing on adhoc is new; I'll be adding secrets for these to sops.


Eijebong (commented):
Presumably we want to add:


```
test_var_set 'AUTOGRAPH_RPMSIGN_USERNAME'
test_var_set 'AUTOGRAPH_RPMSIGN_PASSWORD'
```

In `init_worker.sh` in the adhoc sections to get a proper error message instead of a json-e rendering failure if the secrets are missing?

Eijebong (approved):

PR #1508, opened 2026-08-19
Bug 2062342 - signingscript: expose RPM-KEY artifact

jcristau:
RPM package managers don't all deal well with revoked subkeys, so filter them out of the public key artifact.

Per @eijebong:

> Old version of dnf won't even import a key that contains a hard
> revoked key (expired keys are fine). We'll have to publish the key
> with the compromised one omitted from it instead. On top of that old
> versions of dnf don't cleanly reimport keys when the main one is the
> same. They say they do, return 0, and then don't do it... dnf5 which
> is what recent-ish fedora runs self-heals but that's a fraction of
> users.

### DIFF
--- a/signingscript/src/signingscript/script.py
+++ b/signingscript/src/signingscript/script.py
@@ -5,6 +5,7 @@ import asyncio
 import json
 import logging
 import os
+import tempfile
 from dataclasses import asdict
 
 import aiohttp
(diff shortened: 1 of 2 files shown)


signingscript/src/signingscript/script.py
```diff
@@ -83,6 +90,51 @@ async def async_main(context):
...
+
+    Args:
+        context (Context): the signing context.
+        dest (str): the path to write the filtered pubkey to.
+
+    """
+    with tempfile.TemporaryDirectory(dir=context.config["work_dir"]) as gnupghome:
+        await _run_gpg(gnupghome, ["--import", context.config["gpg_pubkey"]])
+        with open(dest, "wb") as pubkey:
+            await _run_gpg(gnupghome, ["--armor", "--export-filter", "drop-subkey=revoked -t", "--export"], stdout=pubkey)
```
Eijebong, line 135: `"drop-subkey=revoked -t"` looks like a typo but isn't... What a lovely CLI tool

Eijebong (approved):
Thank you

PR #1514, opened 2026-09-04
fix(pushapk): Actually check apk signatures

hneiva:
## Summary

pushapkscript has never verified an APK or AAB signature, for any product, in any environment. `skip_check_signature` was read from the wrong config dict, so the `jarsigner` and `MANIFEST.MF` checks in `script.py` were dead code. This predates the initial monorepo import.

Switching them on needed four more fixes, one of which would otherwise have broken production.

## The bug

`script.py`:

```python
if not publish_config.get("skip_check_signature", True):
```

`skip_check_signature` is declared at the *product* level of the config schema, but this reads it from `publish_config`, the per-app/channel/store dict built by `get_publish_config()`. That function never emits the key, so the lookup always missed and always fell back to the default.

Two bugs, not one:

1. **Wrong dict.** The option is unreachable.
2. **Wrong default.** `True` means skip, and the `else` branch claims the product is configured with the option, but that branch is what ran when the key was *absent*, which is every product in `worker.yml`.

Fixing only (1) changes nothing, since nothing sets the option. Both had to move together. No test covered either branch, which is why this survived.

## What else was blocking it

**`jarsigner` was not installed anywhere.** Both the deployed image and the test image installed `default-jre-headless`, which ships `keytool` but not `jarsigner`; jarsigner is JDK only. Verified on `debian:bookworm`, where the JRE package provides just `/usr/bin/keytool` while `default-jdk-headless` provides both. Turning verification on without this would have failed every task in production.

**Samsung and Huawei had no `certificate_alias`.** The non-Google branch of `_get_channel_publish_config` never copied it into its result, so `jarsigner._pluck_configuration` raised `KeyError` on it.

**`mozillavpn` dev/fake-prod has no alias at all,** and `init_worker.sh` imports no certificate for it ("no dep signing for mozillavpn yet"). A `None` alias reached the jarsigner argv as a `TypeError` from subprocess.

**The test fixtures cannot be verified.** The APKs date from 2017 and use SHA1withDSA with a 1024-bit key. Modern JDKs disable both, so jarsigner reports them as unsigned and `-verify -strict` exits 16. Their manifests also list entries that were stripped out of the archives, which `-strict` rejects on its own.

## Commits

- **`set certificate_alias for non-Google target stores`** One line, matching what the Google paths already did. The alias identifies the certificate the incoming artifact was signed with, which the upstream signing task decides, so it is the same whichever store the artifact goes to. Huawei support, currently in flight, needs no further change.
- **`fail loudly when no certificate_alias is configured`** `ConfigValidationError` naming the store and pointing at `skip_check_signature`, instead of a `KeyError` or a `TypeError` from inside subprocess.
- **`install the JDK, so jarsigner exists`** `default-jdk-headless` in both the deployed and the test image.
- **`sign the integration fixtures so they can be verified`** Generate a throwaway RSA-2048 keypair per test and re-sign each fixture under the alias its config expects, after dropping the stale `META-INF` so jarsigner writes a fresh manifest. The generated configs now declare SHA-256, which is what `worker.yml` uses in production. Also fixes `test_main_with_samsung_store`, which signed for `nightly` while the fenix release app it targets declares `fenix-production`.
- **`verify APK signatures, which never actually ran`** Read `product_config`, default to verifying. `mozillavpn` non-prod opts out explicitly, as the one config that genuinely cannot verify.

## Testing

- 91 pass in a `default-jdk-headless` container, with the integration tests genuinely running `jarsigner -verify -strict` against freshly signed APKs.
- 84 pass and 7 skip locally without a JDK. Detection runs the binaries rather than looking them up, because macOS ships stubs on PATH that exist but exit non-zero.
- 264 repo-level init tests pass. ruff, ruff-format and yamllint are clean.
- The new `test_script.py` test fails against the old code with `assert 0 == 2`, so it is a real regression test.
- Audited every rendered `worker.yml` combination for a missing alias. `mozillavpn` dev/fake-prod is the only one, and it opts out.

## Risk

This enables a check that has been dead since before the monorepo import, so production artifacts have never been through `jarsigner -verify -strict` on a modern JDK. Autograph signs with PKCS#7 and SHA-256, so it should hold, but the fixtures show that weak-algorithm rejection is not hypothetical.

Worth watching the first dev/fake-prod task before this reaches prod. `skip_check_signature: true` on a product is the escape hatch if some artifact turns out not to verify.

### DIFF
--- a/pushapkscript/src/pushapkscript/jarsigner.py
+++ b/pushapkscript/src/pushapkscript/jarsigner.py
@@ -1,7 +1,7 @@
 import logging
 import subprocess
 
-from pushapkscript.exceptions import SignatureError
+from pushapkscript.exceptions import ConfigValidationError, SignatureError
 
 log = logging.getLogger(__name__)
 
@@ -36,4 +36,11 @@ def _pluck_configuration(context, publish_config):
     # Uses jarsigner in PATH if config doesn't provide it
     binary_path = context.config.get("jarsigner_binary", "jarsigner")
 
-    return binary_path, keystore_path, publish_config["certificate_alias"]
+    certificate_alias = publish_config.get("certificate_alias")
+    if not certificate_alias:
+        raise ConfigValidationError(
+            'No "certificate_alias" is configured for the "{}" target store, so the signature cannot be verified. Either configure one or set '
+            '"skip_check_signature" on the product.'.format(publish_config.get("target_store"))
+        )
+
+    return binary_path, keystore_path, certificate_alias
--- a/pushapkscript/src/pushapkscript/jarsigner.py
+++ b/pushapkscript/src/pushapkscript/jarsigner.py
@@ -1,16 +1,33 @@
 import logging
+import os
 import subprocess
 
 from pushapkscript.exceptions import ConfigValidationError, SignatureError
 
 log = logging.getLogger(__name__)
 
+# Re-enables SHA-1 for verification only. Without it, `-strict` exits 16 on every
+# artifact Autograph currently produces. See the file itself and bug 1838680.
+SECURITY_PROPERTIES_PATH = os.path.join(os.path.dirname(__file__), "data", "jarsigner.java.security")
+
 
 def verify(context, publish_config, apk_path):
     binary_path, keystore_path, certificate_alias = _pluck_configuration(context, publish_config)
 
     completed_process = subprocess.run(
-        [binary_path, "-verify", "-strict", "-verbose", "-keystore", keystore_path, apk_path, certificate_alias],  # Needed to check the digest algorithm
+        [
+            binary_path,
+            "-verify",
+            # Without `-strict`, jarsigner exits 0 even when the archive was signed by a
+            # certificate other than `certificate_alias`, which is the whole point here.
+            "-strict",
+            "-verbose",  # Needed to check the digest algorithm
+            "-J-Djava.security.properties={}".format(SECURITY_PROPERTIES_PATH),
+            "-keystore",
+            keystore_path,
+            apk_path,
+            certificate_alias,
+        ],
         stdout=subprocess.PIPE,
         stderr=subprocess.STDOUT,
         universal_newlines=True,
--- a/pushapkscript/src/pushapkscript/script.py
+++ b/pushapkscript/src/pushapkscript/script.py
@@ -36,7 +36,10 @@ async def async_main(context):
         # Google Play won't accept both APK and AAB
         raise TaskVerificationError("The configuration is invalid: Unable to push both APK and AAB files for the same product.")
 
-    if not publish_config.get("skip_check_signature", True):
+    # `skip_check_signature` is a product-level option, and omitting it means "do verify".
+    if product_config.get("skip_check_signature", False):
+        log.warning('This product is configured with "skip_check_signature", so the signing of the files will not be verified.')
+    else:
         log.info("Verifying APKs' signatures...")
         for apk_path in all_apks_paths:
             jarsigner.verify(context, publish_config, apk_path)
@@ -45,8 +48,6 @@ async def async_main(context):
         for aab_path in all_aabs_paths:
             jarsigner.verify(context, publish_config, aab_path)
             manifest.verify(product_config, aab_path)
-    else:
-        log.info('This product is configured with "skip_check_signature", so the signing of the files will not be verified.')
 
     log.info("Delegating publication to mozapkpublisher...")
     with contextlib.ExitStack() as stack:
--- a/pushapkscript/tests/integration/test_integration_script.py
+++ b/pushapkscript/tests/integration/test_integration_script.py
@@ -5,7 +5,10 @@ import shutil
 import subprocess
 import tempfile
 import unittest
+
 import pytest
+
+from scriptworker.constants import STATUSES
 
 import pushapkscript
 from pushapkscript.script import main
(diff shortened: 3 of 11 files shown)


JohanLorenzo (commented):
> pushapkscript has never verified an APK or AAB signature, for any product, in any environment.

While mostly true, this is an overstatement. We used to check the signature of Fennec (the former Fenix) builds. Looking back at the history, I believe I made a mistake when reviewing the [introduction of this if statement](https://github.com/mozilla-releng/pushapkscript/pull/112/changes#diff-963480320f135664fb6313236426541bf76efed3b4995d156b73935dbe5206d8R36). 

It should have defaulted to checking the signature and that's not what the implementation does.

>  That function never emits the key, so the lookup always missed and always fell back to the default.

That's another thing I missed, yeah.

> **jarsigner was not installed anywhere**

It [used to be](https://github.com/mozilla-releng/pushapkscript/pull/99/changes#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R7). It was changed not too long ago in [#1399](https://github.com/mozilla-releng/scriptworker-scripts/pull/1399/changes#diff-044a59cbb025a6de64afb476663d3d23625bc2c8968db6f1726a89e7eb90a581L12). 

> The APKs date from 2017 and use SHA1withDSA with a 1024-bit key. Modern JDKs disable both, so jarsigner reports them as unsigned and `-verify -strict` exits 16. 

Yup. That's tracked under [bug 1838680#c6](https://bugzilla.mozilla.org/show_bug.cgi?id=1838680#c6)

pushapkscript/tests/test_publish_config.py
```diff
@@ -200,6 +204,24 @@ def test_target_samsung_submit():
     }
 
 
+def test_certificate_alias_does_not_depend_on_the_target_store():
+    # The alias identifies the certificate the incoming artifact was signed with, which is
+    # decided by the upstream signing task, so it is the same whichever store it goes to.
+    google = get_publish_config(FENIX_CONFIG, {"channel": "production", "target_store": "google"}, "fenix")
+    samsung = get_publish_config(FENIX_CONFIG, {"channel": "production", "target_store": "samsung"}, "fenix")
```
JohanLorenzo, line 211: This commit overall makes sense to me. We originally didn't add the certificate alias because the Amazon store shipped a different product than Firefox for Android, if I remember correctly.

pushapkscript/src/pushapkscript/jarsigner.py
```diff
@@ -36,4 +36,11 @@ def _pluck_configuration(context, publish_config):
     # Uses jarsigner in PATH if config doesn't provide it
     binary_path = context.config.get("jarsigner_binary", "jarsigner")
 
-    return binary_path, keystore_path, publish_config["certificate_alias"]
+    certificate_alias = publish_config.get("certificate_alias")
+    if not certificate_alias:
+        raise ConfigValidationError(
```
JohanLorenzo, line 58: Thanks for raising this error. That makes sense we're hardening this piece.

taskcluster/docker/pushapkscript/Dockerfile
```diff
@@ -8,8 +8,10 @@ FROM $DOCKER_IMAGE_PARENT
 ADD --chown=app:app topsrcdir/pushapkscript /app/pushapkscript
 
 USER root
+# jarsigner verifies the signature of the APKs/AABs before they are published, and it
+# ships in the JDK rather than the JRE, so this needs default-jdk-headless.
 RUN apt-get update \
- && apt-get install --no-install-recommends -y curl default-jre-headless \
```
JohanLorenzo, line 12: Note to future readers: this effectively reverts #1399

pushapkscript/src/pushapkscript/script.py
```diff
@@ -36,7 +36,10 @@ async def async_main(context):
         # Google Play won't accept both APK and AAB
         raise TaskVerificationError("The configuration is invalid: Unable to push both APK and AAB files for the same product.")
 
-    if not publish_config.get("skip_check_signature", True):
+    # `skip_check_signature` is a product-level option, and omitting it means "do verify".
+    if product_config.get("skip_check_signature", False):
+        log.warning('This product is configured with "skip_check_signature", so the signing of the files will not be verified.')
```
JohanLorenzo, line 41: Thanks for making it a warning too.

pushapkscript/tests/integration/test_integration_script.py
```diff
@@ -5,42 +5,114 @@
...
+    return True
+
+
+requires_jdk = pytest.mark.skipif(
+    not _has_working_jdk(),
+    reason="signing and verifying the test APKs needs keytool and jarsigner from a JDK",
+)
+
+
+def _strip_v1_signature(apk_path):
```
JohanLorenzo: Blocker: I'd prefer not to go down that path until [bug 1838680](https://bugzilla.mozilla.org/show_bug.cgi?id=1838680#c6) is fixed. Let's have the integration test as close to production and make sure we pass the right flags to `jarsigner`. With the current implementation, we will end up with a green test while not knowing if we are going to break the pushapk tasks at all.

JohanLorenzo (changes requested):
Thanks for discovering this and for taking the time to take a stab at it. You uncovered a regression that occurred 7 years ago had never been found since then. 

I agree with the commits overall. The one thing that worries me is reenabling `jarsigner` without testing it locally. The integration test is on the right path. In my opinion, we should replicate what happens on the worker as much as possible, which means keeping the existing signatures with SHA-1 hashes. 

What do you think?

hneiva:
I agree, I think we should check SHA-1 hashes until autograph is ready to move algos.

I'll add a comment to the bug once this lands reminding to switch the manifest to SHA-256 once ready.

pushapkscript/tests/integration/test_integration_script.py
```diff
@@ -422,3 +465,21 @@ def test_main_with_samsung_store(self, push_apk):
             sgs_access_token="456",
             submit=False,
         )
+
+    @unittest.mock.patch("pushapkscript.publish.push_apk")
+    def test_main_rejects_an_apk_signed_by_another_certificate(self, push_apk):
+        """The check has to actually bite. The fixtures carry the Fennec nightly
+        certificate, so trusting a different one under the alias the config expects must
+        stop the task before anything is published."""
```
JohanLorenzo, line 473: Nit: Useless comment. This information is already captured by the function name.

JohanLorenzo (approved):
Looks good to me! Thanks for fixing this check broken for so many years!

PR #1519, opened 2026-09-14
feat(pushapk): Merge mozapkpublisher repo into pushapkscript

hneiva:
`pushapkscript` is the only consumer of `mozapkpublisher`, the whole integration surface is two imports in `publish.py`, and to our knowledge nothing else uses it.

Yet every change to the publisher needed a full release before `pushapkscript` could use it: land it upstream, tag, publish to PyPI, bump the pin here, relock. A one-line fix cost a release cycle before the only thing that calls it could
exercise it.
Side by side, the publisher and pushapkscript can change in the same pull request, reviewed and tested together.

Note: migrated only 293 commits, not all 428. `git-filter-repo` pruned 135 commits that became empty once `.github/`, `uv.lock`, `pyproject.toml` etc. were dropped. All commits affecting the imported code are present.

Prefixed all imported commits with `[mozapkpublisher]` and replaced issue numbers with the right reference.


pushapkscript/tox.ini
```diff
@@ -8,7 +8,7 @@ setenv =
 runner = uv-venv-lock-runner
 package = editable
 commands=
-    {posargs:py.test --cov-config=tox.ini --cov-append --cov={toxinidir}/src/pushapkscript --cov-report term-missing tests}
+    {posargs:py.test --cov-config=tox.ini --cov-append --cov={toxinidir}/src/pushapkscript --cov={toxinidir}/src/mozapkpublisher --cov-report term-missing tests src/mozapkpublisher/test}
```
JohanLorenzo, line 11: Good idea on putting `mozapkpublisher` under `pushapkscript/src/mozapkpublisher`. That enables to preserve the history as much as possible.

pushapkscript/tox.ini
```diff
@@ -28,4 +28,4 @@ python_files = test_*.py
 
 [coverage:run]
 branch = true
-omit = */mozapkpublisher/test/*
+omit = */mozapkpublisher/tests/*
```
JohanLorenzo, line 31: I like the fact this rename was a dedicated commit. That made the review much easier. Thanks!

.git-blame-ignore-revs
```diff
@@ -0,0 +1,6 @@
+# Revisions to skip when running `git blame`.
+# Enable locally with:
+#     git config blame.ignoreRevsFile .git-blame-ignore-revs
+
+# lint(pushapk): apply the repo ruff config to mozapkpublisher
+af5fbde0bf855374062f9e759c82744816ab7701
```
JohanLorenzo, line 6: Good thinking 👍

JohanLorenzo (approved):
This is great! For many years, the people (myself included) who worked on mozapkpublisher have wondered whether it was worth merging it inside this repo. I'm glad it's the case today. Thank you for addressing this tech debt in such a way that history is preserved. For future readers, we did a similar thing with the Firefox for Android repositories [with this script](https://github.com/mozilla-mobile/firefox-android/blob/e9c2039e498cb58ca2ca4c3005ce988bd142bbc8/monorepo-migration/sync-fenix.sh) ([bug 1782733](https://bugzilla.mozilla.org/show_bug.cgi?id=1782733)). Based on that experience, I spotted-checked the commit history and messages. I thoroughly reviewed the additional 4 commits. Everything looks good to me. Great job!

jcristau:
The PR description could do with some motivation for this change?

PR #1523, opened 2026-09-17
fix(bitrisescript): stop the task payload from overriding workflow_id

hneiva:
workflow_id is derived from the task's scopes, but workflow_params were merged on top of it, letting a task run any workflow regardless of the scopes it holds. Set it after the merge and reject payloads that try.


ahal (approved):
Thanks, makes sense to me!

jcristau:
This should have referenced https://bugzilla.mozilla.org/show_bug.cgi?id=2071867

PR #1524, opened 2026-09-17
fix(beetmoverscript): don't overwrite objects in create-only buckets

hneiva:
The integration archive's credentials lack storage.objects.delete, so overwriting there fails with an opaque 403. Skip uploads whose content is already identical and raise a clear error when it differs.


beetmoverscript/src/beetmoverscript/utils.py
```diff
@@ -147,6 +159,12 @@ def is_upload_translations_artifacts_action(action):
     return action in TRANSLATIONS_ACTIONS
 
 
+def is_overwrite_forbidden(resource):
+    """Return True if the resource's (task bucket's) credentials are create-only
+    and must never overwrite existing objects (e.g. the integration archive)."""
+    return resource in NO_OVERWRITE_RESOURCES
```
bhearsum: I know it's a thing to have functions like this in scriptworker/these scripts...but this single-statement function is really adding no value, even considering readability.

Up to you whether or not it's kept, but I suggest removing it.

beetmoverscript/src/beetmoverscript/constants.py
```diff
@@ -89,6 +89,13 @@
 
 TRANSLATIONS_ACTIONS = ("upload-translations-artifacts",)
 
+# Resources (task buckets) whose credentials are create-only and therefore must
+# never overwrite existing objects -- overwriting in GCS requires
+# storage.objects.delete, which these intentionally lack. The autoland/mozregression
+# integration archive is immutable per-revision, so uploads of identical content
+# are skipped and differing content raises rather than attempting a forbidden overwrite.
+NO_OVERWRITE_RESOURCES = ("integration",)
```
ahal, line 97: Sounds a bit awkward, maybe `OVERWRITES_FORBIDDEN` to match the name of the utility function?

ahal (approved):
Thanks, lgtm!

beetmoverscript/src/beetmoverscript/gcloud.py
```diff
@@ -134,11 +135,24 @@ async def upload_to_gcs(context, target_path, path, expiry=None, fail_on_unknown
...
         if allow_overwrites:
+            # Normal buckets overwrite in place (e.g. nightly *-latest dirs), which
+            # also refreshes metadata such as custom_time/expiry.
             log.warning("upload_to_gcs: Overriding file: %s", target_path)
+        elif existing_blob.md5_hash == get_md5_base64(path):
+            # Create-only buckets (e.g. the autoland integration archive) can't
+            # overwrite -- that needs storage.objects.delete, which they lack. The
+            # identical object is already there, so skip and keep retries idempotent.
+            log.info("upload_to_gcs: %s already exists with identical content, skipping upload.", target_path)
+            return existing_blob
```
ahal, line 151: Actually just to clarify, are you sure that we want this behaviour in all cases? Previously this would raise which maybe would help detect a misconfiguration or something?

PR #1525, opened 2026-09-18
revert huawei store support for bustage

bhearsum:
Workers running with this crashed with:
```
error: HUAWEI_SERVICE_ACCOUNT is not set
```

@hneiva - if there's a quick, confident fix here we can avoid merging this...but I want to make sure that `master` gets back to a state where we can confidently deploy it before EOD.


jcristau (approved):

hneiva:
Thanks for putting this together. It's fixed in sops already, no need to revert.

